PR #129, opened 2015-12-16
Add Schematron rule to check for uniqueness macro names

rmzelle:
Coincidentally, two recent pull request failed the Travis tests because of duplicated macro names in the style:

https://github.com/citation-style-language/styles/pull/1818#issuecomment-163950384
https://github.com/citation-style-language/styles/pull/1831#issuecomment-164572121

This hasn’t come up often, but it would be good if the validator would flag it.

### DIFF
--- a/csl.rnc
+++ b/csl.rnc
@@ -18,7 +18,8 @@ dc:rights [
 dc:description [
   "RELAX NG compact schema for the Citation Style Language (CSL)."
 ]
-# Embedded Schematron rules to detect calls to nonexistent macros
+
+# Embedded Schematron rules
 
 sch:ns [ uri = "http://purl.org/net/xbiblio/csl" prefix = "cs" ]
 sch:pattern [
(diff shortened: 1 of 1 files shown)


rmzelle:
We can merge this once we confirm that this Schematron code is correct and works (I don't remember). It's not very crucial, though, since Travis CI already checks styles for unused macros since 2014: https://github.com/citation-style-language/styles/pull/1086

bdarcus:
The xpath expressions are correct on this, but one issue: in 2016, ISO released the standardized version of schematron, which has a different namespace URI, and seems very slightly different (I had to remove the name attribute to validate on the validator I tested with it, at least).

So I see three options: 

1. update to ISO, and merge this
2. remove schematron from the rnc file, and simply maintain the separate csl.sch file,updating it with this
3. remove schematron entirely and close this without merging

My impulse is to go with 2.

Anyone have any thoughts?

bdarcus:
I updated the csl.sch file per 2 [here](https://github.com/citation-style-language/schema/blob/iso-schematron/csl.sch).

bdarcus:
I'm going to close this. I ended up adding your new rule and also updating both rnc.csl and rnc.sch to ISO Schematron on #192.

PR #142, opened 2017-10-13
Add "table" to locator terms

sdruskat:
It should be possible to reference a table as well as figures, etc.


adam3smith:
I'll close this since we'll update locators as part of a larger schema update and table is covered in #166 together with others. Thanks!

Issue #126, opened 2015-06-14
Relicense schema under MIT license


rmzelle:
As discussed at http://xbiblio-devel.2463403.n2.nabble.com/MIT-license-as-default-for-CSL-software-tp7579193p7579395.html, it probably makes sense to relicense all the versions of the CSL schema under a more permissible license. Bruce, Simon, and I have agreed on the MIT license.

adam3smith:
Frank agrees, too.

rmzelle:
We currently claim copyright in 3 files: https://github.com/citation-style-language/schema/search?utf8=%E2%9C%93&q=copyright&type=

I think we can add an MIT license to the entire repo using the standard GitHub template. Will prep a PR.

PR #143, opened 2017-10-24
Switch licensing over to MIT license

rmzelle:
As decided in https://github.com/citation-style-language/schema/issues/126.

Probably still want to relicense the actual 0.8.1, 1.0, and 1.0.1 releases as well by making these same changes in release branches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,3 +145,7 @@ An example of an embedded citation object from Mendeley:
     }
 }
 ```
+
+# Licensing
+
+This repository is released under the MIT license.
--- a/csl-data.rnc
+++ b/csl-data.rnc
@@ -2,7 +2,8 @@ namespace dc = "http://purl.org/dc/elements/1.1/"
 
 dc:title [ "Citation Style Language Data" ]
 dc:creator [ "Bruce D'Arcus" ]
-dc:copyright [ "Bruce D'Arcus, 2009" ]
+dc:rights [ "Copyright 2009-2017 Bruce D'Arcus" ]
+dc:license [ "MIT license" ]
 dc:description [ "A schema for the CSL data model." ]
 include "csl-types.rnc"
 include "csl-variables.rnc"
--- a/csl-repository.rnc
+++ b/csl-repository.rnc
@@ -4,9 +4,8 @@ namespace dc = "http://purl.org/dc/elements/1.1/"
 
 dc:title [ "Extension schema for the Citation Style Language styles repository" ]
 dc:creator [ "Rintze M. Zelle" ]
-dc:rights [
-  "Copyright 2013 Rintze M. Zelle. Permission to freely use, copy and distribute."
-]
+dc:rights [ "Copyright 2013-2017 Rintze Zelle" ]
+dc:license [ "MIT license" ]
 dc:description [
   "Enforces stricter requirements for the styles in the official CSL styles repository."
 ]
--- a/csl.rnc
+++ b/csl.rnc
@@ -13,8 +13,9 @@ dc:creator [ "Simon Kornblith" ]
 bibo:editor [ "Frank Bennett" ]
 bibo:editor [ "Rintze Zelle" ]
 dc:rights [
-  "Copyright 2007-2012 by Frank Bennett, Bruce D'Arcus, Simon Kornblith, and Rintze Zelle. Permission to freely use, copy and distribute."
+  "Copyright 2007-2017 Frank Bennett, Bruce D'Arcus, Simon Kornblith, and Rintze Zelle"
 ]
+dc:license [ "MIT license" ]
 dc:description [
   "RELAX NG compact schema for the Citation Style Language (CSL)."
 ]


adam3smith:
Looks perfect to me -- thanks!

rmzelle:
Simon signed off on it as well with a thumbs-up, so just waiting for @bdarcus and @fbennett :).

bdarcus:
I need to think on the copyright issue; whether it should individual (as in
past), or assigned to "CSL."

@fbennett - you're a legal scholar. Do you have any concerns from that
angle?

In my brief looking at other examples, it seems the tendency is to assign
copyright to what are in effect legal entities: people, or corporate
entities (corporations usually).

Not sure what the status of CSL is there.

adam3smith:
To the extent that this is copyrighteable, you, as the authors, own the copyright. You would have to explicitly sign it over to CSL -- just writing here that CSL has the copyright won't do in legal terms. Given that CSL is not a legal entity, I think signing copyright over would put this on much less solid legal foundations and we shouldn't do it. 

We can ask the people at the software conservancy -- that type of advice is one of the things they do.

rmzelle:
Yes, so I guess the main question is to whether replace "Citation Style Language" in the newly added LICENSE.txt by our names.

I could also just remove LICENSE.txt, but it might be nice to license the entire repository contents.

fbennett:
I think @bdarcus is right about copyright to legal entities. If you think
about enforcement, CSL is not a personality that could file suit.

I'll check with people who know IP better, and get back.

rmzelle:
> We can ask the people at the software conservancy -- that type of advice is one of the things they do.

That would be good, especially since CSL's legal status will likely change if we become a member of the conservancy.

(we also have the similar question for other CSL repositories that already use the MIT license, such as https://github.com/citation-style-language/distribution-updater/blob/master/LICENSE.txt and https://github.com/citation-style-language/csl-validator/blob/gh-pages/LICENSE.txt)

fbennett:
We can ask the people at the software conservancy -- that type of advice
is one of the things they do.

... and I am reminded that most of what I know of the ins and outs of
copyright, I have learned from librarians. Plus-one for Sebastian's
suggestion.

dhimmel:
Why not just say:

> Copyright (c) 2007-2017 Contributors

Or

> Copyright (c) 2007-2017 Citation Style Language and Contributors

Hopefully this issue won't hold up merging this PR for much longer. The lack of a license is a much bigger issue!

Also would not CC0 make more sense? Just waive all copyright protection here, since individuals likely want to reuse these schema files in many places without worrying about legal stuff.

adam3smith:
let's not use CC0: https://wiki.creativecommons.org/wiki/CC0_FAQ#May_I_apply_CC0_to_computer_software.3F_If_so.2C_is_there_a_recommended_implementation.3F -- we want an OSI approved license to be safe, and it's not like using MIT is all that hard.

dhimmel:
From the Creative Commons FAQ:

> However, CC0 has not been approved by the Open Source Initiative and does not license or otherwise affect any patent rights you may have.

The main issue why CC0 was not OSI approved was the patent rights, which I don't think are relevant in the case of this repository. However, it's a good point that some users may want to use this repo under an OSI-approved license.

Hence, I suggest dual-licensing the repo as MIT and CC0. Thus users can choose which license they would like to reuse the content under. For more information on dual licensing see https://github.com/cognoma/cancer-data/issues/5 and the implementation in https://github.com/cognoma/cancer-data/commit/1d961fb2201209b3ec4774aaadd36f36cf6d165e.

> it's not like using MIT is all that hard.

First I don't think MIT is as appropriate as a CC license, because the content of this repository is less of code and more along the lines of documentation.

The use case I'm imagining is using `csl-data.json` in another repository, with some modifications to customize it, thereby creating a derivative. According to MIT:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

Therefore, I have to now store the MIT license as part of my repo. Obviously, non-compliance with this clause is high, which reflects that even the loose requirements of MIT are still too onerous / unclear for many users. I like having a CC0 option because it accommodates users without the legal expertise to comply with a license.

rmzelle:
> I suggest dual-licensing the repo as MIT and CC0

I'm not sure I'm on board with CC0. Let's first figure out how to license the schema/repo under the MIT license. It would help if we had some legal advise on the best party to claim copyright (individuals or "Citation Style Language" ). "Contributors" seems a bit too vague.

There are also some related discussions at:
https://softwareengineering.stackexchange.com/questions/157968/how-to-manage-a-copyright-notice-in-an-open-source-project
https://softwareengineering.stackexchange.com/questions/117572/mit-and-copyright
https://softwareengineering.stackexchange.com/questions/210472/is-renewal-of-mit-license-needed-on-github-at-the-beginning-of-each-year
https://opensource.stackexchange.com/questions/5778/why-do-licenses-such-as-the-mit-license-specify-a-single-year

bdarcus:
I agree. We could use some expert input.

adam3smith:
So I talked to some copyright people and the copyright statement is basically meaningless, so we can have it say whatever we want. These statements are relevant for earlier works, but according to current US copyright law, copyright rests automatically with the author(s) unless it is explicitly transferred (in writing) so that such statements have no legal effect. I think 

> Copyright (c) 2007-2018 Citation Style Language and Contributors

meets the spirit of what we want to say and rather than further delaying adding this license, we should just go ahead with that.

csl.rnc
```diff
@@ -13,8 +13,9 @@ dc:creator [ "Simon Kornblith" ]
 bibo:editor [ "Frank Bennett" ]
 bibo:editor [ "Rintze Zelle" ]
 dc:rights [
-  "Copyright 2007-2012 by Frank Bennett, Bruce D'Arcus, Simon Kornblith, and Rintze Zelle. Permission to freely use, copy and distribute."
+  "Copyright 2007-2017 Frank Bennett, Bruce D'Arcus, Simon Kornblith, and Rintze Zelle"
```
dhimmel: Reminder to update to 2018 or even removing these dates if they're more trouble than they're worth.

csl-repository.rnc
```diff
@@ -4,9 +4,8 @@ namespace dc = "http://purl.org/dc/elements/1.1/"
 
 dc:title [ "Extension schema for the Citation Style Language styles repository" ]
 dc:creator [ "Rintze M. Zelle" ]
-dc:rights [
-  "Copyright 2013 Rintze M. Zelle. Permission to freely use, copy and distribute."
-]
+dc:rights [ "Copyright 2013-2017 Rintze Zelle" ]
```
dhimmel: 2018

csl-data.rnc
```diff
@@ -2,7 +2,8 @@ namespace dc = "http://purl.org/dc/elements/1.1/"
 
 dc:title [ "Citation Style Language Data" ]
 dc:creator [ "Bruce D'Arcus" ]
-dc:copyright [ "Bruce D'Arcus, 2009" ]
+dc:rights [ "Copyright 2009-2017 Bruce D'Arcus" ]
```
dhimmel: 2018

README.md
```diff
@@ -145,3 +145,7 @@ An example of an embedded citation object from Mendeley:
     }
 }
 ```
+
+# Licensing
+
+This repository is released under the MIT license.
```
dhimmel: Consider linking to license file like:

```md
This repository is released under the [MIT license](LICENSE.txt).
```

Or even removing this section since most users will know that the LICENSE file is the relevant reference here.

LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2007-2017 Citation Style Language
```
dhimmel: See suggestion by @adam3smith in https://github.com/citation-style-language/schema/pull/143#issuecomment-428329245

> Copyright (c) 2007-2018 Citation Style Language and Contributors

rmzelle:
> I think
> 
> > Copyright (c) 2007-2018 Citation Style Language and Contributors
> 
> meets the spirit of what we want to say and rather than further delaying adding this license, we should just go ahead with that.

@fbennett, @bdarcus, @simonster, please let us know within a week if you have any objections to this change. Fine with me.

bdarcus:
Fine with me.

rmzelle:
@adam3smith (or others), can you double-check this PR before I merge this?

adam3smith:
Thanks! Ready to merge from my end.

PR #149, opened 2018-03-23
Fix references

thewilkybarkid:
ab8f90ddeb9824028e86a0cd86f6e1e4986d1c6b had renamed the definitions, so the file doesn't work in 1.0.1.

### DIFF
--- a/csl-data.rnc
+++ b/csl-data.rnc
@@ -11,7 +11,7 @@ reference = element reference { type, id, uri?, container-uri?, (contributor* & 
 
 ## Types 
 div {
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks! Sorry for the very long wait.

Were you actually using csl-data.rnc for anything? The JSON schemas are probably more accurate in describing the CSL JSON format.

thewilkybarkid:
I just had a look at it while investigating existing XML-based schemas for reference list data for https://libero.pub/. I remember not being able to find much documentation about it, so tried using the schemas directly to interrogate.

PR #156, opened 2018-08-11
Travis CI: basic Python test suite

dhimmel:
In response to discussion at https://github.com/citation-style-language/schema/issues/154#issuecomment-411871866, this is a PR to add some basic testing. Currently it only tests the JSON files and schemas.

It does not incorporate existing tests at https://github.com/citation-style-language/test-suite/issues/1#issuecomment-282565231, but these could be added in the future (if someone wants to help me).

I choose to use Python / pytest since that's what I'm most familiar with.

PR #158, opened 2018-10-15
CSL JSON: date-parts set minItems to 1

dhimmel:
Currently, the JSON Schema allows empty arrays in date-parts, which can cause downstream citeproc utilities to crash. See https://github.com/greenelab/manubot/issues/66

This PR sets minItems to 1 for the date-parts array as well as the nested arrays with year, month, day info.

CCing @adam3smith who I met last week at FORCE2018. IIRC, @adam3smith did not think there was any valid reason for CSL JSON to use empty date-parts arrays rather than omit the variable altogether.


dhimmel:
P.S. I'll write tests for this if https://github.com/citation-style-language/schema/pull/156 is merged.

rmzelle:
@adam3smith, @fbennett, this is fine with me. I came across some tests in the test suite that have empty date-parts arrays (see e.g. https://github.com/citation-style-language/test-suite/blob/1b7ce3ceb460dda689f3281c3a5f0617ed6d0805/processor-tests/humans/date_InPress.txt#L46; doesn't seem to be on purpose), but we can fix those up once we have CI set up.

Would make the schema stricter, though, so similar to our discussion at https://github.com/citation-style-language/schema/pull/155#issuecomment-444947021 on being cautious to invalidate what was previously valid CSL JSON.

rmzelle:
(will merge if one of you two gives the green light)

dsifford:
FYI: This is also included in #161

dhimmel:
> will merge if one of you two gives the green light

Pinging @adam3smith and @fbennett

rmzelle:
Let's wait with this until we update the schema draft.

adam3smith:
(as Daniel referenced in the initial post, I agree this is unproblematic)

dhimmel:
> Let's wait with this until we update the schema draft.

Not sure how progress on the schema draft updates is going, but doesn't it make sense to merge this change ASAP?

bdarcus:
@dhimmel - where does this now stand?

dhimmel:
want to add a test to see if it works.

@bdarcus i think your force push to to master might have confused to the git diff. probably best in future to not rewrite history on `master` except in extreme cases.

bdarcus:
OK, sorry.

Not followed this, but does it allow for undated entries?

dhimmel:
> Not followed this, but does it allow for undated entries?

Yes, but if you provide a date using `date-parts` it must include at least one element (like the year). But you can still omit any date field altogether.

dhimmel:
> where does this now stand?

good on my end pending review

bdarcus:
Thanks for the PR, and bearing with us.

Issue #127, opened 2015-06-23
JSON Schema (Draft 4)


mrenvoize:
JSON Schema has hit draft 4 now, but the json schema's here are still following the draft 3 spec.

rmzelle:
Good point. Did you create https://github.com/mrenvoize/schema/commit/f29d6c65d998a33cebdbc824068da15767e4e212 by hand, or is there a draft3-to-draft4 converter?

mrenvoize:
I created it.. just verifying it's all correct then I was going to make a pull request ;)

rmzelle:
Okay, let me know if you come across any issues. When submitting the PR, just modify the existing schema files. I might not be able to accept it right away, since I think Zotero and Mendeley both rely on the current JSON schema for validation, and we should make sure they both can handle draft 4. We'll deal with that once the PR is in good shape, though.

rmzelle:
See http://stackoverflow.com/questions/17205260/json-schema-draft4-vs-json-schema-draft3 and https://github.com/json-schema/json-schema/wiki/ChangeLog, and https://jsonschemalint.com/#/version/draft-04/markup/json.

rmzelle:
(Zotero developer `@adomasven` said that Zotero doesn't use these schema right now: https://forums.zotero.org/discussion/comment/283533/#Comment_283533)

mrenvoize:
I'd be really interested to see these updated.. Both to the latest schema sealy versions and to reflect the data structure zotero now uses as alluded to in that linked thread.  

_Sent from my Sony D5803 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #153, opened 2018-08-07
Specify draft via $schema in JSON schema

dhimmel:
The goal is to allow validators to detect which draft to use, without having to hardcode it in downstream applications. For example, I want my application to use the latest CSL Items schema and to continue working even if this repository upgrades to a newer schema draft.

PR #161, opened 2018-12-10
update and improve json schemas to draft 7

dsifford:
Follow up from #153, closes https://github.com/citation-style-language/schema/issues/127.

This _should_ be backwards compatible with the old schemas. And because of that, there's some glaring issues with some of the fields and types scatter throughout that were grandfathered in. These I think should be discussed and improved upon in another PR.

If you'd like, copy and paste the `csl-data.json` schema to [this validator](https://www.jsonschemavalidator.net/) to give it a shot. The `csl-citation.json` schema will not validate because it references the current broken schema url.

Let me know if you have any questions.

Ping: @rmzelle @dhimmel

### DIFF
--- a/csl-citation.json
+++ b/csl-citation.json
@@ -1,101 +1,103 @@
 {
+    "$id": "https://raw.githubusercontent.com/citation-style-language/schema/master/csl-citation.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
     "description": "JSON schema for CSL citation objects",
-    "$schema" : "http://json-schema.org/draft-03/schema#",
-    "id": "https://github.com/citation-style-language/schema/raw/master/csl-citation.json",
+    "definitions": {
+        "citationItemDefinition": {
+            "type": "object",
+            "properties": {
+                "id": {
+                    "type": [
+                        "string",
+                        "number"
+                    ]
+                },
+                "itemData": {
+                    "$ref": "csl-data.json"
+                },
+                "prefix": {
+                    "type": "string"
+                },
+                "suffix": {
+                    "type": "string"
+                },
+                "locator": {
+                    "type": "string"
+                },
+                "label": {
+                    "type": "string",
+                    "enum": [
+                        "book",
+                        "chapter",
+                        "column",
+                        "figure",
+                        "folio",
+                        "issue",
+                        "line",
+                        "note",
+                        "opus",
+                        "page",
+                        "paragraph",
+                        "part",
+                        "section",
+                        "sub verbo",
+                        "verse",
+                        "volume"
+                    ]
+                },
+                "suppress-author": {
+                    "type": [
+                        "string",
+                        "number",
+                        "boolean"
+                    ]
+                },
+                "author-only": {
+                    "type": [
+                        "string",
+                        "number",
+                        "boolean"
+                    ]
+                },
+                "uris": {
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "format": "uri"
+                    }
+                }
+            },
+            "additionalProperties": false
+        }
+    },
     "type": "object",
     "properties": {
         "schema": {
             "type": "string",
-            "required": true,
-            "enum" : [
-                "https://github.com/citation-style-language/schema/raw/master/csl-citation.json" 
-            ] 
+            "format": "uri"
         },
         "citationID": {
             "type": [
                 "string",
-                "number" 
-            ],
-            "required": true 
+                "number"
+            ]
         },
         "citationItems": {
             "type": "array",
             "items": {
-                "type": "object",
-                "properties": {
-                    "id": {
-                        "type": [
-                            "string",
-                            "number" 
-                        ],
-                        "required": true 
-                    },
-                    "itemData": {
-                        "$ref": "csl-data.json/#/items" 
-                    },
-                    "prefix": {
-                        "type": "string" 
-                    },
-                    "suffix": {
-                        "type": "string" 
-                    },
-                    "locator": {
-                        "type": "string" 
-                    },
-                    "label": {
-                        "type": "string",
-                        "enum": [
-                            "book",
-                            "chapter",
-                            "column",
-                            "figure",
-                            "folio",
-                            "issue",
-                            "line",
-                            "note",
-                            "opus",
-                            "page",
-                            "paragraph",
-                            "part",
-                            "section",
-                            "sub verbo",
-                            "verse",
-                            "volume" 
-                        ] 
-                    },
-                    "suppress-author": {
-                        "type": [
-                            "string",
-                            "number",
-                            "boolean" 
-                        ] 
-                    },
-                    "author-only": {
-                        "type": [
-                            "string",
-                            "number",
-                            "boolean" 
-                        ] 
-                    },
-                    "uris": {
-                        "type": "array",
-                        "items": {
-                            "type": "string" 
-                        } 
-                    } 
-                },
-                "additionalProperties" : false 
-            } 
+                "$ref": "#/definitions/citationItemDefinition"
+            }
         },
         "properties": {
             "type": "object",
             "properties": {
                 "noteIndex": {
-                    "type": "number" 
-                } 
+                    "type": "number"
+                }
             },
-            "additionalProperties" : false 
-        } 
+            "additionalProperties": false
+        }
     },
-    "additionalProperties" : false 
+    "required": ["schema", "citationID", "citationItems"],
+    "additionalProperties": false
 }
--- a/csl-citation.json
+++ b/csl-citation.json
@@ -67,6 +67,9 @@
                     }
                 }
             },
+            "required": [
+                "id"
+            ],
             "additionalProperties": false
         }
     },
--- a/csl-data.json
+++ b/csl-data.json
@@ -33,6 +33,14 @@
                                 }
                             },
                             "required": ["raw"]
+                        },
+                        {
+                            "properties": {
+                                "literal": {
+                                    "type": "string"
+                                }
+                            },
+                            "required": ["literal"]
                         }
                     ]
                 },
--- a/csl-data.json
+++ b/csl-data.json
@@ -15,13 +15,14 @@
                                     "items": {
                                         "type": "array",
                                         "items": {
-                                            "type": "number"
+                                            "type": [
+                                                "string",
+                                                "number"
+                                            ]
                                         },
-                                        "maxItems": 3,
-                                        "minItems": 1
+                                        "maxItems": 3
                                     },
-                                    "maxItems": 2,
-                                    "minItems": 1
+                                    "maxItems": 2
                                 }
                             },
                             "required": ["date-parts"]
@@ -55,10 +56,9 @@
                         },
                         "season": {
                             "type": [
+                                "string",
                                 "number"
-                            ],
-                            "minimum": 1,
-                            "maximum": 4
+                            ]
                         }
                     }
                 }
@@ -103,25 +103,8 @@
                                     "number",
                                     "boolean"
                                 ]
-                            },
-                            "static-ordering": {
-                                "type": [
-                                    "string",
-                                    "number",
-                                    "boolean"
-                                ]
-                            },
-                            "parse-names": {
-                                "type": [
-                                    "string",
-                                    "number",
-                                    "boolean"
-                                ]
                             }
-                        },
-                        "required": [
-                            "family"
-                        ]
+                        }
                     }
                 ]
             }
--- a/csl-data.json
+++ b/csl-data.json
@@ -127,14 +127,9 @@
             }
         },
         "numberDefinition": {
-            "oneOf": [
-                {
-                    "type": "number"
-                },
-                {
-                    "type": "string",
-                    "pattern": "^[0-9]+$"
-                }
+            "type": [
+                "number",
+                "string"
             ]
         },
         "requiredFields": {
--- a/csl-data.json
+++ b/csl-data.json
@@ -346,10 +346,16 @@
                 "ISSN": {
                     "type": "string"
                 },
+                "journalAbbreviation": {
+                    "type": "string"
+                },
                 "jurisdiction": {
                     "type": "string"
                 },
                 "keyword": {
+                    "type": "string"
+                },
+                "language": {
                     "type": "string"
                 },
                 "locator": {
@@ -398,6 +404,9 @@
                     "type": "string"
                 },
                 "section": {
+                    "type": "string"
+                },
+                "shortTitle": {
                     "type": "string"
                 },
                 "source": {
--- a/csl-data.json
+++ b/csl-data.json
@@ -1,447 +1,440 @@
 {
+    "$id": "https://raw.githubusercontent.com/citation-style-language/schema/master/csl-data.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
     "description": "JSON schema for CSL input data",
-    "$schema" : "http://json-schema.org/draft-03/schema#",
-    "id": "https://github.com/citation-style-language/schema/raw/master/csl-data.json",
-    "type": "array",
-    "items": {
-        "type": "object",
-        "properties": {
-            "type": {
-                "type": "string",
-                "required": true,
-                "enum" : [
-                    "article",
-                    "article-journal",
-                    "article-magazine",
-                    "article-newspaper",
-                    "bill",
-                    "book",
-                    "broadcast",
-                    "chapter",
-                    "dataset",
-                    "entry",
-                    "entry-dictionary",
-                    "entry-encyclopedia",
-                    "figure",
-                    "graphic",
-                    "interview",
-                    "legal_case",
-                    "legislation",
-                    "manuscript",
-                    "map",
-                    "motion_picture",
-                    "musical_score",
-                    "pamphlet",
-                    "paper-conference",
-                    "patent",
-                    "personal_communication",
-                    "post",
-                    "post-weblog",
-                    "report",
-                    "review",
-                    "review-book",
-                    "song",
-                    "speech",
-                    "thesis",
-                    "treaty",
-                    "webpage" 
-                ] 
-            },
-            "id": {
-                "type": [
-                    "string",
-                    "number" 
-                ],
-                "required": true 
-            },
-            "categories": {
-                "type": "array",
-                "items": {
-                    "type": "string" 
-                } 
-            },
-            "language": {
-                "type": "string" 
-            },
-            "journalAbbreviation": {
-                "type": "string" 
-            },
-            "shortTitle": {
-                "type": "string" 
-            },
-            "author": {
-                "type": "array",
-                "items": {
-                    "id": "name-variable",
-                    "type": [
+    "definitions": {
+        "dateDefinition": {
+            "type": "object",
+            "allOf": [
+                {
+                    "oneOf": [
                         {
                             "properties": {
-                                "family" : {
-                                    "type": "string" 
-                                },
-                                "given" : {
-                                    "type": "string" 
-                                },
-                                "dropping-particle" : {
-                                    "type": "string" 
-                                },
-                                "non-dropping-particle" : {
-                                    "type": "string" 
-                                },
-                                "suffix" : {
-                                    "type": "string" 
-                                },
-                                "comma-suffix" : {
-                                    "type": [
-                                        "string",
-                                        "number",
-                                        "boolean" 
-                                    ] 
-                                },
-                                "static-ordering" : {
-                                    "type": [
-                                        "string",
-                                        "number",
-                                        "boolean" 
-                                    ] 
-                                },
-                                "literal" : {
-                                    "type": "string" 
-                                },
-                                "parse-names" : {
-                                    "type": [
-                                        "string",
-                                        "number",
-                                        "boolean" 
-                                    ] 
-                                } 
+                                "date-parts": {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "array",
+                                        "items": {
+                                            "type": "number"
+                                        },
+                                        "maxItems": 3,
+                                        "minItems": 1
+                                    },
+                                    "maxItems": 2,
+                                    "minItems": 1
+                                }
                             },
-                            "additionalProperties" : false 
+                            "required": ["date-parts"]
                         },
                         {
                             "properties": {
-                                "literal" : {
-                                    "type": "string" 
-                                } 
+                                "raw": {
+                                    "type": "string"
+                                }
                             },
-                            "additionalProperties" : false 
-                        } 
-                    ] 
-                } 
-            },
-            "collection-editor": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "composer": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "container-author": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "director": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "editor": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "editorial-director": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "interviewer": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "illustrator": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "original-author": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "recipient": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "reviewed-author": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "translator": {
-                "type": "array",
-                "items": {
-                    "$ref": "name-variable" 
-                } 
-            },
-            "accessed": {
-                "id": "date-variable",
-                "type": [
+                            "required": ["raw"]
+                        }
+                    ]
+                },
+                {
+                    "properties": {
+                        "circa": {
+                            "type": [
+                                "string",
+                                "number",
+                                "boolean"
+                            ]
+                        },
+                        "season": {
+                            "type": [
+                                "number"
+                            ],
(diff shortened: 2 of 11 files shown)


rmzelle:
Great, thanks. Will you address the merge conflict?

And can you briefly explain how you managed to upgrade the schemas? Are there software libraries to upgrade schemas to newer versions, or was it hand-work?

> This _should_ be backwards compatible with the old schemas. And because of that, there's some glaring issues with some of the fields and types scatter throughout that were grandfathered in. These I think should be discussed and improved upon in another PR.

Can you give an example of what I should be looking for in this PR?

dsifford:
Whoops, didn't see that there was a conflict.

> And can you briefly explain how you managed to upgrade the schemas? Are there software libraries to upgrade schemas to newer versions, or was it hand-work?

Mostly by hand.. I used [quicktype](https://quicktype.io/app) to quickly generate raw object definitions, but most of those had to be adjusted for better accuracy. Then I used the validator I liked above to double check that the object types checked out using a few known-accurate objects.

I also used a few schemas on [schemastore.org](http://schemastore.org/json/) for inspiration.

> Can you give an example of what I should be looking for in this PR?

I'll annotate some lines where changes were made and where I think the schema can be improved in this PR after I fix the conflict.

csl-citation.json
```diff
@@ -1,101 +1,103 @@
...
+                        "string",
+                        "number",
+                        "boolean"
+                    ]
+                },
+                "uris": {
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "format": "uri"
```
dsifford: Changed this to specifically require `uri` format.

csl-citation.json
```diff
@@ -1,101 +1,103 @@
...
+        "citationItemDefinition": {
+            "type": "object",
+            "properties": {
+                "id": {
+                    "type": [
+                        "string",
+                        "number"
+                    ]
+                },
+                "itemData": {
```
dsifford: This is not used whatsoever in citeproc-js.

csl-citation.json
```diff
@@ -1,101 +1,103 @@
...
+                        "format": "uri"
+                    }
+                }
+            },
+            "additionalProperties": false
+        }
+    },
     "type": "object",
     "properties": {
         "schema": {
```
dsifford: Why is this necessary?

csl-citation.json
```diff
@@ -1,101 +1,103 @@
...
             "type": "string",
-            "required": true,
-            "enum" : [
-                "https://github.com/citation-style-language/schema/raw/master/csl-citation.json" 
-            ] 
+            "format": "uri"
         },
         "citationID": {
```
dsifford: Wishlist: This should be called `id`.

csl-citation.json
```diff
@@ -1,101 +1,103 @@
...
+                    "type": "number"
+                }
             },
-            "additionalProperties" : false 
-        } 
+            "additionalProperties": false
+        }
     },
-    "additionalProperties" : false 
+    "required": ["schema", "citationID", "citationItems"],
```
dsifford: Changed this to include `citationItems` because it should always return an array, even if empty IMHO.

csl-data.json
```diff
@@ -1,431 +1,440 @@
...
+                            "properties": {
+                                "date-parts": {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "array",
+                                        "items": {
+                                            "type": "number"
+                                        },
+                                        "maxItems": 3,
+                                        "minItems": 1
```
dsifford: Added `minItems`

csl-data.json
```diff
@@ -1,431 +1,440 @@
...
-                                    "type": "string" 
-                                },
-                                "parse-names" : {
-                                    "type": [
-                                        "string",
-                                        "number",
-                                        "boolean" 
-                                    ] 
-                                } 
+                                "raw": {
```
dsifford: Made this an "either-or" between `raw`, and `date-parts`.

csl-data.json
```diff
@@ -1,431 +1,440 @@
...
-                } 
-            },
-            "accessed": {
-                "id": "date-variable",
-                "type": [
+                    ]
+                },
+                {
+                    "properties": {
+                        "circa": {
```
dsifford: Boolean? What is this exactly.

csl-data.json
```diff
@@ -1,431 +1,440 @@
...
+                                "string",
+                                "number",
+                                "boolean"
+                            ]
+                        },
+                        "season": {
+                            "type": [
+                                "number"
+                            ],
+                            "minimum": 1,
```
dsifford: Changed this to be a better spec fit.

csl-data.json
```diff
@@ -1,431 +1,440 @@
...
-                                    "items": {
-                                        "type": [
-                                            "string",
-                                            "number" 
-                                        ] 
-                                    },
-                                    "maxItems": 3 
-                                },
-                                "maxItems": 2 
+                            "literal": {
```
dsifford: Made this an "either-or" between `literal` and the rest of the fields.

If no `literal`, then it should at least have `family` (see below)

csl-data.json
```diff
@@ -1,431 +1,440 @@
...
+                                "type": "string"
+                            },
+                            "non-dropping-particle": {
+                                "type": "string"
                             },
-                            "season": {
+                            "suffix": {
+                                "type": "string"
+                            },
+                            "comma-suffix": {
```
dsifford: Not sure what this is, nor `static-ordering` or `parse-names`

csl-data.json
```diff
@@ -1,431 +1,440 @@
...
+            }
+        },
+        "numberDefinition": {
+            "oneOf": [
+                {
+                    "type": "number"
+                },
+                {
+                    "type": "string",
+                    "pattern": "^[0-9]+$"
```
dsifford: Changed this so that if it's a string, it is still just numbers in the string.

dsifford:
Just realized that I removed `literal` from the date definition because I thought they were the same thing.. That's not correct -- `raw` needs to be a parsable date (using ideally the json schema `date-time` [RFC3339](https://tools.ietf.org/html/rfc3339#section-5.6) type), whereas `literal` is a literal date string that isn't parsed or touched in any way.

I'll add that fix in after I hear back.

dsifford:
@rmzelle Yikes... also just realized that there are **dozens** of fields listed here that are not listed in `csl.variables.rnc` --- gotta add those in as well.

csl-citation.json
rmzelle: It's not?

https://github.com/rmzelle/ref-extractor/wiki#single-item-zotero-citation shows an example of how Zotero embeds CSL citation metadata in Word documents. It shows "itemData" as a field of "citationItems", and it contains an array of item metadata as defined by csl-data.json.

csl-citation.json
rmzelle: See https://github.com/rmzelle/ref-extractor/wiki#single-item-zotero-citation again. Zotero and Mendeley both reference the "csl-citation.json" schema in their embedded citation metadata.

csl-citation.json
rmzelle: Any reason why?

I think it was chosen to clarify that this is the ID of the citation object, not of the items referenced within.

csl-data.json
rmzelle: It's a toggle to indicate whether a date is uncertain. See http://docs.citationstyles.org/en/stable/specification.html#approximate-dates.

csl-data.json
rmzelle: static-ordering: https://github.com/Juris-M/citeproc-js/blob/master/manual/citeproc-doc.rst#non-byzantine-names
parse-names: https://github.com/Juris-M/citeproc-js/blob/master/manual/citeproc-doc.rst#names-1

rmzelle:
> there are ... fields listed in the old csl-data.json that are not listed in csl.variables.rnc --- gotta add those in as well. ... `journalAbbreviation`, `language`, `shortTitle`

None of these have a corresponding CSL variable with the same name.

"language" indicates the language of the item in question, and is used by the CSL processor to determine if special processing is necessary (see e.g. http://docs.citationstyles.org/en/stable/specification.html#non-english-items). We chose not to make it a CSL variable, since it generally wouldn't make much sense to print the field contents in a formatted reference. The field contents should ideally be an easily parseable locale code. And even if the field would contain a human readable language descriptor (e.g. "English"), we don't have the infrastructure in place to translate that for CSL styles set to non-English locales (e.g. to "Anglais" for French).

> What is the difference between `shortTitle` and `title-short`? Can one of these be axed?

Regarding "journalAbbreviation" and "shortTitle", CSL originally only made the abbreviated "short" versions of "container-title" and "title" available in CSL styles through `variable="container-title" form="short"` and `variable="title" form="short"`. The CSL JSON still needed to store these abbreviations somehow, and @fbennett came up with "journalAbbreviation" and "shortTitle". Only later, in CSL 1.0.1, did we allow these abbreviated versions to be called directly through `variable="container-title-short"` and `variable="title-short" form="short"`. At that time, we added corresponding fields to CSL JSON but kept the old fields for backwards compatibility. We should drop these old fields eventually.

See also http://docs.citationstyles.org/en/stable/release-notes.html#new-variables-container-title-short-and-title-short

rmzelle:
@dsifford, do you still have the versions of the CSL JSON schemas on hand right after you upgraded them from schema draft 3 to 7?

I'm happy to review any changes you'd like to suggest to the schema itself, but I much rather have separate pull requests for the schema upgrade and such changes.

csl-citation.json
dsifford: Hm.. Must be a single one-off case built into citeproc-js then. Because that doesn't exist anywhere in my implementation.

csl-citation.json
dsifford: Right, but why cant they use JSON schema as it's indended and just reference the base schema using the top-level `$id` property? It seems redundant and unnecessary.

rmzelle:
> Though a couple will make checks a bit more strict (as noted above).

It would be better to put anything that is not purely related to the draft-3 to draft-7 upgrade in a dedicated pull request.

> After I get around to adding in those 3 missing fields

Add which fields where?

dsifford:
> Add which fields where?

`journalAbbreviation`, `language`, and `shortTitle` are still missing

rmzelle:
> `journalAbbreviation`, `language`, and `shortTitle` are still missing

They're in csl-data.json now:

https://github.com/citation-style-language/schema/blob/f01ba9c5ec2055e381a38598919a379255c496c5/csl-data.json#L63
https://github.com/citation-style-language/schema/blob/f01ba9c5ec2055e381a38598919a379255c496c5/csl-data.json#L66
https://github.com/citation-style-language/schema/blob/f01ba9c5ec2055e381a38598919a379255c496c5/csl-data.json#L69

dsifford:
Right, but I removed them in this PR. I have to add them back in.

rmzelle:
> Right, but I removed them in this PR. I have to add them back in.

Ah, okay. Yes. Sorry for the confusion.

dhimmel:
I created a PR into this PR to update the Python test suite: https://github.com/dsifford/schema/pull/1.

In short, we must upgrade the version of the python `jsonschema` package to support draft-07. Also removed jsonref, because I believe @dsifford fixed the bug that required that package in the first place.

There is still [one failure](https://travis-ci.com/dsifford/schema/builds/95034455#L480) in `test_data_schema_with_extra_property_fails` because it seems extra properties are not causing validation failure. If that is intended, you should change the test.

dsifford:
Nice job @dhimmel... Just pushed 2 of the spec fixes that were talked about above that add back in a few fields I missed. Only thing left to do is rewind some things that aren't fully 100% backwards compatible and then we can probably merge this and open a new PR for draft-7 related schema updates.

csl-data.json
dhimmel: I think that will break many things. `edition`, `issue`, `volume`often contain characters. I think it makes sense to save any functional changes for subsequent PRs.

csl-data.json
```diff
@@ -1,431 +1,457 @@
...
-                    "number" 
-                ] 
+            {
+                "$ref": "#/definitions/numberFields"
             },
-            "year-suffix": {
-                "type": "string" 
-            } 
-        },
-        "additionalProperties" : false
```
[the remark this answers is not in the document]
dsifford: Nope, it was not.

Although, just so you're aware, there are some large APIs that currently serve "CSL JSON" responses that are peppered with all kinds of additional properties, so not sure how that would possibly affect those API consumers/maintainers...

<details>
<summary><strong>Here's an example of a typical response from CrossRef</strong></summary>
<pre>
{
    "indexed": {
        "date-parts": [[2018, 12, 18]],
        "date-time": "2018-12-18T19:41:10Z",
        "timestamp": 1545162070737
    },
    "reference-count": 0,
    "publisher": "BMJ",
    "license": [
        {
            "URL": "http://www.bmj.com/company/legal-information/terms-conditions/legal-information/tdm-licencepolicy",
            "start": {
                "date-parts": [[2018, 12, 13]],
                "date-time": "2018-12-13T00:00:00Z",
                "timestamp": 1544659200000
            },
            "delay-in-days": 0,
            "content-version": "tdm"
        }
    ],
    "content-domain": { "domain": ["bmj.com"], "crossmark-restriction": true },
    "abstract": "<jats:title>Abstract</jats:title>\n          <jats:sec>\n            <jats:title>Objective</jats:title>\n            <jats:p>To determine if using a parachute prevents death or major traumatic injury when jumping from an aircraft.</jats:p>\n          </jats:sec>\n          <jats:sec>\n            <jats:title>Design</jats:title>\n            <jats:p>Randomized controlled trial.</jats:p>\n          </jats:sec>\n          <jats:sec>\n            <jats:title>Setting</jats:title>\n            <jats:p>Private or commercial aircraft between September 2017 and August 2018.</jats:p>\n          </jats:sec>\n          <jats:sec>\n            <jats:title>Participants</jats:title>\n            <jats:p>92 aircraft passengers aged 18 and over were screened for participation. 23 agreed to be enrolled and were randomized.</jats:p>\n          </jats:sec>\n          <jats:sec>\n            <jats:title>Intervention</jats:title>\n            <jats:p>Jumping from an aircraft (airplane or helicopter) with a parachute versus an empty backpack (unblinded).</jats:p>\n          </jats:sec>\n          <jats:sec>\n            <jats:title>Main outcome measures</jats:title>\n            <jats:p>Composite of death or major traumatic injury (defined by an Injury Severity Score over 15) upon impact with the ground measured immediately after landing.</jats:p>\n          </jats:sec>\n          <jats:sec>\n            <jats:title>Results</jats:title>\n            <jats:p>\n              Parachute use did not significantly reduce death or major injury (0% for parachute\n              <jats:italic>v</jats:italic>\n              0% for control; P&gt;0.9). This finding was consistent across multiple subgroups. Compared with individuals screened but not enrolled, participants included in the study were on aircraft at significantly lower altitude (mean of 0.6 m for participants\n              <jats:italic>v</jats:italic>\n              mean of 9146 m for non-participants; P&lt;0.001) and lower velocity (mean of 0 km/h\n              <jats:italic>v</jats:italic>\n              mean of 800 km/h; P&lt;0.001).\n            </jats:p>\n          </jats:sec>\n          <jats:sec>\n            <jats:title>Conclusions</jats:title>\n            <jats:p>Parachute use did not reduce death or major traumatic injury when jumping from aircraft in the first randomized evaluation of this intervention. However, the trial was only able to enroll participants on small stationary aircraft on the ground, suggesting cautious extrapolation to high altitude jumps. When beliefs regarding the effectiveness of an intervention exist in the community, randomized trials might selectively enroll individuals with a lower perceived likelihood of benefit, thus diminishing the applicability of the results to clinical practice.</jats:p>\n          </jats:sec>",
    "DOI": "10.1136/bmj.k5094",
    "type": "article-journal",
    "created": {
        "date-parts": [[2018, 12, 13]],
        "date-time": "2018-12-13T22:34:15Z",
        "timestamp": 1544740455000
    },
    "page": "k5094",
    "update-policy": "http://dx.doi.org/10.1136/crossmarkpolicy",
    "source": "Crossref",
    "is-referenced-by-count": 0,
    "title": "Parachute use to prevent death and major trauma when jumping from aircraft: randomized controlled trial",
    "prefix": "10.1136",
    "author": [
        {
            "given": "Robert W",
            "family": "Yeh",
            "sequence": "first",
            "affiliation": []
        },
        {
            "given": "Linda R",
            "family": "Valsdottir",
            "sequence": "additional",
            "affiliation": []
        },
        {
            "given": "Michael W",
            "family": "Yeh",
            "sequence": "additional",
            "affiliation": []
        },
        {
            "given": "Changyu",
            "family": "Shen",
            "sequence": "additional",
            "affiliation": []
        },
        {
            "given": "Daniel B",
            "family": "Kramer",
            "sequence": "additional",
            "affiliation": []
        },
        {
            "given": "Jordan B",
            "family": "Strom",
            "sequence": "additional",
            "affiliation": []
        },
        {
            "given": "Eric A",
            "family": "Secemsky",
            "sequence": "additional",
            "affiliation": []
        },
        {
            "given": "Joanne L",
            "family": "Healy",
            "sequence": "additional",
            "affiliation": []
        },
        {
            "given": "Robert M",
            "family": "Domeier",
            "sequence": "additional",
            "affiliation": []
        },
        {
            "given": "Dhruv S",
            "family": "Kazi",
            "sequence": "additional",
            "affiliation": []
        },
        {
            "given": "Brahmajee K",
            "family": "Nallamothu",
            "sequence": "additional",
            "affiliation": []
        }
    ],
    "member": "239",
    "published-online": { "date-parts": [[2018, 12, 13]] },
    "container-title": "BMJ",
    "original-title": [],
    "language": "en",
    "link": [
        {
            "URL": "http://data.bmj.org/tdm/10.1136/bmj.k5094",
            "content-type": "unspecified",
            "content-version": "vor",
            "intended-application": "text-mining"
        },
        {
            "URL": "https://syndication.highwire.org/content/doi/10.1136/bmj.k5094",
            "content-type": "unspecified",
            "content-version": "vor",
            "intended-application": "similarity-checking"
        }
    ],
    "deposited": {
        "date-parts": [[2018, 12, 18]],
        "date-time": "2018-12-18T19:05:10Z",
        "timestamp": 1545159910000
    },
    "score": 1.0,
    "subtitle": [],
    "short-title": [],
    "issued": { "date-parts": [[2018, 12, 13]] },
    "references-count": 0,
    "alternative-id": ["10.1136/bmj.k5094"],
    "URL": "http://dx.doi.org/10.1136/bmj.k5094",
    "relation": {},
    "ISSN": ["0959-8138", "1756-1833"],
    "subject": ["General Medicine"],
    "container-title-short": "BMJ"
}
</pre>
</details>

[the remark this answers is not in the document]
dhimmel: > so not sure how that would possibly affect those API consumers/maintainers...

Not at all... they currently return CSL that does not abide by the JSON Schema. I.e. this is the status quo. For Manubot, we actually use the JSON schema to filter the CSL returned by Crossref and remove all the crap.

rmzelle:
> Only thing left to do is rewind some things that aren't fully 100% backwards compatible and then we can probably merge this and open a new PR for draft-7 related schema updates.

@dsifford, I haven't merged this yet since it's my understanding you still have some rewinding to do.

csl-data.json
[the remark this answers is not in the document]
dsifford: If we continue to use `allOf` and use refs as I do in this PR, then there's no way currently possible to have `additionalProperties: false`. So pick your poison. I personally think it's fine, but if you feel strongly about it, we'd have to scrap this.

cc: @rmzelle

csl-data.json
dsifford: To my knowledge, this is the only potentially breaking change that I made. I'll revert this now and then we should be good to go @rmzelle.


@dhimmel feel free to give this another once-over if you'd like.

csl-citation.json
dsifford: This shouldn't be breaking so I'm gonna leave it.

dsifford:
@rmzelle This should be good to go now, pending any objections by @dhimmel

csl-data.json
```diff
@@ -1,431 +1,452 @@
...
+                                    "string",
+                                    "number",
+                                    "boolean"
+                                ]
                             }
                         },
-                        "additionalProperties" : false 
+                        "required": [
+                            "family"
+                        ]
```
rmzelle: I'm not sure we can require "family".

I just checked and a Zotero user can store just a given name, which gives this as exported CSL JSON:

```
			{
				"family": "Nielsen",
				"given": "Jens"
			},
			{
				"family": "",
				"given": "bla"
			}
```

Other implementations might leave out the field entirely.

csl-data.json
```diff
@@ -1,431 +1,452 @@
...
+                },
+                "jurisdiction": {
+                    "type": "string"
+                },
+                "keyword": {
+                    "type": "string"
+                },
+                "language": {
+                    "type": "string"
+                },
```
rmzelle: Maybe it would make sense to carve out a separate definition for non-CSL variable metadata fields like "language"?

csl-data.json
```diff
@@ -1,431 +1,452 @@
...
+                "ISBN": {
+                    "type": "string"
+                },
+                "ISSN": {
+                    "type": "string"
+                },
+                "journalAbbreviation": {
+                    "type": "string"
```
rmzelle: Maybe we should use a separate definition here too, to indicate that "journalAbbreviation" is deprecated in favor of "container-title-short". Idem for "shortTitle" and "title-short".

csl-citation.json
```diff
@@ -1,101 +1,103 @@
...
+    "definitions": {
+        "citationItemDefinition": {
+            "type": "object",
+            "properties": {
+                "id": {
+                    "type": [
+                        "string",
+                        "number"
+                    ]
+                },
```
rmzelle: The original schema requires "id" for each citation item, which is a requirement we should keep.

csl-citation.json
```diff
@@ -1,101 +1,103 @@
...
+                "itemData": {
+                    "$ref": "csl-data.json"
```
rmzelle: I haven't tried using your version schema, but are you sure this shouldn't be ` "$ref": "csl-data.json/#/items"` like in the original?

csl-data.json
```diff
@@ -1,431 +1,452 @@
...
-                                        "boolean" 
-                                    ] 
-                                } 
+                                "date-parts": {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "array",
+                                        "items": {
+                                            "type": "number"
+                                        },
```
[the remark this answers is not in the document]
dhimmel: Zotero [currently formats](https://github.com/zotero/zotero/issues/1603) years as strings rather than integers during CSL export.

I think upgrading the schema draft is a great enhancement. However, I think it is crucial that backwards incompatible changes are not made in the initial draft change.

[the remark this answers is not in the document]
rmzelle: There are a few things that got more strict that I didn't flag, like setting required fields for date-parts (either a "date-parts", "literal" or "raw" date), but those seemed pretty basic. Agree we should be careful here.

csl-data.json
fbennett: These two attributes were cruft from the early days. I've removed them from fixtures in the test suite, they can be removed from the schema.

csl-data.json
```diff
@@ -1,447 +1,440 @@
...
+                "title-short": {
+                    "type": "string"
+                },
+                "URL": {
+                    "type": "string"
+                },
+                "version": {
+                    "type": "string"
+                },
+                "year-suffix": {
```
fbennett: This is a property of the item in rendered context. Any input value would be ignored.

csl-data.json
```diff
@@ -1,447 +1,440 @@
...
+                "archive-place": {
+                    "type": "string"
+                },
+                "authority": {
+                    "type": "string"
+                },
+                "call-number": {
+                    "type": "string"
+                },
+                "citation-label": {
```
fbennett: This is generated by processors, not set on input.

csl-data.json
```diff
@@ -1,447 +1,440 @@
...
+                    "type": "string"
+                },
+                "citation-number": {
```
fbennett: This is a property of the item in context. Processors generate the value internally. Any input value will be ignored.

csl-data.json
```diff
@@ -1,447 +1,440 @@
...
+                "DOI": {
+                    "type": "string"
+                },
+                "event": {
+                    "type": "string"
+                },
+                "event-place": {
+                    "type": "string"
+                },
+                "first-reference-note-number": {
```
fbennett: This is a processor-generated value that depends on the position of references in a document. Any value set on input will be ignored.

dsifford:
I'll try and look at this again this weekend and address issues raised. Sorry for stalling!

rmzelle:
> I'll try and look at this again this weekend and address issues raised. Sorry for stalling!

@dhimmel, I prepared a minimal update to draft 6 (it looks like no changes are necessary for draft 7) at https://github.com/citation-style-language/schema/pull/168, inspired by this PR. I would prefer to merge that, and then create one or more additional PRs to pull in improvements/changes proposed here (such as using proper definitions etc.).

csl-data.json
dsifford: So can this be removed?

csl-data.json
dsifford: So can this be removed?

csl-data.json
dsifford: So can this be removed?

csl-data.json
dsifford: So can this be removed?

dsifford:
@rmzelle Ah, shoot. I missed your message.

I went ahead and addressed all(?) of the comments/concerns raised here. It should theoretically be good to go. But your call.

csl-data.json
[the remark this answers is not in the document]
dsifford: Ping @rmzelle -- My comment above still holds relevance.

dhimmel:
Note @dsifford that I'm not a maintainer of this repo, but I appreciate the work you've put in, and hope to see as much of it integrated as makes sense.

Is it possible to split this PR up into smaller ones following #168? I think it will help with review and the pace of things, if the PRs are very focused and do one and only one thing. In addition, each change should indicate whether it functionally changes the schema or not.

> It should theoretically be good to go.

Note that you have a [test failure](https://travis-ci.org/citation-style-language/schema/builds/526713099) of `test_data_schema_with_extra_property_fails`. If you think think this constraint should be removed as per https://github.com/citation-style-language/schema/pull/161#discussion_r247332076, you should either delete that test or add a skip decorator like:

```python
@pytest.mark.skip(reason="Setting additionalProperties to false not possible with allOf")
```

clbarnes:
Any blockers on getting this merged? I'm working on a downstream package and would prefer not to have to start everything over when the updated schema lands.

bdarcus:
Closed this in favor of #168. Any missing pieces can be re-added via dedicated PRs.

Issue #141, opened 2017-10-06
add support for "narrative citations" (aka natbib \citet)


bdarcus:
This has been a long time feature request, and connects to issues like #140.

Goal: to support this feature while retaining a key early design goal of CSL, which is that an author can easily switch back and forth between note-based styles, and other style formats.

Currently, to get this sort of formatting, the expectation is that the author manually writes the author name(s), and then tells the editing environment that the author name should be suppressed on output, so that in an author-date style, only the date outputs.

This solves the design goal mentioned above, since the formatting does not change when switching to a note style, with the downside that authors must manually maintain the author names in text.

Nobody has figured out how to both support this feature and retain the original design goal, but I would be thrilled if someone could figure it out and post it below.

bwiernik:
The general idea that seems to have emerged from various discussions on x-biblio and the Zotero forums that styles would need to specify two different citation formats. For author-date styles, these would generally take the forms: (Author, Date) and Author (Date). For numeric and note styles, these
would be: [1] and Author [1].

These two formats solve the design goal—the innovation for numeric/note styles is that they need to list both the Author in text and the citation number.

jalvesaq:
Thanks, @rmzelle for linking my request to the right place.
It would be great to see this feature implemented!
I have a solution ([abntref](https://github.com/jalvesaq/abntref)) for people who have to follow the ABNT (Brazilian Association of Technical Norms) rules, but it is useful only for those who write (R)Markdown documents on Linux.

fbennett:
There is now an [intext-element branch](https://github.com/Juris-M/citeproc-js/tree/intext-element) in the `citeproc-js` repo, and a [simple test](https://github.com/Juris-M/citeproc-js/blob/intext-element/tests/fixtures/local/intext_SimpleName.txt) that illustrates what it does. A calling application would invoke the processor with an "author-only" tag to get the intext element, then with "suppress-author" to get the remainder of the citation, as shown in [this earlier test](https://github.com/Juris-M/citeproc-js/blob/intext-element/tests/fixtures/local/discretionary_AuthorDateAuthorOnlyThenSuppressAuthor.txt). (More cumbersome than a single call, but in a footnote style, two insert operations will be required anyway.)

The `intext-element` branch is just an experimental frolic at the moment---if something like this were to move toward production, the "author-only" and "suppress-author" flags should be set on the `citation` object, rather than on individual items.

bwiernik:
[...]

PR #173, opened 2020-04-15
add cs:intext element

bdarcus:
This adds an optional cs:intext element to support definition of AuthorOnly rendering required to assemble intext (aka narrative, or textual) citations. 

It addresses #141 by implementing the syntax used by @fbennett in his citeproc-js [intext-element](https://github.com/Juris-M/citeproc-js/tree/intext-element) branch.

Note: 

1. this change does not require changes to existing styles, nor should it break processors, which can just ignore the new element
2. there is still a question about what the spec should say about how this new element works with number and note styles. In author-date, authors get suppressed when rendered in-text. But I'm not sure in the other style classes, but we can leave that for a documentation PR

### DIFF
--- a/csl.rnc
+++ b/csl.rnc
@@ -13,7 +13,7 @@ dc:creator [ "Simon Kornblith" ]
 bibo:editor [ "Frank Bennett" ]
 bibo:editor [ "Rintze Zelle" ]
 dc:rights [
-  "Copyright 2007-2018 Citation Style Language and contributors"
+  "Copyright 2007-2020 Citation Style Language and contributors"
 ]
 dc:license [ "MIT license" ]
 dc:description [
@@ -66,6 +66,7 @@ div {
       (style.locale*
        & style.macro*
        & style.citation
+       & style.intext?
        & style.bibliography?)
     }
   
@@ -401,6 +402,14 @@ div {
     ## Use to describe the formatting of citations.
     element cs:citation { citation.options, sort?, citation.layout }
   
+  style.intext =
+
+    ## Defines the author-only rendering for assembly of a textual citations
+    ## (for example, "Doe [3]" or "Doe (2018)"), where the output expectations
+    ## for in-text authors are different than the default citation rendering;
+    ## for example, in APA, or numeric styles.
+    element cs:intext { citation.options, sort?, citation.layout }
+
   style.bibliography =
     
     ## Use to describe the formatting of the bibliography.


jalvesaq:
@bdarcus and all others who contributed with the discussion that resulted in this pull request: Thank you for working on this issue!

I currently use [zotcite](https://github.com/jalvesaq/zotcite) (a [Vim](https://www.vim.org/)/[Neovim](https://neovim.io/) plugin) to extract reference data from Zotero database and put them in the YAML header of the Markdown document which is then passed to [pandoc-citeproc](https://github.com/jgm/pandoc-citeproc). Is it possible to test this pull request with `pandoc-citeproc`?

denismaier:
No. 
Pandoc-citeproc has long had a feature to produce in-text citations. But this here is new and currently only implemented in citeproc-js.

bdarcus:
A couple of things;

1. Maybe the content of the element should.only be a required author names element? Is there value in my broader definition?

2. Technically, this probably could be done without a new element, with a reserved macro, and in that case, the primary change would only be in the spec.

3. author-in-text would be more verbose, but also more self-explanatory

bwiernik:
1. It may or may not be a `names` element. For example, some numeric citation styles want the intext element to be `[1]` (not superscript) for in-text citations, but <sup>`[1]`</sup> for regular citations. We should think of `cs:intext` as being an override to the default form being the author `names` element extracted from `cs:citation`.

2. I'm not sure what you mean. So you saying instead of `cs:intext`, this could be a macro with `name="intext"` or `name="narrative"` or similar? Again with the default being the processor-extracted `author-only` information? So then, the change would only need to be a change in spec instructing processors (1) how to implement the `author-only`, `suppress-author`, and `composite` behavior, and (2) to use `macro name="intext"` as a reserved name and substitute that for the regular `author-only` contents? That sounds reasonable to me. 

3. I think we should avoid `author-in-text` considering that it may not always be an author (e.g., some numeric styles, legal cases would be short title, etc.). I think that explanation in the spec can elaborate that the main function of `intext`/`narrative` is to control the author-in-text behavior, but I prefer `intext`/`narrative` for the label.

bdarcus:
I need to fix the git commits, but the 1.1 one is splitting up the schema into a 1.1 version and 1.0 version, with the intext element on 1.1.

rmzelle:
> Anyway, there are a couple of comments like on line 217 "Will be disallowed with CSL 1.1." What shall we do about these?

We tried to avoid any situations with CSL 1.0.1 where CSL 1.0 styles that validated fine against the 1.0 schema would suddenly throw errors against the 1.0.1 schema, to avoid causing any breakage.

The 1.0.1 schema isn't bifurcated between 1.0 and 1.0.1 styles, so I just marked the things that should be removed in a future version where full backwards-compatibility isn't a concern as obsolete. Note also that CSL 1.0 didn't yet have separate definitions for `<info/>` for independent and dependent styles, so when we split things out in CSL 1.0.1, there were some combinations that were previously allowed but didn't make sense (like having `<link href="..." rel="independent-parent"/>` in an independent style.

See also https://github.com/citation-style-language/schema/commit/ac98d787a86a64facb4f5a8971519748ffd90892.

bdarcus:
Question: should we change this to an attribute on citation? Would give us more flexibility going forward. So something like:

``` xml
<citation command="intext">
```

bwiernik:
So you are thinking a style might have two `<citation>` elements, one for regular and one for intext?

bdarcus:
Yes, or in the future, potentially more.

>

bdarcus:
Obvious downside is it's potentially less backward compatible.

bwiernik:
Upsides of `command` syntax:
  1. More extensible in the future if additional citation modes are added.
  2. Unifies the style writing syntax with the recommended processor API.
      - [citeproc-js has](https://citeproc-js.readthedocs.io/en/latest/running.html#special-citation-forms) regular, author-only (intext), suppress-author, and composite modes. 
      - Styles require a `command="regular"` (default) `cs:citation` element
      - Styles could have a `command="intext"` (and perhaps even a `command="suppress-author"`) `cs:citation` element to override the default format generated by the processor.

Downsides:
  1. `cs:intext` is already implemented as a separate element in citeproc-js. So, this would involve a change to the most widely used CSL processor. 
  2.  Less backwards compatible because current processors will trip up on multiple `cs:citation` elements. This would push the intext feature to CSL 1.1 at the earliest.

I don't know that either approach is necessarily easier or more elegant.

bdarcus:
I don't have a strong opinion either, but I thought we should discuss.

On the citeproc-js point, that should involve a trivial code change for
Frank; the model wouldn't change, other than searching for
cs:citation/@command="in-text" rather than cs:in-text.

denismaier:
But doesn't that come back to what I suggested a while ago?
https://github.com/citation-style-language/schema/issues/141#issuecomment-613026829

I do like the idea due to the advantages already mentioned. If we're only talking about the in-text variant I don't know if it's that much better, but extensibility and transparence is strong argument.

> On the citeproc-js point, that should involve a trivial code change for Frank; the model wouldn't change, other than searching for cs:citation/@command="in-text" rather than cs:in-text.

Here, I'm not sure. If we have two `citation` elements on the same level, just with different `commands`, I'd expect both to be complete with no suppression/substitution logic involved.

bwiernik:
> Here, I'm not sure. If we have two citation elements on the same level, just with different commands, I'd expect both to be complete with no suppression/substitution logic involved.

This will need to be explained in the documentation clearly, but it is really important to understand that `intext` **does not** refer to the full "narrative" form of the citation. It refers *only* to the part of the citation that is actually in text. The full "narrative" citation is the "compound" form that combines the "intext" form with the "suppress-author" form, with an optional `infix`.

This structure is important because the Zotero developers have emphasized that there needs to be an option for `infix` specification to avoid easy document corruption with some WYSIWYG editors. A `mode="narrative"` structure would thus require including an `<infix/>` element in the style code to specify where the infix would go. That would be odd and confusing for style writers.

The most important benefit of the intext solution that is in the PR here is that the automatic default forms based on author-only and suppress-author make this feature automatically available to most existing CSL styles with no or trivial modification. That is critical for this feature to practically see the light of day.

Thinking it through, I think that separate `cs:intext` and `cs:citation` would be the least confusing for style writers. `cs:intext` provides the structure that appears in-text when the citation mode is specified as `compound` or `intext`.

denismaier:
> This will need to be explained in the documentation clearly, but it is really important to understand that intext does not refer to the full "narrative" form of the citation. It refers only to the part of the citation that is actually in text. The full "narrative" citation is the "compound" form that combines the "intext" form with the "suppress-author" form, with an optional infix.

Yes. And it's much easier to make this clear if the element us called `intext` and not `citation command=intext`.

bdarcus:
Ok, I'm fine with keeping it as is then.

denismaier:
> This will need to be explained in the documentation clearly, but it is really important to understand that `intext` **does not** refer to the full "narrative" form of the citation. It refers _only_ to the part of the citation that is actually in text. The full "narrative" citation is the "compound" form that combines the "intext" form with the "suppress-author" form, with an optional `infix`.

Perhaps we should not only make this clear in the documentation but also choose a better name for the element?

Something like `author-in-text`. Or `author-only` because this indeed renders the author-only part of a narrative citation. (Might be slightly misleading though because you could expect an author-only citation in parentheses or in a note.)

(Or we could even use a second `citation` element with a `command` attribute, like `<citation command="author-in-text>`. This would be extensible, but it would not change Frank's mechanism. @bwiernik I understand your previous comments as if your main concern is the mechanism, but that an extensible naming scheme would be a good idea. Right?)

bwiernik:
`author-in-text` is more confusing. This element is not always an author, and one of the major reasons that this would be specified is to specify a structure that is *not* an author (e.g., for numeric styles that should appear as `[1] found supportive results.`

I’m not sure what exactly is confusing about a name and description ala `cs:intext`: “output that is shown in the document body (outside of the citation parentheses or note) when citation mode is ‘in-text’ (or ‘ suppress-author’ whatever we call this) or ‘compound’`

denismaier:
Ok. Yeah, your right. With a good explanation that should be clear. Then let's just leave it as is (and get this out asap).

bdarcus:
@rmzelle @adam3smith - can you approve this and the other PR, or request a change, so we can merge?

bwiernik:
As fixed by the surgery, looks good to me.

denismaier:
Good for me as well.

bdarcus:
closing this in favor of #193 (long story)

bdarcus:
As I may have mentioned on some issue somewhere recently, I think one of the things we need to think about, and probably discuss at the meeting next month, is reverting #193, and replacing with the more general mechanism discussed above.

I'm convinced that's a better, more general, solution, that will be easier for developers to implement.

An upgrade script can also automate the implicit support now for suppress author and narrative commands.

The only downside I can see is backward compatibility. But surely that's worth solving? Maybe we call 1.1 instead 2.0, for example?

WDYT @denismaier @bwiernik?

PS - I thought about pulling this out into a new issue, but we lose the full thread then.

bdarcus:
> So long as an alternative is something like `<citation mode="intext">` and retains the behavior we built into `cs:intext`, I am okay with that. The key benefit of the way that `cs:intext` was designed is that all of the existing styles in the repository are automatically able to be used with in-text citations without modification. If every style would have to be manually updated, it would functionally mean that this feature would never be widely used.

I'm not sure we want to formally treat `intext` as different from a processing POV, because that means exceptions for developers to account for.

The nice thing about the command approach of multiple template is they just use consistent logic across any of them. So there's the default template, and the rest, and they all work the same. And that spec language can be simple and straightforward.

But the issue you raise is important, and I had in mind two workarounds:

1. as I mentioned above, a script can explicitly apply the logic that 1.0 developers used to implement these to automatically upgrading the styles
2. we can also document that logic, say in an appendix or something, so developers can even apply it to 1.0 styles, as no doubt processors like `citeproc-rs` already do.

Certainly there would be details to sort out, perhaps to discuss at the meeting, but would that basically work?

BTW, I (and I think Denis) came to this position after working a lot on the Emacs/Org-mode citation system. Its csl export processor already has support for the following styles (which I've indicated with comments). 

```elisp
ELISP> (org-cite-supported-styles '(csl))
((("author" "a") ; author-only
  ("caps-full" "cf")
  ("caps" "c")
  ("full" "f"))
 (("noauthor" "na") ; aka suppress-author
  ("bare-caps" "bc")
  ("caps" "c")
  ("bare" "b"))
 (("year" "y") ; year only
  ("bare" "b"))
 (("text" "t") ; in-text/narrative
  ("caps-full" "cf")
  ("full" "f")
  ("caps" "c"))
 (("nil") ; default
  ("bare-caps" "bc")
  ("caps" "c")
  ("bare" "b"))
 (("nocite" "n"))) ; nocite
```

Admittedly, @andras-simonyi has done this without explicit style support, but I'm hoping to get this input too. He would have to write code to handle either of these to support APA and such.

Also worth noting org-cite distinguish "styles" (aka "commands") from "variants" ("caps" and such).

bwiernik:
I am really not following what your proposal is. Can you describe it in a self-contained place?

The fundamental design choice we made that I think we need to preserve is:

1. There must be an automatic default method for creating an in-text citation layout for styles that don't provide one.


This has 3 major benefits:

1. It ensures that the feature can actually be used. If styles had to be manually updated to add this feature, then it would frankly never happen. There is no way we are going to be able to manually update the thousands of styles in the repo. 
2. It is already exactly or effectively implemented in the two most prominent citation processors (citeproc-js and pandoc). Fundamentally changing the design now means _more_ implementation work for processor developers.

bdarcus:
> I am really not following what your proposal is. Can you describe it in a self-contained place?

1. revert `cs:intext`, replace with allowing more than one `cs:citation`.
2. when more than one `cs:citation`, each must have a unique `@command` attribute (or alternately, always required that attrbute, with at least one with a value of "default"; details here still TBD).

And then separately:
 
> The fundamental design choice we made that I think we need to preserve is:
> 
> 1. There must be an automatic default method for creating an in-text citation layout for styles that don't provide one.
> 
> This has 3 major benefits:
> 
> 1. It ensures that the feature can actually be used. If styles had to be manually updated to add this feature, then it would frankly never happen. There is no way we are going to be able to manually update the thousands of styles in the repo.

No, but why not automate it? If it can be automated in processor, then it can be automated in a style update script. But see below.

> 2. It is already exactly or effectively implemented in the two most prominent citation processors (citeproc-js and pandoc). Fundamentally changing the design now means _more_ implementation work for processor developers.

And also `citeproc-el`.

TBH, this is really a question for those developers; how to deal with this.

I see a few options, which aren't necessarily mutually-exclusive:

1. document the algorithms for these (both suppress-author and in-text)
2. use 1 to automate style updates
3. require 1.1/2.0-compliant processors to support 1 also (which means works with 1.0 also)
4. leave 1 optional

E.g. there are a variety of options.

Does that make sense then?

bdarcus:
I'll try to do a PR that implements this, and ask for feedback from folks.

PR #182, opened 2020-04-23
adding variables for next minor releases

bdarcus:
This PR adds variable strings for next minor release, with one issue addressed (usually one addition) per commit. 

We can also squash the commit if preferable, but I figured better to give us flexibility.


bdarcus:
Sorry; made a dumb mistake on this and closed it. Unfortunately, I can't delete it.

PR #183, opened 2020-04-23
adding variables for 1.0.x minor release

bdarcus:
This PR bundles together some variable additions, one per issue. In most cases, this means one commit per variable.

This gives us the flexibility to squash them all together on merge, or keep the separate commit history.


csl-variables.rnc
```diff
@@ -15,6 +15,7 @@ div {
   ## Date variables
   variables.dates =
     "accessed"
+    | "available-date"
     | "container"
```
denismaier, line 19: Not related to you commit: But why is container a date variable?

csl-variables.rnc
```diff
@@ -80,6 +83,8 @@ div {
     | "original-title"
     | "page"
     | "page-first"
+    | "part-number"
```
denismaier, line 86: Is there a special reason why part number is not a number variable?

csl-variables.rnc
```diff
@@ -94,5 +99,6 @@ div {
     | "title-short"
     | "URL"
     | "version"
+    | "volume"
```
denismaier, line 102: Shouldn't that be `volume-title`?

csl-variables.rnc
bdarcus: I vaguely remember some issue mentioning this as a bug, but not 100% sure.

csl-variables.rnc
bdarcus: Yep; stupid mistake. Thanks.

csl-variables.rnc
bdarcus: no; are we 100% sure it should be moved? I'm forgetting.

denismaier: I'm not... @adam3smith @bwiernik What do you think?

bdarcus:
I just squashed the typo and fix commits for volume-title.

csl-variables.rnc
rmzelle: I don't recall any previous discussion on this, but as far as I can tell, "container" being a date variable dates back to at least 0.8: https://github.com/citation-style-language/schema/blob/v0.8/csl.rnc#L409 (and I was never able to figure out what "container" was for, hence the "?" at https://docs.citationstyles.org/en/1.0.1/specification.html#date-variables)

csl-variables.rnc
bwiernik: I think it is probably here because it is similar to `page`, which also isn't a number variable. The main function of a variable being a number is so that it can be formatted as an ordinal. While usually they would be formatted as "Part 2", etc., I can imagine cases where a style might call for an ordinal form (especially in French --e.g., `2ème partie`). So making `part-number` a number makes sense to me.

csl-variables.rnc
bwiernik: There is only one style in the repository that uses `container` (http://www.zotero.org/styles/wiley-vch-books), and there it just includes it in a list with all of the other date variables. I suggest we remove `container` in this release (https://github.com/citation-style-language/csl-evolution/issues/12). 

If a change to a hierarchical data model ever happens, we can add `container` back in at that time.

denismaier: Removing `container` is a good idea, but what does that mean for our versioning discussing. Then we *theoretically* backwards incompatible, albeit without any real consequences...

csl-variables.rnc
```diff
@@ -94,5 +96,6 @@ div {
     | "title-short"
     | "URL"
     | "version"
+    | "volume-title"
```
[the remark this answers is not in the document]
bwiernik: For workflow, should I keep this comment here or move it to an issue like the volume-title one?

csl-variables.rnc
bwiernik: The [CSL-M schema](https://citeproc-js.readthedocs.io/en/latest/csl-m/#page-and-page-first-modification) changes `page` and `page-first` to numeric variables. I think this should be done in CSL as well. Current citeproc-js behavior on hyphen-substitution in `cs:text` elements produces undesirable output: https://discourse.citationstyles.org/t/en-dash-subsittution-in-text/1497

Relatedly, I think `locator` should also be a number?

bdarcus:
Do those sort of changes matter from a compatability perspective?

Are they just as minor a change as the current commit?

bwiernik:
> Do those sort of changes matter from a compatability perspective?
> 
> Are they just as minor a change as the current commit?

I think moving `page`, `page-first`, `locator`, and `part-number` to numeric is sufficiently minor. From a processor perspective, it just adjusts the hash table for valid `cs:number` variables.

bdarcus:
Does mean styles need to be changed, though. That's a pretty big deal;
isn't it?

bwiernik:
No, number variables can be rendered in `cs:text` (and there is not difference in their behavior if they are (cf. [this discussion](https://discourse.citationstyles.org/t/en-dash-subsittution-in-text/1497/13)). All this change would do is _permit_ (not require) these variables to also be rendered in `cs:number`.

bdarcus:
Done.

There are some other number-related variables. Any of those that should be
moved too?

bwiernik:
> There are some other number-related variables. Any of those that should be moved too?

`call-number` should not be. `citation-number` and `first-reference-note-number` might conceivably be formatted using roman numerals, so moving them seems good. I don't see any other number-like variables.

bdarcus:
OK, I added those two and rebased.

>

denismaier:
> No, number variables can be rendered in `cs:text` (and there is not difference in their behavior if they are (cf. [this discussion](https://discourse.citationstyles.org/t/en-dash-subsittution-in-text/1497/13)). All this change would do is _permit_ (not require) these variables to also be rendered in `cs:number`.

But will it have consequences for validation?

bwiernik:
> But will it have consequences for validation?

No, only Jurism enforces that number variables must be rendered using `cs:number`. In fact, in the CSL repository, most styles render number variables using `cs:text`.

bdarcus:
> But will it have consequences for validation?

To understand why it won't (currently): see [this](https://github.com/citation-style-language/schema/blob/master/csl-variables.rnc#L13).

csl-variables.rnc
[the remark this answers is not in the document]
bdarcus: > For workflow, should I keep this comment here or move it to an issue like the volume-title one?

I missed this earlier. I'm not really sure. Just leave it here for now, and we can figure it out over time?

PR #187, opened 2020-05-21
add github actions test workflow

dhimmel:
Haven't test this yet... will update with whether it works.

I find GitHub actions is a bit slicker than Travis and is nicely integrated with GitHub. If we merge this, could then remove Travis.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,16 @@
+name: Python Test Suite
+on:
+- push
+- pull_request
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+    - name: Install dependencies
+      run: pip install --requirement=tests/requirements.txt
+    - name: Run Tests
+      run: pytest --verbose tests
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -13,4 +13,4 @@ jobs:
     - name: Install dependencies
       run: pip install --requirement=tests/requirements.txt
     - name: Run Tests
-      run: pytest --verbose tests
+      run: pytest --verbose --color=yes tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,0 @@
-language: python
-python:
-  - "3.6"
-cache: pip
-install:
-  - pip install --requirement=tests/requirements.txt
-script:
-  - pytest --verbose tests


dhimmel:
looks to be working https://github.com/dhimmel/csl-schema/runs/697715882

will remove Travis since not much point in having both services active

dhimmel:
@bdarcus care to review?

bdarcus:
I don't really know anything about this, but I merged, figuring if there's some minor problem, we can fix it :-)

bdarcus:
And of course, as soon as I say that, I get this:

https://github.com/citation-style-language/schema/actions/runs/111868243

bdarcus:
Maybe the error reported was just temporary; related to the transition?

dhimmel:
> And of course, as soon as I say that, I get this:

haha. it's trying to resolve some non-existent URL ([log](https://github.com/citation-style-language/schema/runs/697807024#step:5:26)):

```
url = 'https://github.com/citation-style-language/schema/raw/master/date-variable'
```

I'll look into it

dhimmel:
I think the problem has to do with how we define `date-variable`:

https://github.com/citation-style-language/schema/blob/991d8f606c8193c39debfdc07de18f9f88f20205/csl-data.json#L195

https://github.com/citation-style-language/schema/blob/991d8f606c8193c39debfdc07de18f9f88f20205/csl-data.json#L240

In draft-07, it seems like this might need to be placed in a definition and referenced like `{ "$ref": "#/definitions/date-variable" }`. I think @dsifford dealt with this in https://github.com/citation-style-language/schema/pull/161. What we want is the most minimal change to get our definitions working.

bdarcus:
Oh, so the GitHub action did work correctly?

If yes, do you think you could submit another PR to fix the bug it
uncovered?

dhimmel:
> Oh, so the GitHub action did work correctly?

I think so, the error message is not super clear, but I think the issue is that the upgrade to draft-07 (and draft-06) in https://github.com/citation-style-language/schema/commit/7179185c9b886eb44cf29eef0617081a6ba5dfd0 has some problems. These problems did not surface until we added the test in this pull request, since previously no tests were checking for whether definitions worked.

> do you think you could submit another PR to fix the bug it uncovered?

probably not soon. I don't have much experience with JSON schemas. @dsifford might be able to take part of #161 to fix it without too much hassle.

bdarcus:
That validators aren't very helpful.

> @dsifford might be able to take part of #161 to fix it without too much hassle.

That'd be awesome if he could; I'm struggling with this.

bdarcus:
So the problem may be simply that the pattern that's referenced is in the wrong place.

dsifford:
Let me know if you aren't able to get it sorted @bdarcus and I'll have a look later this afternoon.

dhimmel:
Okay, I opened a dedicated issue for this at https://github.com/citation-style-language/schema/issues/188

PR #191, opened 2020-05-22
Reformat readme for improved aesthetics

dhimmel:
use one sentence per line to improve git diffs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 # Schemas describing the Citation Style Language
 
+<!--
+When editing this file, use line breaks to separate sentences or phrases, rather than wrapping the text at a fixed character count.
+This helps git produce clean diffs and keeps reflowing to minimum.
+More info at https://rhodesmill.org/brandon/2012/one-sentence-per-line/
+-->
+
 This is the official repository for schemas describing the Citation Style Language (CSL).
 Current schemas include:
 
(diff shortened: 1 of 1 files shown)


bdarcus:
Can you say a little more, for people that might edit the README in the future?

What's the editor setting we want to avoid using; the equivalent of auto-fill on emacs?

dhimmel:
> Can you say a little more, for people that might edit the README in the future?

This PR does many types of small formatting changes, but generally they'd be picked up by a markdown linter. Other changes are using the right element types for things, for example heading levels and `code`.

Regarding one sentence per line, the idea is described at https://rhodesmill.org/brandon/2012/one-sentence-per-line/. The idea is to place source newlines between phrases or sentences. It's not supported by most automatic markdown formatters, but they generally leave line breaks alone. So just when editing or adding to the README, it's best if contributors use one sentence per line.

bdarcus:
Thanks.

Markdownlint is complaining about a few things, the most relevant of which is probably line length. What do you make of that? In the article you linked, he breaks lines, and starts each sentence on a new line.

bdarcus:
Seems to be issues with that too related to trailing whitespace.

dhimmel:
I would disable the line length warning. Most text editors wrap long lines, so there's not actually a problem with long sentences on a single line.

> Seems to be issues with that too related to trailing whitespace.

we don't want trailing whitespace (unless rare case when its used for `<br>`). Sorry if I added any of this.

bdarcus:
No problem; am just asking.

PR #193, opened 2020-05-24
add new cs:intext element

### LINKED ISSUE #141: add support for "narrative citations" (aka natbib \citet)

bdarcus:
Supports definition of AuthorOnly rendering required to assemble
intext (aka narrative, or textual) citations.

Closes #141.

### DIFF
--- a/csl.rnc
+++ b/csl.rnc
@@ -72,6 +72,7 @@ div {
       (style.locale*
        & style.macro*
        & style.citation
+       & style.intext?
        & style.bibliography?)
     }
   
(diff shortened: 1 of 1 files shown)


bdarcus:
Should be OK; it's a simple change.

bdarcus:
In part because of recent discussions, I'm thinking this should go in 1.1.

So I've changed the PR to merge to the v1.1 branch, to make it easier to manage.

That OK?

denismaier:
Ok for me.

PR #196, opened 2020-05-25
Change "and" to allow "short", use term

### LINKED ISSUE #137: Make "and" attribute in name lists more configurable
The current definition of the "and" attribute in the specification is:

> Specifies the delimiter between the second to last and last name of the names in a name variable. Allowed values are “text” (selects the “and” term, e.g. “Doe, Johnson and Smith”) and “symbol” (selects the ampersand, e.g. “Doe, Johnson & Smith”).

So there are only two options: the "long" form of the "and" term or the hardcoded ampersand. There have been several reports that this is too limited. Some languages don't seem to use ampersands at all (we received some private communication from @cpina that this is e.g. the case for Spanish and Catalan: '"y" (and) and in Catalan they do like in Spanish ("i", it's "and" in Catalan).'), and some styles require different text forms between the in-text citations and bibliography, e.g. an abbreviated and unabbreviated form (see https://github.com/citation-style-language/styles/pull/2571#issuecomment-287253895).

Most of this could be solved by changing the spec to have `and="symbol"` select the "symbol" form of the "and" term, and by defining the ampersand as translation for this term form in the locale files, which is a relatively easy fix.

A larger but more flexible change would be to make it possible to also e.g. choose the "short" form of the "and" term. This would be possible by e.g. having the "and" attribute directly select the desired term-form, with e.g. attribute values "long", "short", and "symbol" (`and="long"` would replace `and="text"`). Might also want to rename the attribute to "and-form", although I don't think that's very pretty.

bdarcus:
`and="text"` -> `and="long"`; makes clear rendering for "and" should be based on term definition, and so also to allow configuration of `and="symbol"`.

closes #137

### DIFF
--- a/csl.rnc
+++ b/csl.rnc
@@ -13,7 +13,7 @@ dc:creator [ "Simon Kornblith" ]
 bibo:editor [ "Frank Bennett" ]
 bibo:editor [ "Rintze Zelle" ]
 dc:rights [
-  "Copyright 2007-2018 Citation Style Language and contributors"
+  "Copyright 2007-2020 Citation Style Language and contributors"
 ]
 dc:license [ "MIT license" ]
 dc:description [
--- a/csl.rnc
+++ b/csl.rnc
@@ -59,10 +59,10 @@ include "csl-categories.rnc"
 div {
   start =
     independent-style.style | dependent-style.style | locale-file.locale
-  
+
   independent-style.style =
     element cs:style {
-      
+
       ## Select whether citations appear in-text or as notes.
       attribute class { "in-text" | "note" },
       style.default-locale,
@@ -74,7 +74,7 @@ div {
        & style.citation
        & style.bibliography?)
     }
-  
+
   dependent-style.style =
     element cs:style {
       style.default-locale,
@@ -82,27 +82,27 @@ div {
       dependent-style.style.info,
       dependent-style.style.legacy-attributes
     }
-  
+
   locale-file.locale =
     element cs:locale {
-      
+
       ## Specify the locale of the locale file.
       attribute xml:lang { xsd:language },
       version,
       locale-file.locale.info?,
       (locale.style-options & locale.date+ & locale.terms)
     }
-  
+
   style.default-locale =
-    
+
     ## Set a default style locale.
     attribute default-locale { xsd:language }?
-  
+
   version =
-    
+
     ## Indicate CSL version compatibility ("1.0" for CSL 1.0).
     [ a:defaultValue = "1.0" ] attribute version { "1.0" }
-  
+
   ## Obsolete for dependent styles. Will be disallowed with CSL 1.1.
   dependent-style.style.legacy-attributes =
     attribute class { "in-text" | "note" }?,
@@ -114,7 +114,7 @@ div {
   ]
 ]
 div {
-  
+
   ## Metadata for independent styles.
   independent-style.style.info =
     element cs:info {
@@ -133,7 +133,7 @@ div {
       & info.title-short?
       & info.updated
     }
-  
+
   ## Metadata for dependent styles.
   dependent-style.style.info =
     element cs:info {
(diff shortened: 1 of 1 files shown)


bdarcus:
Note: I used the Atom editor here, which removed a bunch of whitespace characters that didn't need to be there.

csl.rnc
```diff
@@ -430,174 +430,178 @@ div {
...
+
       ## Use the "and" term (e.g., "Doe, Johnson and Smith").
       "text"
-      | 
-        ## Use the "ampersand" (e.g., "Doe, Johnson & Smith").
+      |
+        ## Use the "ampersand" by default, or whatever is defined in the term (e.g., "Doe, Johnson & Smith").
         "symbol"
+      |
+        ## Use a "short" for as defined in the term. Not common in English, but is seen in other languages.
```
fbennett, line 477: for -> form

bdarcus: Doh! Thanks; fixed.

bdarcus:
> My inclination is to simplify this explanation. Replace `"text"` with `"long"`.

What would be the style compatibility implications of that though? This was why I didn't make that change.

PS - this is the kind of comment that ideally is in the "review" mode.

bdarcus:
Except, maybe because frank did a review you  don't have that option?

denismaier:
> 
> 
> Except, maybe because frank did a review you don't have that option?

No, should be possible. I could request a review from @bwiernik (or myself, for that matter), so bwiernik might be able to do so as well.

bdarcus:
OK,  but if I request reviews from the "PR reviewers" team, I
basically want whoever has the time to review. So if it were to close
when one review, that would be less than ideal.

bdarcus:
> With that in mind, I think replacing `and="text"` with `and="long"` is the best state to reach. At a minimum, we should add `and="long"` as the default and describe `and="text"` as a deprecated alias of `and="long"`.
> 
> Doing a batch replace of `and="text"` with `and="long"` in the official styles is simple enough. If we want to maintain compatibility for user styles in the wild, then let's add `and="long"` now and deprecate `and="text"`, then remove `and="text"` in CSL 1.1.

OK, gotcha.

So this is the key strategic question. Anyone else want to weigh in here?

denismaier:
So, what exactly is the current suggestion?

bdarcus:
> So, what exactly is the current suggestion?

Really two options:

1. deprecate `and="text"` now and replace with `and="long"`
2. add  `and="long"`now and deprecate `and="text"` later (say 1.1)

denismaier:
> * deprecate `and="text"` now and replace with `and="long"`
That would entail a schema update to 1.1?

> * add  `and="long"`now and deprecate `and="text"` later (say 1.1)
This would be possible with a 1.0.x minor release?

Am I right here?

And we will also have a third option `symbol` alongside `long` and `short`? You weren't talking about dropping this as well?

bdarcus:
> This would be possible with a 1.0.x minor release?

I think so.

I'd rather just do 1 for 1.1. It's cleaner and less confusing.

> And we will also have a third option symbol alongside long and short? 

Yes.

bdarcus:
I've updated the code and doc strings to where I think it needs to be, for 1.1 release.

bwiernik:
To be clear on my word usage here: “Deprecate” means “still around, but don’t use. We are warning you that it will go away in the future”. “Remove” means actually removing it.

denismaier:
> To be clear on my word usage here: “Deprecate” means “still around, but don’t use. We are warning you that it will go away in the future”. “Remove” means actually removing it.

Would also be my understanding.

denismaier:
I have no general objections to adding this with a 1.1. release. But I just hope that will be before 2025.

bdarcus:
:-)

I hope we can do it this summer, accepting that we won't have the spec
stuff that Norm is looking for done at that point.

bdarcus:
I'm hoping to merge this to the v1.1 branch later on Friday, so if any input (including affirmative, that result in an "approve" review), please get it in by then.

bdarcus:
> My inclination is to simplify this explanation. Replace `"text"` with `"long"`. Describe in the spec that the localized term `and` is rendered, with `form` given by the `and` attribute on `names`.

I missed this earlier. I can still update the language, as this is a bit more precise.

Issue #197, opened 2020-05-25
add RNC schema validation check to CI


bdarcus:
We would like to confirm that a commit to the included RNC files results in a valid `csl.rnc` schema.

We need to extend the current GitHub Actions to support this.

The only way I know to validate RNC is by running the java-based Trang. If the conversion to XML completes without error, it's valid.

This is how one does it in a terminal:

``` sh
java -jar trang.jar csl.rnc csl.rng
```

bdarcus:
@dhimmel - is there a straightforward way to do this?

rmzelle:
Maybe set this up in Travis CI?

https://docs.travis-ci.com/user/languages/java/
https://mvnrepository.com/artifact/org.relaxng/trang/20181222

bdarcus:
@dhimmel removed Travis in favor of GitHub Actions. Either way, we do need this.

dhimmel:
GitHub Actions has Java images:

https://help.github.com/en/actions/language-and-framework-guides/github-actions-for-java
https://github.com/actions/setup-java

Where does `trang.jar` come from? Is there a command to install it?

bdarcus:
I think most package managers have trang. That's how I install it on fedora.

bdarcus:
I think this is where the source is these days:

https://code.google.com/archive/p/jing-trang/downloads

bdarcus:
This *should* be working now. 

Thanks much for the quick PR @dhimmel!

bdarcus:
Example is here:

https://github.com/citation-style-language/schema/runs/707588711

rmzelle:
> I think this is where the source is these days: https://code.google.com/archive/p/jing-trang/downloads

It moved to https://github.com/relaxng/jing-trang after Google Code shut down a few years ago, and has seen some recent development activity on GitHub.

PR #198, opened 2020-05-25
CI: RNC schema validation in via Java trang

dhimmel:
refs https://github.com/citation-style-language/schema/issues/197

[DO NOT MERGE YET]

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -3,7 +3,8 @@ on:
 - push
 - pull_request
 jobs:
-  test:
+  csl-json:
+    name: CSL JSON Tests
     runs-on: ubuntu-latest
     steps:
     - name: Checkout Repository
@@ -14,3 +15,15 @@ jobs:
       run: pip install --requirement=tests/requirements.txt
     - name: Run Tests
       run: pytest --verbose --color=yes tests
+  csl-xml:
+    name: CSL XML Tests
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v2
+    - name: Set up JDK
+      uses: actions/setup-java@v1
+    - name: Install dependencies
+      run: apt-get install --yes trang
+    - name: Run Tests
+      run: java -jar trang.jar csl.rnc csl.rng
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -26,6 +26,6 @@ jobs:
       with:
         java-version: '11'
     - name: Install dependencies
-      run: apt-get install --yes trang
+      run: sudo apt-get install --yes trang
     - name: Run Tests
       run: java -jar trang.jar csl.rnc csl.rng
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -14,3 +15,17 @@ jobs:
...
+    - name: Checkout Repository
+      uses: actions/checkout@v2
+    - name: Set up JDK
+      uses: actions/setup-java@v1
+      with:
+        java-version: '11'
+    - name: Install dependencies
+      run: sudo apt-get install --yes trang
+    - name: Run Tests
+      run: java -jar trang.jar csl.rnc csl.rng
```
dhimmel, line 31: With this version, we get the CI error:

```
Run java -jar trang.jar csl.rnc csl.rng
Error: Unable to access jarfile trang.jar
##[error]Process completed with exit code 1.
  Post Checkout Repository
```

That makes sense because `sudo apt-get install --yes trang` is not making `trang.jar` available.

Also it seems that `sudo apt-get install --yes trang` may be adding an additional java version, so it may duplicate efforts with `actions/setup-java@v1`.

@bdarcus do you want to look into this and take it from here? It's just a matter of getting access to the jar file (or an alternative run command) for how we install it.

bdarcus:
I should be able to figure it out. It's probably just replacing "Java -jar trang.jar" with just "trang," because most distros include a wrapper script.

.github/workflows/test.yaml
bdarcus: I just edited test.yaml per the below. I'm good if the checks pass?

dhimmel: > I just edited test.yaml per the below. I'm good if the checks pass?

Yeah in general. Since this is the first time running it, you might want to look at the logs and see that all the output looks okay.

I don't see any problems on a quick glance.

bdarcus: On this:

> Also it seems that sudo apt-get install --yes trang may be adding an additional java version, so it may duplicate efforts with actions/setup-java@v1.

So should we try removing the latter?

Java should be a dependency of trang, so that apt would indeed install it.

dhimmel: You could remove the `uses: actions/setup-java@v1` step. It would speed up build times a tiny amount.

The main downside I think to the installation via apt-get, is that you might not get the most recent version of `trang`. Not sure if `trang` is actively updated or whether you want the newest version.

rmzelle: > The main downside I think to the installation via apt-get, is that you might not get the most recent version of `trang`. Not sure if `trang` is actively updated or whether you want the newest version.

It's probably fine as is.

It looks like https://github.com/citation-style-language/schema/runs/707588711 uses V20151127, which is two versions behind V20181222 (https://github.com/relaxng/jing-trang/releases). I can't locate a changelog, but RELAX NG itself hasn't changed in decades, and Trang was always robust. The only area we might expect some differences is in support for (embedded) Schematron schema code, I think.

rmzelle:
Thanks!

Issue #167, opened 2019-02-19
Expanded -short, original-, reviewed- variables


bwiernik:
Recently, in response to user needs for the SBL style, citeproc-js added support for providing `-short` versions of all CSL variables, to be rendered with `form="short"`. (https://forums.zotero.org/discussion/comment/324592/#Comment_324592)

In my work on apa.csl, I'm finding that it wants a lot more detailed information for reviews (e.g., medium of item being reviewed, date of item being reviewed), as well as for original publication information (e.g., original medium, original container title, original pages, original editor) than is currently possible with existing CSL variables. As far as I am aware, MLA and Chicago have similar requirements.

I suggest that `-short`, `original-`, `reviewed-` should be expanded so that they can be applied to any CSL variable. This would allow maximum flexibility without having to individually specify each possible variable of this kind.

denismaier:
Sounds good to me. Any potential drawbacks when this is implemented?

bwiernik:
These could be implemented in CSL-JSON as arrays for `short`, `original`, `reviewed`: https://github.com/citation-style-language/schema/issues/169

denismaier:
Hmm. @dstillman Looks Zotero devs are not big fans arrays and objects. Suggestions concerning the data structure here?

dstillman:
> These could be implemented in CSL-JSON as arrays for short, original, reviewed.

I'm not clear what the suggestion is here. Can you give an example?

bwiernik:
I should have said objects, not arrays. Example:

```
"reviewed": {
  "type": "motion_picture",
  "medium": "DVD",
  "title": "Title of reviewed movie"
}
```

vs listing these as individual `reviewed-` variables.
```
reviewed-type: motion_picture
reviewed-medium: DVD
reviewed-title: Title of reviewed movie
```

bdarcus:
So your wanting to change a mostly flat (aside from contributors and dates) data model to a more structured one.

denismaier:
Ultimately, this could lead to a data model as outlined [here](https://github.com/citation-style-language/csl-evolution/issues/12#issuecomment-617029729).

dstillman:
[...]

PR #203, opened 2020-05-27
Split title -> Title-main/title-sub

denismaier:
Closes https://github.com/citation-style-language/csl-evolution/issues/31

Updated proposal:

# Summary

Adds variables and attributes for splitting and re-assembling titles

This adds distinct variables for different parts of a title, and attributes that instruct processors how these variables can be derived from regular title variables, how titles should be re-assembled from these derived variables, and which delimiters to use for that. 

It also adds a new option "capitalize-subtitle" to `text-case`. 

# Suggestion for spec addition
To be added ...

# Requirements

CSL processors currently accept titles as a single string, and CSL styles can only define formatting for the title as a whole. 

This leads to the following (documented) problems regarding style agnostic data entry, which this proposal addresses:
1. delimiters between title and subtitle can't be changed on a per style basis.
2. delimiters between multiple subtitles can't be changed on a per style basis.
3. capitalizing the first word of a subtitle can't be applied on a per style basis.
4. title and subtitle can only be formatted as a whole.

# Proposed solution

Build upon existing functionality in citeproc-js to split `title` variables into `title-main` and `title-sub` variables, and provide ways to configure the splitting and re-assembling mechanism, and to define delimiters.

This is probably the solution that has the least impact on existing styles and reference management tools, it is consistent with existing behaviour, it does not break existing workflows.

So, we have to decide:

1. How to render `title` if the data provides `title-main` and `title-sub` (what delimiters to use)
2. How to split `title` into `title-main` and `title-sub`
3. Whether and how to normalize delimiters

There are several variations across styles that need to accommodated:

1. Styles vary in what delimiters are used to split (e.g., APA splits on `—`, Chicago does not; Chicago splits on `; or,` and `, or,` but most other styles do not)
2. Some styles (e.g., Chicago) normalize delimiters, other (e.g., APA) do not; also there's some variation across styles which delimiters should be normalized.
3. Styles vary in whether second+ subtitles use the same or a different delimiter (e.g., `;` instead of `:`).

With these in mind, here is the proposal:
1. **Variables**: Add `title-main` and `title-sub` variables. Create a new variable type `variables.titles` in `csl-variables.rnc`
2. **Data structure:** In CSL JSON, in addition to `title` allow `title-main` and `title-sub`. Both are flat strings. Citeprocs should parse and normalize delimiters between multiple subtitles from `title-sub`. Multiple subtitles can be explicitly provided by separating them with two vertical bars: `First subtitle || Second subtitle`.
3. **Parsing by citeproc:** If `title-main` and `title-sub` are not supplied in the data, the citeproc will derive them from `title` following these rules (based on existing citeproc-js behavior):

  1. Split the title based on the inheritable attribute `title-split`. Options for this attribute are:   
    * `"simple"`: Matches `. `, `: `, `:: `, `! `, `? `
    * `"extended"`: Matches the values for `"simple"`, plus `; `
    * `"full"`: Matches the values for `"simple"`, plus `—` and `; `
    * `"chicago"`: Matches the values for `"simple"`, plus `; ` and `[;,] or[,:]`

  2. `title-main` is the text before the first delimiter. `title-sub` is an array of the text following each delimiter. (How such an array should be entered is still undecided.)

  3. If both `title` and `title-main` are provided and `title-main` matches the beginning of `title`, only split the delimiter following the `title-main` text if any. (This provides an override to prevent the standard splitting behavior if needed.) The same applies for `title-sub`: If both `title` and `title-sub` are provided, only split the delimiter following the first subtitle if any.

 4. **Rendering title:**: `<text variable="title">` renders `title-main` followed by `title-sub`.
    * The delimiter is set by the inheritable attribute `title-delimiter`, a text string. `title-sub` renders all of its elements, with the delimiter set by the inheritable attribute `title-sub-delimiter` (the same as `title-delimiter` if empty).
    * Don't add another delimiter if one is already at the end of the previous title chunk (e.g., don't add ":" to the end of 'title-main` "Dude, where's my car?").

5. **Normalizing delimiters:** Delimiters are changed to the values specified by `title-delimiter` and `title-sub-delimiter`. Which delimiters are normalized is controlled via an inheritable attribute `normalize-title-delimiters`. Options for this attribute are:
    * `"simple"`: Normalize `. `, `: `, `:: `
    * `"extended"`: Normalize the values for `"simple"`, plus `; `
    * `"full"`: Normalize the values for `"simple"`, plus `—` and `; `

6. Splitting and normalization behavior regarding dashes: As some locales prefer en to em dashes citeprocs should check against both if the "full" options are selected on `normalize-title-delimiters` and/or `title-split`.

7. **Text casing:** When calling `<text variable="title">`:
    - `text-case="title"` title casing is applied to `title-main` and each element of `title-sub`
    - `text-case="capitalize-subtitle"` the first letter of `title-main` and each element of `title-sub` is capitalized
    - Other text cases have the current behavior.

8. **Other titles:** All of the above apply to all `title` variables (`container-`, `collection-`, `volume-`, etc.)

----

Original proposal

I think `title-main` and `title-sub` should be adopted. I think that the citeproc should split `title` if that is all that is provided. 

With that in mind, several behaviors have to be defined:
1. How to render `title` if the data provides `title-main` and `title-sub` (what delimiters to use)
1. How to split `title` into `title-main` and `title-sub`
1. Whether and how to normalize delimiters 

There are several variations across styles that need to accommodated:

1. Styles vary in what delimiters are used to split (e.g., APA splits on `—`, Chicago does not; Chicago splits on `; or,` and `, or,` but most other styles do not)
1. Some styles (e.g., Chicago) normalize delimiters, other (e.g., APA) do not
1. Styles vary in whether second+ subtitles use the same or a different delimiter (e.g., `;` instead of `:`).

<details>
<summary>With these in mind, here is my proposal:</summary>

1. Add `title-main` and `title-sub` as variables.
1. **Data structure:** In CSL JSON, `title` can be provided as an array with `full`, `main`, and `sub` elements. `sub` can be be an array with multiple subtitles. `full` should always be specified to 
1. **Parsing by citeproc:** If `title-main` and `title-sub` are not supplied in the data, the citeproc will derive them from `title` following these rules (based on existing citeproc-js behavior):
    1. Split the title based on the inheritable attribute `title-split`. Options for this attribute are:
        - `"simple"`: Matches `. `, `: `, `:: `, `! `, `? `
        - `"full"`: Matches the values for `"simple"`, plus `—` and `; `
        - `"chicago"`: Matches the values for `"simple"`, plus `; ` and `[;,] or[,:]`
        - (Not exactly sure about `; `. APA, e.g., doesn't clearly mention it, but every use a semicolon in a title I have seen is to delimit a subtitle.)
        - (Possibly could just have `"regular"` [`"full"` above] and `"chicago"` if that would suffice.)
    1. `title-main` is the text before the first delimiter. `title-sub` is an array of the text following each delimiter.
    1. If both `title` and `title-main` are provided and `title-main` matches the beginning of `title`, only split the delimiter following the `title-main` text if any. (This provides an override to prevent the standard splitting behavior if needed.)
1. **Rendering title:** `<text variable="title">` renders `title-main` followed by `title-sub`. 
    - The delimiter is set by the inheritable attribute `title-delimiter`, a text string. `title-sub` renders all of its elements, with the delimiter set by the inheritable attribute `title-sub-delimiter` (the same as `title-delimiter` if empty). 
    - Don't add another delimiter if one is already at the end of the previous title chunk (e.g., don't add ":" to the end of 'title-main` "Dude, where's my car?").
    - Child elements `<main>` and `<sub>` control text styling for each part separately (e.g., ABNT rules about `title-main` in bold but `title-sub` in regular face).
    - This might warrant a new `cs:title` element sibling to `cs:text`.
1. **Normalizing delimiters:** If the inheritable attribute `normalize-sub-delimiter` is true, `. ` and `: ` delimiters are changed to the values specified by `title-delimiter` and `title-sub-delimiter`. Other delimiters are left unchanged.
1. **Text casing:** When calling `<text variable="title">`:
    1. `text-case="title"` title casing is applied to `title-main` and each element of `title-sub`
    1. `text-case="capitalize-subtitle"` the first letter of `title-main` and each element of `title-sub` is capitalized
    1. Other text cases have the current behavior.
1. **Other titles:** All of the above apply to all `title` variables (`container-`, `collection-`, `volume-`, etc.)

</details>

<details>
<summary>Links to relevant citeproc-js behavior:</summary>

  - [here](https://github.com/Juris-M/citeproc-js/blob/5bbf98b9ed534d778d72e8f8257894b580feee90/src/load.js#L740)
  - [here](https://github.com/Juris-M/citeproc-js/blob/5bbf98b9ed534d778d72e8f8257894b580feee90/src/load.js#L826)
  - [here](https://github.com/Juris-M/citeproc-js/blob/5bbf98b9ed534d778d72e8f8257894b580feee90/citeproc.js#L874)
  - [here](https://github.com/Juris-M/citeproc-js/blob/5bbf98b9ed534d778d72e8f8257894b580feee90/citeproc.js#L1236)

</details>

_Originally posted by @bwiernik in https://github.com/citation-style-language/csl-evolution/issues/31#issuecomment-625514420_


denismaier:
@bdarcus Do you see why the test fails here? There seems to be a syntax error on line 774, character 14, but I don't see why that shouldn't work.

bdarcus:
I can't see anything wrong. Also, seems odd that the latest commit doesn't touch line 774.

What editor are you using for the rnc editing? Seems it engaged it some significant modification of whitespace, which is showing up on the git diff?

bdarcus:
Did you try running trang locally?

bdarcus:
Note: this is an example of one of the problems having the documentation repo separate from this one. Really we need to be able to evaluation PRs that impact the spec alongside the spec language.

bdarcus:
Question: since this makes a recommendation on csl-data.rnc, which I think may be controversial, should this PR also include the proposed changes to that file?

denismaier:
Error found. The error were the wrong braces on line 778. Fixed now. (The editor is VS Code.)

denismaier:
> Note: this is an example of one of the problems having the documentation repo separate from this one. Really we need to be able to evaluation PRs that impact the spec alongside the spec language.

Perhaps we should do things the other way round? Write documentation first, schema second?

denismaier:
> Question: since this makes a recommendation on csl-data.rnc, which I think may be controversial, should this PR also include the proposed changes to that file?

Sure. I'll look into it, but need to understand csl-data.rnc first. How do you transform title into an array. @bwiernik hints?

bdarcus:
> Perhaps we should do things the other way round? Write documentation first, schema second?

I suspect that would end introducing the same problem.

I'm seriously wondering if we should move the spec files here, and simply include them in the documentation repo as a submodule. The documentation repo would then effectively be more of a publishing repo.

Just mentioning it while I think about it.

bdarcus:
> > Question: since this makes a recommendation on csl-data.rnc, which I think may be controversial, should this PR also include the proposed changes to that file?
> 
> Sure. I'll look into it, but need to understand csl-data.rnc first. How do you transform title into an array? @bwiernik hints?

So you are proposing to turn ...

``` json
"title": "Some Time: With Subtitle"
```

... into:

``` json
"title": {"title-main": "Some Time", "title-sub": "With Subtitle"}
```
And the first example would thus be deprecated?

If yes, why not simply?

``` json
"title-main": "Some Time", 
"title-sub": "With Subtitle"
```

.. and retain "title"?

bdarcus:
Also, should this go in master, or the v1.1 branch?

bdarcus:
> > Perhaps we should do things the other way round? Write documentation first, schema second?
> 
> I suspect that would end introducing the same problem.
> 
> I'm seriously wondering if we should move the spec files here, and simply include them in the documentation repo as a submodule. The documentation repo would then effectively be more of a publishing repo.
> 
> Just mentioning it while I think about it.

DocBook has the spec maintain within a subdir of the [integrated schema+spec repo](https://github.com/docbook/docbook).

denismaier:
> DocBook has the spec maintained within a subdir of the [integrated schema+spec repo](https://github.com/docbook/docbook).

So does [TEI](https://github.com/msDesc/consolidated-tei-schema).

denismaier:
@bwiernik What do you think so far? And what do you think regarding Bruce's questions? What I was also thinking of: If `title-sub` is again an array, then this means that we need another optional `sub` element under the `sub` element under `cs:title`. This should be possible:

```xml
<title>
  <main/>
  <sub>
    <sub/>
  </sub>
<title>
```  

Note also that in the current version `normalize-title-delimiters` and `title-split` are completely in sync. `title-split="never"` is still there even if that is probably not necessary.

I'm not sure about the dashes yet, but semicolons should be normalized, therefore `normalize-title-delimiters="chicago"`. And I've also included a full variant where dashes are normalized as well.

bwiernik:
In the .rnc, make a new section that is variables.titles and move the title variables there. 

In the .json schema for data, specify a definition for titles like for names or dates. The slots for that would be `main`, `sub`, `full` (or `literal` or `raw`) (not `title-main`, etc.). Set all the title variables to be either a string or an array with that definition.

csl-variables.rnc
```diff
@@ -53,6 +54,30 @@ div {
     | "page-first"
     | "volume"
   
+  ## Title variables
+  variables.titles =
+    "collection-title"
+    | "collection-title-main"
```
bwiernik, line 60: The main and sub versions don't need to be here. They could just be supplied as part of the data. Similar to how we don't have `issued-year`, `issued-month`. If someone wants to render just part of the title, they should use `cs:title`.

csl.rnc
```diff
@@ -717,6 +718,71 @@ div {
        attribute variable { variables.standard },
        [ a:defaultValue = "long" ] attribute form { "short" | "long" }?)
 }
+[ a:documentation [ xhtml:h2 [ "cs:title Rendering Element" ] ] ]
+div {
+  ## Use to render titles if "title-main" and "title-sub" need different formatting
```
bwiernik, line 723: Add `## or to omit either "title-main" or "title-sub"`

csl.rnc
```diff
@@ -717,6 +718,71 @@ div {
...
+      ## Don't split titles
+      "never" 
+      | 
+      ## Matches "."", ":", "::", "!", "?" 
+      "simple" 
+      | 
+      ## Matches the values for "simple" plus "—" and ";"
+      "full" 
+      | 
+      ## Matches the values for "simple" plus ";" and ";," or ",:"
```
bwiernik, line 746: Need to double check the meaning of ` ";," or ",:"` in the Chicago manual. I remember in my original proposal I had included in them in square brackets, but I don't recall the details of what I was thinking.

csl-variables.rnc
denismaier: So I move them back to the regular variables? Also the short versions stay with the string variables?

csl.rnc
denismaier: Thanks.

denismaier:
> And the CSL JSON change?

Comes next. I am just thinking if I should add `sub` as an array or if a three component solution would be better: `main`, `sub`, `addon`. That would allow for different formatting of the last part, which is sometimes required. (Or you can even omit certain parts of the title if you want. Could be useful for Festschriften etc.)
And, should we add someting for the delimiters? I was thinking about this structure:

```json
"title-variable": {
	"anyOf": [
		{
			"properties": {
				"main": {
					"type": "string"
				},
				"sub": {
					"type": "string"
				},
				"addon": {
					"type": "string"
				},
				"sub-delimiter": {
					"type": "string"
				},
				"addon-delimiter": {
					"type": "string"
				},
				"literal": {
					"type": "string"
				}
			},
			"additionalProperties": false
		}
	]
}
```

Am I overthinking this?

Then, what about the `short` forms?

bwiernik:
What is `addon`?

csl.rnc
bwiernik: Got it. `or` is part of the delimiter.

The additional Chicago delimiters are:
- `;`
- `; or,`
- `, or:`

Expressed as a regular expression, the last two are together: `[;,] or[,:]`

csl-variables.rnc
```diff
@@ -88,6 +73,11 @@ div {
     | "authority"
     | "call-number"
     | "citation-label"
+    | "collection-title-main"
```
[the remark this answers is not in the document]
bwiernik: It's part of the data structure in the CSL JSON. This would work the same way as for a date. We don't specify `issued-year` as a separate variable. `year` is just a property of the `issued` array.

A title might be supplied as a string or as an array containing main, sub, etc.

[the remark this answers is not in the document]
denismaier: Ok, have removed the spurios `-main` and `-sub` variables.
But again, how can users supply the main part if the heuristics fail?

csl.rnc
denismaier: I can't follow. How can `or` be part of the delimiter? Can you spell this out in an example?

csl-variables.rnc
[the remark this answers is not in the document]
bwiernik: In the CSL JSON data.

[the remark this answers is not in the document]
bwiernik: Oh, I see that case. Let me think.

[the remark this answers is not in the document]
denismaier: I mean: in pandoc it won't be a problem. But what about Zotero?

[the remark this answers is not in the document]
denismaier: Ok, this here does work for issued: I have a date with 2019 in the regular `date` field, and this here in `extra`: `issued: ["year": "2020"]. Result is 2020. Not a pretty solution, but it works.

denismaier:
> `addon`: biblatex has a field three part structure for titles:
> title, subtitle, titleaddon.
> 
> So I was thinking if we should do use`title-main`, `title-sub`, `title-addon` rather than having `title-sub` as an array with unlimited strings. Users could then specify each part of the title. If `sub` is a array, how would you access, format or suppress the second part of the subtitle?

@bwiernik What do you think about `addon` vs. `sub` as an array?

bwiernik:
I don't think that `addon` and `sub` are conceptually the the same thing. `addon` would be something like "a novel" or "Official Journal of the American Cardiac Society"—something that wouldn't generally be included in references. "High School Musical: The Musical: The Series" just has two subtitles, both of which should included in a citation.

bwiernik:
I guess that multiple subtitles is a rare enough case that we can probably just let `sub` be a flat string and always rely on the heuristic splitting.

denismaier:
Ok, so no `addon`. Need to think about `sub` as an array or string. 

What about `short`? 
And what about the delimiters? Do you think they should be part of the array?

csl.rnc
denismaier: Bump

bdarcus:
I'm still improving my knowledge of git.

Note for @bwiernik and @denismaier:

I earlier said to update branch code using `git merge master` but I believe that should be `git rebase master`.

More:

https://spin.atomicobject.com/2019/01/10/git-merging-vs-rebasing/

bwiernik:
@bdarcus What exactly are you trying to do? Generally, rebasing isn’t necessary (and can cause issues). 

@denismaier I think `addon` would be fine to include (probably not necessary, but not a problem). Just that it’s different from sub. `short` in the array seems fine.

bdarcus:
> @bdarcus What exactly are you trying to do? Generally, rebasing isn’t necessary (and can cause issues).

I created a broader issue at #215 to try the nail this down. I may be overthinking this, but git just sometimes confuses me.

denismaier:
@bwiernik 
Ok, so I'll add a maximalist version: i.e., `short` and `addon` to the definition, `sub` stays an array. I'll also add the delimiters. (We can still remove them later if this is too much.)

I am still not sure how users will override wrong heuristics. My solution with `issued` was probably just plain luck.

And, in case you haven't seen it, do you have an example for the chicago case (`[;,]` or `[,:]`)?

denismaier:
`short` and `addon` added.
But no delimiters for the moment: if we also always specify full, we can use heuristics to get the delimiters.

bwiernik:
Examples of `or` subtitles:
> The Tempest, or The Enchanted Island
> Moby-Dick; or, The Whale
> Dr. Strangelove, or: How I Learned to Stop Worrying and Love the Bomb

denismaier:
Now that `short` is in the array: should we remove ` title-short` etc?

bwiernik:
There is a lot of CSL data out there already with `title-short` and `shortTitle` in it already, so I think these need to be available.

bdarcus:
While I guys are working on this, on the json schema, will this
require a new version, and hence new ids?

Related to #216.

denismaier:
> There is a lot of CSL data out there already with `title-short` and `shortTitle` in it already, so I think these need to be available. But I'm not sure how that needs to be represented in the RNC vs the data schema?

There's currently no `shortTitle` in the RNC, it's only in the json. Not sure what that means...

denismaier:
@bwiernik I have removed Option "chicago" from "normalize-title-delimiters". Now that I've seen your examples, it's clear that this option does not make much sense here.

Open question:
What about the options "never". Should I keep that, or can we rely on `title-delimiter` being set as an indicator?
That would mean the defaults would be:
title-split => simple
normalize-title-delimiters => simple (?)

Both would only have visible effects with `title-delimiter` set to something, or with `textcase="uppercase-subtitle"`.

bwiernik:
> While I guys are working on this, on the json schema, will this require a new version, and hence new ids?

I think it will require a new version, since the structure of title variables is changing to permit an array instead of just a flat string.

> There's currently no shortTitle in the RNC, it's only in the json. Not sure what that means...

`shortTitle` was added by Frank when he created the data schema for citeproc-js, before it was formally incorporated 

My understanding is that things only need to be a part of the RNC if they are "first class" variables that could be directly called by a style? Looking at the styles repo, `title-short` is called explicitly in 138 styles. Most of those could be replaced automatically. But, it does seem important in a few legal citation styles to be able to determine if a `legal_case` has a short title or only a long title, so I suggest we keep `title-short` in the RNC. `container-title-short` is only called explicitly in 26 styles; all of those cases could (and should) be replaced with `variable="container-title" form="short"`. So, I think the explicit `-short` forms of other titles could be removed without much issue, so long as they are accessible in the data schema. 

So:
  - `title-short` seems like it should be kept in the RNC
  - other `-short` titles could be removed from the RNC or kept; I don't have a strong preference. The necessary update of removing them would be minor.
  - `-short` title values (and `shortTitle` and `journalAbbr`) should be available in the data schema.
  - We could also permit `-main` and `-sub` forms of the title variables in the data schema for easier user entry if the implementing client doesn't provide an interface for producing a title array.

> What about the options "never". 

It seems like it isn't necessary, so long as `title-delimiter` isn't set on the style.

denismaier:
> * We could also permit `-main` and `-sub` forms of the title variables in the data schema for easier user entry if the implementing client doesn't provide an interface for producing a title array.

Ok, we should do this.

@dstillman What do you think about this? Any chance Zotero could provide an interface for producing a title array?

denismaier:
Concerning `-sub` forms in the data schema: That should probably only define the first subtitle in a subtitle array, so to provide a way for overriding the heuristics here as well... Is that reasonable?

dstillman:
> Any chance Zotero could provide an interface for producing a title array?

What's an example where this would be used by the user, rather than just letting citeproc parse a single title string?

(There've probably been others, but the first Zotero forum post that I can recall that seems to perhaps be requesting this was actually [just posted today](https://forums.zotero.org/discussion/83477/web-page-title), though they haven't responded, so it's not entirely clear what they're asking for.)

denismaier:
The main use case is to override automatic split points. E.g. when a dash is part of the main title and not a delimiter between title and subtitle.

dstillman:
So the default would still be a sole title field passed as a string, and this would need to be an alternative entry mode? That would be a pretty significant UI and data-model change, so I'm not super eager to do that without a lot of demand for it, which, again, I don't recall from the Zotero Forums. It would be possible to use the Extra field for it, regardless. If Zotero switches to rich-text markup for titles, there might also be a creative way to allow different parts to be marked up and then represent it with HTML in what would still be a single underlying field.

On a technical note, I'm a bit confused by at least some of the uses of the term `array` here. Are we talking about an array/list, or are we talking about an object/dictionary?

bwiernik:
An array/list—a title could be structured as an array, like a date or name is.

bwiernik:
This discussion arose because of requests from primarily Brazilian researchers on the Zotero Forums for finer control over title and subtitle formatting. 

I imagine that a switch to data structure in Zotero would only occur as part of a switch to a rich text title editor (e.g., there could be one-field/two-field button like with names). In the interim, users needing fine control over title splitting could enter title-main and title-sub in Extra. For the vast majority of users, the heuristic splitting of titles would work. 

Regarding the subtitle, I’m not exactly sure the structure. There exist items with multiple subtitles, but they are rare, and I think that relying on heuristic formatting would be fine.

denismaier:
> An array/list—a title could be structured as an array, like a date or name is.

Are you sure? 
Exporting a title gives me this for a name:

```json
"author": [
	{
		"family": "Sühl-Strohmenger",
		"given": "Wilfried"
	}
],
```
So, `author` is an array consisting of multiple author objects? Right?

For `issued`:

```
"issued": {
	"date-parts": [
		[
			2019
		]
	]
},
```

Issued is an object, the key `date-parts` consists of two nested arrays.

Am I wrong here?

Anyway, I think the current proposal will give us this as data input:

```json
"title": {
	"full": "A title: with a subtitle: and a second subtitle",
	"main": "A title",
	"sub": ["with a subtitle","and a second subtitle"],
},
```

Is that it?

denismaier:
@bwiernik I think we're almost there. I've added the `title-main` and `title-sub` etc.
I've also removed "never" from the options. New default is "simple". Ok?

I'm just asking myself what should be done with semicolons. Add them to option "simple"? Or add a new option "extended" between simple and full:

```
## Normalize the values for "simple" plus ";"
      "extended"
```

denismaier:
@dstillman Concerning the UI: One additional requirement is that you'll need to provide the complete title as a literal string in addition to the individual parts. That's necessary for the extraction of the delimiters.
I can imagine using a one-field/two-field solution, tagging the titles (or the breakpoints), or using `title-main` etc. in the `extra` field. Either way, we assume that the heuristics should work for most users in most cases, and you shouldn't need to make title structures explicit too often, no matter which solution is finally implemented.

denismaier:
Another thing I'm a bit worried about: Will we still allow simple string input for `title` in a plain text systems like pandoc?
Will this be possible?

```
title: A very simple title
```

Or will users need to do (at least):
```
title: 
  full: A very simple title
```

Or is it up to implementers to allow both variants?

bwiernik:
I might have misunderstood the definitions here. My bad. I guess I meant that title would be an object, with sub potentially being an array:
```
"title": {
	"full": "A title: with a subtitle: and a second subtitle",
	"main": "A title",
	"sub": ["with a subtitle","and a second subtitle"],
},
```

bwiernik:
I imagine most input of titles would be flat strings (whether entered in JSON or in a plain text format like pandoc's CSL YAML):
`title: A simple title`
`title: A title: with a subtitle`

The object structure being for more nuanced control. Processors would need to implement both variants. A client like Zotero could implement both or just the object format. 

I'd like @dstillman to weigh in some more on the client side.

denismaier:
> The object structure being for more nuanced control. Processors would need to implement both variants. A client like Zotero could implement both or just the object format.

What does that mean for the json schema?

denismaier:
@bwiernik Have you seen the other question about "never", and the semicolon? (Sorry, I don't want to be pushy, but sometimes such questions get lost in these long threads.)

bwiernik:
> What does that mean for the json schema?

I think it would be something like this, but someone would need to verify this is valid:
```
            "issue": {
                "type": [
                    "string",
                    "array"
                ],
                "items": {
                    "$ref": "#/definitions/title-variable"
                }
            },
```

@dstillman Implementation-wise, would just expecting separate `title`, `title-main`, `title-sub` string variables in the data be better? I think that we could accomplish most of the parsing needs with that. The only limitation would be `title-short` for items with multiple subtitles. `title-sub` could be defined as an array to accommodate that?

dstillman:
> Implementation-wise, would just expecting separate title, title-main, title-sub string variables in the data be better?

Yes, because, again, I'm not sure we'll actually implement this, or do so anytime soon. If they're just string variables, the variables will work in Extra.

bwiernik:
> Removing 'never'

Agreed.

> normalize semicolons?

Where did we land on normalizing dashes? I think the correct behavior would be `"full"` normalizing semicolons, but not dashes.

denismaier:
> > Removing 'never'
> 
> Agreed.

Good, then I'll keep that.
> 
> > normalize semicolons?
> 
> Where did we land on normalizing dashes? I think the correct behavior would be `"full"` normalizing semicolons, but not dashes.

We've first settled on not normalizing dashes. But after finding the SBL example, I think we should normalize dashes with "full". If my understanding of their rule is correct they leave dashes in place if they are part of the title (or a part of the title (`title-main` or `title-sub` in our language), but they consistently use colons as delimiters between title and subtitle. ([Here ](Doing Gender—Doing Religion)is an example where they keep a dash that is part of `title-main`, but I have yet to find one where they actually replace a delimiter dash.)

So I'd propose we have three options:
simple=> period, colon
more than simple, less than full=> period, colon, semicolon
full=> period, colon, semicolon, dash

bwiernik:
Okay, so the "full" would really only normalize dashes if data were supplied with explicit main and sub titles. That sounds find to me. Go with `extended` for the intermediate option.

Based on Dan's feedback, I think we should forego the object data structure and just allow the title subparts to be provided as strings in the data schema. That makes everything easier all around I think. As a related issue, how will https://github.com/citation-style-language/schema/issues/167 be implemented in the RNC and data schemas?

denismaier:
I thought we do it the other way round: "full" will normalize dashes. With the "Doing gender" example above you just add `title-main` to not normalize the dash. I think that would be more consistent with the way we override the heuristics in other cases, e.g. when a colon is not a delimiter.

bwiernik:
Sorry, that wording was unclear. I was agreeing with your recommendation. My comment "would really only normalize dashes" meant " 'full' would only really have any consequence if data were supplied with explicit title parts".

denismaier:
> meant " 'full' would only seem to really have any consequence if data were supplied with explicit title parts".

I'm still not sure about this. I don't think "full" should rely on explicit title parts. My understanding is it will work like this:

```
title: Doing Gender—Doing Religion: blabla
```

Now, with "full" citeprocs will do this:

```
title: 
  full: Doing Gender—Doing Religion: blabla
  main: Doing Gender
  sub: 
  - Doing Religion
  - blabla
```

So users will provide `title-main` explicitly:

```
title: Doing Gender—Doing Religion: blabla
title-main: title: Doing Gender—Doing Religion
```

This will provide the necessary hints for the splitting mechanism, and citeprocs do this:

```
title: 
  full: Doing Gender—Doing Religion: blabla
  main: Doing Gender—Doing Religion
  sub: 
  - blabla
```

Are we on the same page here?

denismaier:
Ok, should mainly be done now. I'll update the description and the proposal tomorrow.

Just: Now that the data structure is flat again, how are we going to deal with multiple subtitles? Three options:
1. Do nothing.
2. Add another sub variable `title-sub-sub`.
3. Citeproc's should internally convert subtitles to an array. An explicit `title-sub` variable will define only the first subtitle. => Users will be able to override the heuristics if necessary. This should be sufficient as it's rather unlikely that there are more than two subtitles.

I'd opt for option 3.

bdarcus:
You might turn draft off now, so the CI will run.

denismaier:
Updated the proposal, according to what we've agreed upon.

Last open question are:
- How to deal with multiple subtitles (as per [above](https://github.com/citation-style-language/schema/pull/203#issuecomment-639129277))
- I've added a note about dashes: "Splitting and normalization behavior regarding dashes: As some locales prefer en to em dashes citeprocs should check against both if the "full" options are selected on `normalize-title-delimiters` and/or `title-split.`" Ok so?

Finally, where should this go? 1.0.2 or 1.1?
As it is not backwards incompatible we could theoretically add it to 1.0.2. On the other hand, it's a slightly bigger change i terms of user experience, and it will require more work in citeprocs than just adding a couple of variables and terms; so 1.1. might be more appropriate...

bdarcus:
You have conflicts. Try "git merge master" and see if that resolves?

bdarcus:
> As it is not backwards incompatible we could theoretically add it to 1.0.2. On the other hand, it's a slightly bigger change i terms of user experience, and it will require more work in citeprocs than just adding a couple of variables and terms; so 1.1. might be more appropriate...

I think you answered your own question ;-)

If that's the case, you want to switch the branch this targets.

bdarcus:
I just made the switch. If you're OK with that, I think you'd want to do `git merger v1.1` instead.

denismaier:
Done---I hope the merge was ok. (Perhaps you can have a quick look...)

schemas/styles/csl-variables.rnc
```diff
@@ -5,7 +5,12 @@ namespace a = "http://relaxng.org/ns/compatibility/annotations/1.0"
 div {
   
   ## All variables
-  variables = variables.dates | variables.names | variables.standard
+  variables =
```
bdarcus, line 8: This changed formatting: produced by trang?

schemas/styles/csl.rnc
```diff
@@ -649,6 +650,72 @@ div {
...
 
+## cs:title Rendering Element
+div {
+  ## Use to render titles if "title-main" and "title-sub" need different formatting
+  ## or to omit either "title-main" or "title-sub"
+  rendering-element.title =
+    element cs:title {
+      title.attributes,
+      title.formatting,
+      (title.main?, title.sub?)
```
bdarcus, line 661: Have you tested this on a validating and tag-completing editor like Atom?

I think the way that's defined, this would be valid:

``` xml
<cs:title>
  <cs:main/>
  <cs:sub/>
</csL:title>
```

... but this would not:

``` xml
<cs:title>
  <cs:sub/>
  <cs:main/>
</csL:title>
```

schemas/styles/csl.rnc
```diff
@@ -649,6 +650,72 @@ div {
 }
 # ==============================================================================
 
+## cs:title Rendering Element
```
bdarcus, line 653: So this is a new rendering element? 

Two things:

1. is it _really_ necessary; this can't be done adequately with attributes on cs:text?
2. not sure if this is a technical problem or not, but will note this is the same element name as `cs:info/cs:title`.

schemas/styles/csl.rnc
```diff
@@ -649,6 +650,72 @@ div {
...
+  ## Use to render titles if "title-main" and "title-sub" need different formatting
+  ## or to omit either "title-main" or "title-sub"
+  rendering-element.title =
+    element cs:title {
+      title.attributes,
+      title.formatting,
+      (title.main?, title.sub?)
+    }
+
+  title.attributes =
```
bdarcus, line 664: I'm not sure if the diffs are just getting screwed up, but I can't tell what's really new. if you haven't run trang to update the formatting, please do that and I'll take another look and see if it's easier to see.

denismaier: Ok. have done.

schemas/styles/csl.rnc
denismaier: I have used `cs:names` as a template. That's why the order matters here. But I agree with you that this probably makes things overly complex. I wouldn't mind removing that constraint if there are no compelling reasons for keeping it.

schemas/styles/csl.rnc
denismaier: ad 1: No, it's not necessary. We could also allow children on `cs:text`. Or it could also work with attributes on `cs:text`, but I think that would be way to complex. You'd need to duplicate every possible attribute with a `subtitle` prefix. 
But I'm not sure which one is a bigger change.
2. I don't know, but we could just change it to `cs:titles`? Would be also more appropriate as this will render different titles (`container-title`, `volume-title`, etc.)

denismaier: On your edit: Could work, but it's not ideal as you won't have access to the original delimiters here. And some styles (APA) require that.

bdarcus: I adjusted the formatting of your main comment at top to make this more clear.

From my understanding, 1, 3, and 4 "requirements" can be addressed without new elements. Is that correct?

It is only 2 that adds "complexity"?

bdarcus: > On your edit: Could work, but it's not ideal as you won't have access to the original delimiters here. And some styles (APA) require that.

I'm not following here.

denismaier: Some styles normalize delimiters, other do not. APA for example will not normalize delimiters, i.e. it will render titles with whichever punctuation was there in the first place. In your solution with a group you'll only be able to use the delimiter specified on `cs:group`, but not the original delimiters.

```
<group delimiter=": ">
  <text variable='title-main" font-weight="bold"/>
  <text variable='title-sub"/>
</group>
```

bdarcus: That's helpful. 

But the language you use to explain that sounds like it could be a global style option?

Hoping @adam3smith and @rmzelle can find time to look at this too.

schemas/styles/csl.rnc
```diff
@@ -649,6 +650,62 @@ div {
...
+## cs:title Rendering Element
+div {
+  
+  ## Use to render titles if "title-main" and "title-sub" need different formatting
+  ## or to omit either "title-main" or "title-sub"
+  rendering-element.title =
+    element cs:title {
+      title.attributes, title.formatting, (title.main?, title.sub?)
+    }
+  title.attributes =
```
bdarcus, line 662: So these are the attributes associated with the proposed new `cs:title` element. 

And they are basically instructions for how processors should parse titles strings into structured data.

Right?

I think for all of us, there are two questions:

1. do we really want to do this? why can't we just demand that processors get properly structured data? 
2. if yes, do we need a dedicated element to hold these attributes? For sake of argument, could they not go on `style`, `citation`, and/or `bibliography`? Related to my broader question about elements vs attributes.

bdarcus: I started to try to draft a git commit message that would concisely explain the changes, but got a bit stuck. We can figure out how to resolve once others weigh in.

> Add support for configuration of complex titles
>                                               
> This adds ability to configure title and subtitle rendering independently. 

> It does this by adding 'title-main' and 'title-sub' and relates variables.                                                              
>                                                                           
>  It also adds new elements and attributes that specify ...

denismaier: > So these are the attributes associated with the proposed new `cs:title` element.
> 
> ~And they are basically instructions for how processors should parse titles strings into structured data.~
> 
> ~Right?~ Well, maybe a bit more than that.

Yes, they instruct processors where they should split titles into segments.

> I think for all of us, there are two questions:
> 
> 1. do we really want to do this? why can't we just demand that processors get properly structured data?

Because that would involve talking to and convincing a lot of people (mainly Zotero, Mendeley, and Papers, but basically everyone on the list on citationstyles.org). They would have to change their internal data structure and their UI fields, or do the parsing themselves before they pass the data to the processor. 
What is more, different styles have different opinions about what exactly constitutes a title or where a subtitle begins. So if we really want style agnostic data, then doing this in the processor is the only real option.
Next, there is legacy data and all of this uses titles in an unstructured single field. With these changes, users will just be able to use the same data as before in new ways. (E.g. A scholar in the social sciences who has added titles according to APA standards, i.e. with the original punctuation, can now produce correct references with Chicago author date just by selecting the correct style.)

> 2. if yes, do we need a dedicated element to hold these attributes? For sake of argument, could they not go on `style`, `citation`, and/or `bibliography`? Related to my broader question about elements vs attributes.

Well, we do that as well. If you look at line 956, you'll find a couple of inheritable options that so essentially the same thing. The idea is that you'll use those on `style`, `citation`, `bibliography` if you render titles with `cs:text`. That should be what do in most cases anyway. `cs:title` should really only be needed if you have special requirements, like different formatting for title vs. subtitle.

denismaier: Maybe an idea: The main use case for `cs:title` is the ABNT style. I don't know other details about this style, but your suggestion from above could work, if a couple of requirements are met:
- they use exclusively use one specific character between title and subtitle, e.g. a colon.
- they normalize all the characters they recognize as splitting characters

Also: I don't know about other styles that have similar requirements.

If we do this, we will also have to make `title-main` and `title-sub` available in styles, which they currently are not.

denismaier: @bwiernik What do you think about this here?

bwiernik: The suggestion being don't have a `cs:title` rendering element and instead expose `-main` and `-sub` as first class variables? How does that handle Chicago normalizing punctuation?

denismaier: I don't think it would affect that at all. You can still use `<text variable="title">` to render the complete title with punctuation normalized. You just can't have independent formatting of title and subtitle.

bwiernik: So, for ABNT, it would be:
```
<group delimiter=": ">
  <text variable='title-main" font-weight="bold"/>
  <text variable='title-sub"/>
</group>
```

Would any delimiters in `title-sub` be normalized?

denismaier: Depending on the attributes, yes, I think they should be normalized.

bdarcus: > `cs:title` should really only be needed if you have special requirements, like different formatting for title vs. subtitle.

I'm not following here. You mean a specific kind of "different formatting"; don't you?

bwiernik: So what can't it do?

denismaier: @bdarcus title bold, subtitle normal font

@bwiernik certain combinations of requirements won't work. APA style not normalization of punctuation and ABNT style title formatting (title bold, subtitle normal).

bwiernik: The only people who've ever asked for separate styling, as you say, is ABNT. I think for now we can reasonably go without `cs:title`. If it ever comes up again, we can revisit `cs:title`.

bwiernik: One thing I'd like to verify. Is it okay for the `title-delimiter` and `title-sub-delimiters` to be empty by default on a style? Are empty attributes acceptable? @bdarcus

bdarcus: > Is it okay for the `title-delimiter` and `title-sub-delimiters` to be empty by default on a style? Are empty attributes acceptable?

What exactly do you mean @bwiernik? 

Acceptable to what or whom?

And what do you mean by "by default"?

Certainly this is valid XML;

``` xml
<foo bar=""/>
```

denismaier: Isn't that something citeproc have to take care if? Like "if the attribute is there then ... otherwise ..."?

bwiernik: Currently, the proposal is, if `title-delimiter` is not assigned on the style (that is, not present at all), then no normalization occurs. I guess that is valid, because styles can have `default-locale` absent.

bwiernik:
Keep them separate. We can use that to generate the `-sub` and `-sub` versions.

schemas/styles/csl.rnc
bdarcus: > @bdarcus title bold, subtitle normal font

I'm admittedly pretty tired today, so might just be me, but why wouldn't this allow just that?

``` xml
<group delimiter=": ">
  <text variable='title-main" font-weight="bold"/>
  <text variable='title-sub"/>
</group>
```

denismaier: It does, but with certain limitations as outlined above. But as it turns out these limitations are of only theoretical relevance and won't really matter much. So, yeah, we should go that route.

bdarcus:
> If we do without cs:title, should I reintegrate the title variables into the standard variables list? There's no technical reason for keeping them in a separate list. But it may still be nice to do so.

Sometimes the patterns are there for technical reasons. But other times, just to enhance clarity.

denismaier:
Update:
1. I have removed `cs:title`, and moved attribute descriptions to the inheritable attributes.
2. Main, sub and short variants of all title variables have been added to cs-variables.rnc. I have also tried to consolidate all title variable in `variables.titles`. Maybe we can get handle the title variables differently later on, but this comes close to what a script should produce then, right?

Are we ok with this then?

bdarcus:
Thanks!

Can you please revise the intro summary to reflect where this ended up, thinking, at least in one clearly labeled section, on what the squashed commit message should be?

@rmzelle - could you please take a look at this? I targeted it to 1.1.

denismaier:
I've added a short description.
I've also updated the proposal to reflect the current state of affairs.

schemas/styles/csl-variables.rnc
```diff
@@ -5,7 +5,12 @@ namespace a = "http://relaxng.org/ns/compatibility/annotations/1.0"
...
+    variables.dates
+    | variables.names
+    | variables.numbers
+    | variables.titles
+    | variables.strings
   
   ## Standard variables
   variables.standard = variables.numbers | variables.strings
```
bdarcus, line 16: not aimed at @denismaier, but why do we have this pattern if we don't use it? the "variables" pattern above could, for example

bdarcus: basically, my suggestion is either we remove it here, or we use it above. either way, I can make the change directly.

bwiernik: `variables.standard` is what can be rendered with `cs:text`. It should be restored above under `variables` and `variables.title` should be added to `variables.standard`.

bdarcus: Good catch @bwiernik. 

In that case, please make that change @denismaier, and make sure you run trang again before you commit it (because I think it will change the formatting of that pattern).

denismaier: Thanks for catching this. Will deal with it tomorrow.

denismaier:
we still need to deal multiple subtitles: https://github.com/citation-style-language/schema/pull/203#issuecomment-639129277

Opinions?

bwiernik:
For subtitles, how about this: 
Subtitles can be provided as a flat string. Citeprocs should parse and normalize delimiters between multiple subtitles from this string. Multiple subtitles can be explicitly provided by separating them with two vertical bars `First subtitle || Second subtitle`.

schemas/styles/csl-variables.rnc
denismaier: Done.

denismaier:
> Subtitles can be provided as a flat string. Citeprocs should parse and normalize delimiters between multiple subtitles from this string. Multiple subtitles can be explicitly provided by separating them with two vertical bars `First subtitle || Second subtitle`.

Great. I've added this to the description.

bdarcus:
I think this is ready to go. 

Adding new variables is a minor change, but the additional processing rules mean this points to v1.1.

This also means I'll leave this here for a bit to see if we can get a review from @adam3smith or @rmzelle.

schemas/styles/csl.rnc
```diff
@@ -893,6 +897,48 @@ div {
         "partial-first"
     }?
   
+  ## Options affecting rendering of title variables, for cs:style, cs:citation and cs:bibliography
+  title-inheritable-options =
+    
+    ## Inheritable title option, specify the delimiter between "title-main" and "title-sub".
```
bdarcus, line 903: Is there a (perhaps locale-specific) default?

schemas/styles/csl.rnc
```diff
@@ -893,6 +897,48 @@ div {
...
+    attribute title-delimiter { text }?,
+    
+    ## Inheritable title option, specify the delimiter between multiple subtitles.
+    ## The same as "title-delimiter" if not specified.
+    attribute title-sub-delimiter { text }?,
```
bdarcus, line 908: Same here.

schemas/styles/csl.rnc
```diff
@@ -893,6 +897,48 @@ div {
...
+    
+    ## Inheritable title option, specify how titles are split into "title-main" and "title-sub".
+    [ a:defaultValue = "simple" ]
+    attribute title-split {
```
bdarcus, line 912: what happens if we come across different lists later? is that even possible?

schemas/styles/csl.rnc
```diff
@@ -893,6 +897,48 @@ div {
...
+        ## Matches the values for "simple" plus ";"" 
+        ## and "[;,] or[,:]" (see CMoS 14.91), e.g.  "Moby-Dick; or, The Whale"
+        "chicago"
+    }?,
+    
+    ## Inheritable title option, specify which delimiters are normalized.
+    [ a:defaultValue = "simple" ]
+    attribute normalize-title-delimiters {
+      
+      ## Normalize "."", ":", "::", "!", "?"
```
bdarcus, line 932: since there's duplication here, does this not suggest a need to break out a separate pattern?

schemas/styles/csl.rnc
```diff
@@ -893,6 +897,48 @@ div {
...
+    attribute normalize-title-delimiters {
+      
+      ## Normalize "."", ":", "::", "!", "?" 
+      "simple"
+      | 
+        ## Normalize the values for "simple" plus ";"
+        "extended"
+      | 
+        ## Normalize the values for "simple" plus "—" and ";"
+        "full"
```
bdarcus, line 939: related to point just above, why is "chicago" not an option here?

schemas/styles/csl.rnc
bwiernik: An empty delimiter means that no normalization of the delimiters occurs

schemas/styles/csl.rnc
denismaier: How would that look like?

bdarcus:
```
myatt =	attribute my-attribute { myatt.values }
myatt.values = "one" | "two" | "three"
```

So you'd just move the values into a pattern, and then refer to that.

denismaier:
Oh, I just thought of one case where I'm not sure it is covered by our solution:

1. APA doesn't normalize delimiters. The standard mode of data entry must be entering titles with original punctuation.
2. If there is no punctuation between title and subtitle, APA (like most/all English language style guides) uses a colon.
3. Now imagine you want to use a style guide that also uses original punctuation, but adds a period if there is no punctuation between title and subtitle: How do you change punctuation only where appropriate without producing false positives. 
The problem is: There's no way to know for a processor if the delimiter is present on the title page or not. Right? 

If that is indeed a real problem (I am not sure yet), how could we solve this? Ideas:
- Marking up/escaping original punctuation or marking up/escaping added punctuation. Not pretty. 
- Changing our expectations what should be supplied in an explicit `title-main`. Right now, we assume that users will supply main part of the title without the subtitle **and without the delimiter between title and subtitle**. We could change that to: "If you specify an explicit `title-main`, also include the delimiter between title and subtitle, if any." Processors could then infer if a colon belongs to the title, or if it was added for bibliographic purposes.

What do you think?

Or, are there other options?

bdarcus:
Your "marking up" option: I've not been paying close attention, but did we
ever address related issues that are solved in the BibTeX world with the
{Braces Solution}?

Never been a fan of this, but may be on the easy option.

I'm meaning in general; not specific to this issue.

denismaier:
Oh, the second option would need a new option for normalize-title-delimiters. Something like "only-added-delimiter"

schemas/styles/csl.rnc
denismaier: Argh, these options are actually identical: simple for splitting is not the same as simple for normalizing. Copy-paste error.
While you split titles at those characters, you only normalize periods and colons (not ? and !)

bdarcus: I don't know if this is a bad idea (and sorry if you already considered or mentioned this), but I figure I may as well ask:

What if these values become localized terms, but where you could override locally like so?

``` xml
<style title-sub-delimiter=":?.;">
```

If that could work, this would just be a string.

Or a space-separated list:

``` xml
<style title-sub-delimiter=": ? . ; ::">
```

There's probably disadvantages, but it's very flexible.

denismaier: Not sure about localized, but yeah I've proposed something similar somewhere along the way. Question us probably if reasonable predefined sets will make style coding easier.

One problem again: how to deal with chicago rules for splitting?
@fbennett thoughts?

denismaier: Perhaps this could work
```xml
<style title-sub-delimiter=": ? . ; :: chicago">
```
Or:
```xml
<style title-sub-delimiter=": ? . ; ::" chicago-subtitle-delimiter="true">
```

bdarcus:
> Oh, I just thought of one case where I'm not sure it is covered by our solution ...

I know this is a complicated topic, but I think this PR is getting way too unwieldy.

There are really two distinct issues here:

1. variable additions 
2. the attributes and associated processing rules

Seems 1 is straightforward, and that most of this discussion is about 2, including potential edge cases we maybe (I don't know) shouldn't worry about.

So I see two options to address 2:

1. remove it from this PR and submit a separate one just on that
2. forget about your latest concern ;-)

denismaier:
> [...] most of this discussion is about 2, including potential edge cases we maybe (I don't know) shouldn't worry about.
Well, I think we should try to cover as many as possible. A new feature leads to new debt. 

>So I see two options to address 2:
>
>   1. remove it from this PR and submit a separate one just on that
>   2. forget about your latest concern ;-)

I was thinking about that as well. Best is maybe to leave this PR as is, and add another one so we can compare and discuss potential advantages and drawbacks of each.

bdarcus:
> Well, I think we should try to cover as many as possible. A new feature leads to new debt.

Yes, but as you think about this more:

We should not try to address every potential case with a solution that makes the feature more difficult for all cases.

If we can find an elegant way to get the benefits without the costs, that's of course ideal.

> Best is maybe to leave this PR as is, and add another one so we can compare and discuss potential advantages and drawbacks of each.

What you could do is:

1. revert this to draft
2. open a new PR for 2 (once you're reasonably confident), and when that is settled, revise this to remove 2, so that in the end we have two PRs, and this one is just for 1?

denismaier:
> We should not try to address every potential case with a solution that makes the feature more difficult for all cases.
> 
> If we can find an elegant way to get the benefits without the costs, that's of course ideal.

Absolutely! This feature should work in 90%-95% of all cases without additional complexity for users, and not too much for style authors. Concerning the three options (1. Markup, 2. adding delimiting punctuation to `title-main`, 3. using `title-parts` for input), I tend to think that 2 and 3 will both give us the benefit without too much costs. Not sure which one is preferable though. 2 would be completely within the scope of this PR, it's just another option. So that would be simple. 3 may be more complicated in some ways (for users), easier in others (for citeprocs).

denismaier:
> So I see two options to address 2:
> 
> 1. remove it from this PR and submit a separate one just on that
> 2. forget about your latest concern ;-)

Ok, I went over this again, and I think the problem is indeed of more theoretical nature. There might be some super rare edge cases with 0.01% probability where this might create problem, but I'll ignore this for now.

Concerning APA style, I went through several lib-guides as well as through the examples on the official APA webpage, and I consistently saw a colon between title and subtitle. So, I'd assume there is some sort of implicit normalization going on here? @bwiernik Is that assumption correct?

Anyway, I think this PR is ok as is.

schemas/styles/csl.rnc
denismaier: @bwiernik 
You were more in favor of pre-defined sets if I remember correctly. Perhaps can answer that question? I think that a space-separated list may be more flexible, but that we also cover the main combinations with the predefined options, which may be more user-friendly.

bdarcus: And to be clear, I'm not sure what's better myself: just want to consider it.

bwiernik:
No, APA does not normalize, ever. It’s just that colons are the most common delimiter.

denismaier:
> No, APA does not normalize, ever. It’s just that colons are the most common delimiter.

Ok.  So, what do you think about the problem described above? Do you think it exists? And do we need to address this somehow?

bwiernik:
I don’t think this problem exists. We would almost always expect the full `title` to be provided.

bdarcus:
Just a head's  up that the merges I did today caused some conflict between master and v1.1 that I'm having trouble resolving.

bwiernik:
@bdarcus I can reconcile in a PR

bdarcus:
> @bdarcus I can reconcile in a PR

Thank you!

bwiernik:
https://github.com/citation-style-language/schema/pull/245

denismaier:
Flagging this for @jgm

Issue #138, opened 2017-07-03
Create terms for "no place" and "no name"


rmzelle:
(a.k.a. "sine loco"/"sine nomine")

See https://github.com/citation-style-language/styles/pull/2722/files/0d1487e66c6c73bce8fa6358cb3094701e9023ec#r122593750 and https://github.com/citation-style-language/schema/issues/90#issuecomment-4329346.

denismaier:
@bwiernik What do you think about this? Is that still relevant?

bwiernik:
Yep

bdarcus:
To  clarify, these don't literally mean "no", but rather "unknown"?

PR #206, opened 2020-05-27
Add terms for "no-place" and "no-name"

denismaier:
Closes https://github.com/citation-style-language/schema/issues/138


bdarcus:
Just bumping this @denismaier; see question above.

bwiernik:
```
no-place: indicates place of publication is unlisted or unknown
no-name: indicates publisher is unlisted or unknown
```

cf.
https://en.wikipedia.org/wiki/Sine_nomine
https://en.wikipedia.org/wiki/List_of_Latin_phrases_(S)#sine_loco

bdarcus:
Perfect; thanks!

This raises another minor question; should we do "no name" or "no-name"?

I know we have a lot of existing terms that are not tokens, but that "sub verbo" issue makes me leery about that.

bwiernik:
Let's adopt an all-token approach "just in case". I've been thinking about a proposal lately that might be much easier if terms were reliably tokens.

denismaier:
Will do tomorrow. 
What about "no date"? Change that to "no-date" as well? I've actually had the all-token version at first, but changed it to the version without hyphens because of "no date"

bdarcus:
> Will do tomorrow.
> What about "no date"? Change that to "no-date" as well? I've actually had the all-token version at first, but changed it to the version without hyphens because of "no date"

I made the changes on the first two.

Is "no date" new, so no legacy?

bdarcus:
Actually, no it's not new. 

So let's hold off on that now. Perhaps we can contemplate a retroactive fix as a separate step.

denismaier:
No. It's not new.
I'd say we clean this up later together with the other non-token terms, but having "no date" next to "no-place" is particularly ugly.

bdarcus:
Yeah, I know.

It's a big schema, that's now almost 15 years old, so there's going to
be some ugly corners.

Hopefully we can clean this up sooner rather than later.

PR #210, opened 2020-05-29
Simplify and add structure to csl.rnc formatting

### LINKED ISSUE #209: simplify and restructure rnc formatting
This started as me noticing a problem: every time one of us touches an rnc file, it seems, our editors introduce slightly different modification of whitespace and such, so that commits end up getting confusing (because the diff includes a ton of trivial whitespace changes).

But in that process, I notice there's room to add further structure and simplify formatting in places to make easier to handle.

PR coming.

bdarcus:
This does a few things:

1. It introduces section break comments to make it easier to see the structure of the primary csl.rnc schema. 
2. Related, those section breaks also suggest places where we might seriously consider breaking out groups of patterns into separate files (though perhaps a a separate step). I've flagged those places below.
3. It removes more verbose sectional annotations in favor of the double-hash patterns, which trang will convert to the same rng output.

closes #209


bdarcus:
As I say in #209, this started to see if possible for us to run rnc changes through trang to get a standardized rnc format.

Here's an fragment of that output. The main ugliness is what happens to the embedded schematron.

<details>
``` rnc
namespace a = "http://relaxng.org/ns/compatibility/annotations/1.0"
namespace bibo = "http://purl.org/ontology/bibo/"
namespace cs = "http://purl.org/net/xbiblio/csl"
namespace dc = "http://purl.org/dc/elements/1.1/"
namespace sch = "http://purl.oclc.org/dsdl/schematron"
namespace xhtml = "http://www.w3.org/1999/xhtml"

# CSL schema metadata

dc:title [ "Citation Style Language" ]
dc:creator [ "Bruce D'Arcus" ]
dc:creator [ "Simon Kornblith" ]
bibo:editor [ "Frank Bennett" ]
bibo:editor [ "Rintze Zelle" ]
dc:rights [
  "Copyright 2007-2018 Citation Style Language and contributors"
]
dc:license [ "MIT license" ]
dc:description [
  "RELAX NG compact schema for the Citation Style Language (CSL)."
]
# Embedded Schematron rules to detect calls to nonexistent macros

sch:ns [ uri = "http://purl.org/net/xbiblio/csl" prefix = "cs" ]
sch:pattern [
  "\x{a}" ~
  "    "
  sch:rule [
    context = "//cs:text[@macro]"
    "\x{a}" ~
    "      "
    sch:assert [
      test = "@macro = /cs:style/cs:macro/@name"
      "This macro call has no corresponding macro."
    ]
    "\x{a}" ~
    "    "
  ]
  "\x{a}" ~
  "    "
  sch:rule [
    context = "//cs:key[@macro]"
    "\x{a}" ~
    "      "
    sch:assert [
      test = "@macro = /cs:style/cs:macro/@name"
      "This macro call has no corresponding macro."
    ]
    "\x{a}" ~
    "    "
  ]
  "\x{a}" ~
  "    "
  sch:rule [
    context = "/cs:style/cs:macro"
    "\x{a}" ~
    "      "
    sch:assert [
      test = "count(/cs:style/cs:macro/@name[. = current()/@name]) = 1"
      "This macro does not have a unique name."
    ]
    "\x{a}" ~
    "    "
  ]
  "\x{a}" ~
  "  "
]

## Subparts of the CSL schema
include "csl-terms.rnc"
include "csl-types.rnc"
include "csl-variables.rnc"
include "csl-categories.rnc"

## cs:style and cs:locale - Root Elements
div {
  start =
    independent-style.style | dependent-style.style | locale-file.locale
  independent-style.style =
    element cs:style {
      
      ## Select whether citations appear in-text or as notes.
      attribute class { "in-text" | "note" },
      style.default-locale,
      style.options,
      version,
      independent-style.style.info,
      (style.locale*
       & style.macro*
       & style.citation
       & style.bibliography?)
    }
  dependent-style.style =
    element cs:style {
      style.default-locale,
      version,
      dependent-style.style.info,
      dependent-style.style.legacy-attributes
    }
  locale-file.locale =
    element cs:locale {
      
      ## Specify the locale of the locale file.
      attribute xml:lang { xsd:language },
      version,
      locale-file.locale.info?,
      (locale.style-options & locale.date+ & locale.terms)
    }
  style.default-locale =
    
    ## Set a default style locale.
    attribute default-locale { xsd:language }?
  version =
    
    ## Indicate CSL version compatibility ("1.0" for CSL 1.0).
    [ a:defaultValue = "1.0" ] attribute version { "1.0" }
  
  ## Obsolete for dependent styles. Will be disallowed with CSL 1.1.
  dependent-style.style.legacy-attributes =
    attribute class { "in-text" | "note" }?,
    style.options
}
``` 
<details>

csl.rnc
```diff
@@ -74,47 +73,46 @@ div {
...
 }
-[
-  a:documentation [
-    xhtml:h2 [ "cs:info - Style and Locale File Metadata" ]
-  ]
-]
+
+#==============================================================================
+
+## cs:info - Style and Locale File Metadata
```
bdarcus, line 113: This could be a separate `csl-metadata.rnc` file.

csl.rnc
```diff
@@ -152,171 +150,173 @@ div {
...
+
   info-text =
     attribute xml:lang { xsd:language }?,
     text
 }
-[ a:documentation [ xhtml:h2 [ "cs:locale in Independent Styles" ] ] ]
+
+#==============================================================================
+
+## cs:locale in Independent Styles
```
bdarcus, line 284: This could be a separate `csl-locale.rnc` file.

csl.rnc
```diff
@@ -400,208 +403,217 @@ div {
...
-  
+
   bibliography.layout =
     element cs:layout { affixes, font-formatting, rendering-element+ }
 }
-[ a:documentation [ xhtml:h2 [ "cs:names Rendering Element" ] ] ]
+
+#==============================================================================
+
+## cs:names Rendering Element
```
bdarcus, line 434: This could be a separate `csl-names.rnc` file.

csl.rnc
```diff
@@ -400,208 +403,217 @@ div {
...
-  
+
   ## Short version of cs:names, without children, allowed in cs:substitute.
   substitute.names = element cs:names { names.attributes }
 }
-[ a:documentation [ xhtml:h2 [ "cs:date Rendering Element" ] ] ]
+
+#==============================================================================
+
+## cs:date Rendering Element
```
bdarcus, line 616: This could be a separate `csl-date.rnc` file.

csl.rnc
```diff
@@ -697,59 +712,65 @@ div {
...
     | (
        ## Select a variable.
        attribute variable { variables.standard },
        [ a:defaultValue = "long" ] attribute form { "short" | "long" }?)
 }
-[ a:documentation [ xhtml:h2 [ "cs:number Rendering Element" ] ] ]
+
+#==============================================================================
+
+## cs:number Rendering Element
```
bdarcus, line 738: This could be a separate `csl-number.rnc` file.

csl.rnc
```diff
@@ -697,59 +712,65 @@ div {
...
     ## Number forms: "numeric" ("4"), "ordinal" ("4th"), "long-ordinal"
     ## ("fourth"), "roman" ("iv").
     [ a:defaultValue = "numeric" ]
     attribute form { "numeric" | "ordinal" | "long-ordinal" | "roman" }?
 }
-[ a:documentation [ xhtml:h2 [ "cs:label Rendering Element" ] ] ]
+
+#==============================================================================
+
+## cs:label Rendering Element
```
bdarcus, line 758: This could be a separate `csl-label.rnc` file.

csl.rnc
```diff
@@ -758,10 +779,13 @@ div {
     strip-periods,
     text-case
 }
-[ a:documentation [ xhtml:h2 [ "cs:group Rendering Element" ] ] ]
+
+#==============================================================================
+
+## cs:group Rendering Element
```
bdarcus, line 785: This could be a separate `csl-group.rnc` file.

csl.rnc
```diff
@@ -775,289 +799,296 @@ div {
...
     }
-  
+
   group.attributes = notAllowed?
 }
-[ a:documentation [ xhtml:h2 [ "Options" ] ] ]
+
+#==============================================================================
+
+## Style Options
```
bdarcus, line 808: This could be a separate `csl-style-options.rnc` file.

bdarcus:
Beyond the PR as it exists now, I've flag with line comments where we can, and probably mostly should, split those patterns into separate files.

Let me know what you think reviewers.

bdarcus:
> Here's a fragment of that output. The main ugliness is what happens to the embedded schematron.

And [here's the full output of using the trang conversion](https://gist.github.com/bdarcus/70ba078fe918bd7891402ba79dd321e2), removing the embedded schematron.

bdarcus:
One other thing (sorry the deluge): I'm wondering if we should remove the "a:defaultValue" declarations?

Their only functional purposes is for conversion to DTD, which is a non-issue.

I think I only put them in original for doc purposes. But it doesn't seem validating editors pick them up, and it would simplify things to remove.

bdarcus:
Also wondering about documenting strings, which alas is not ideal. Basically only way to do it is to define separate patterns for each string. On the below example, that's represented by `x-string`.

```
start = test

test = element test { x-string | y-strings }

## the x string
x-string = "x string"

y-strings =
## y-one
"y-one" |
## y-two
"y-two"
```

Converted to XML:

``` xml
?xml version="1.0" encoding="UTF-8"?>
<grammar xmlns:aa="http://relaxng.org/ns/compatibility/annotations/1.0" xmlns:a="http://www.example.com/annotation" xmlns="http://relaxng.org/ns/structure/1.0">
  <start>
    <ref name="test"/>
  </start>
  <define name="test">
    <element name="test">
      <choice>
	<ref name="x-string"/>
        <ref name="y-strings"/>
      </choice>
    </element>
  </define>
  <define name="x-string">
    <aa:documentation>the x string</aa:documentation>
    <value>x string</value>
  </define>
  <define name="y-strings">
    <choice>
      <value>y-one</value>
      <aa:documentation>y-one</aa:documentation>
      <value>y-two</value>
      <aa:documentation>y-two</aa:documentation>
    </choice>
  </define>
</grammar>
```

bwiernik:
> Also wondering about documenting strings, which alas is not ideal with RNC. Basically only way to do it is to define separate patterns for each string. On the below example, that's represented by x-string.

Not following where this is for? Can you give a real example?

bdarcus:
> > Also wondering about documenting strings, which alas is not ideal with RNC. Basically only way to do it is to define separate patterns for each string. On the below example, that's represented by x-string.
> 
> Not following where this is for? Can you give a real example?

The issue that raised the question for me is all the discussion of type, variable, etc. strings (eg [here](https://github.com/citation-style-language/schema/blob/master/csl-types.rnc#L4)), which are completely undocumented in the schema, and so unavailable to validating editors.

Note: I've not yet confirmed validating editing will read these documentation strings for attribute values, but am assuming they do.

tools/rnc-prettyprint.sh
```diff
@@ -0,0 +1,7 @@
+# This assumes you have trang installed via package manager
+# which installs a script to provide a "trang" command.
+# Make sure you run this before commiting any changes to the 
+# rnc files.
+
+trang csl.rnc csl.rng
```
bdarcus, line 6: I haven't yet determined how well this would work, but this is the idea for using trang to provide a consistent format for the rnc files.

bdarcus:
DTD is really old tech that we don't support.

rmzelle:
(for anybody else reviewing this at GitHub, you can append `?w=1` to hide white-space changes: https://github.com/citation-style-language/schema/pull/210/files?w=1 )

csl.rnc
rmzelle: Note that we have `locale-file.locale` (root element of CSL locale files) and `style.locale` (locale section of CSL styles) definitions, so this doesn't seem like the right place for a split.

tools/rnc-prettyprint.sh
```diff
@@ -0,0 +1,7 @@
+# This assumes you have trang installed via package manager
+# which installs a script to provide a "trang" command.
+# Make sure you run this before commiting any changes to the 
+# rnc files.
```
rmzelle, line 4:
```
# Make sure you run this before commiting any changes to the 
# rnc files.
```

It would be nice if CI could take of this. Same for https://github.com/citation-style-language/styles/, where I currently run https://formatter.citationstyles.org/ manually on most pull requests, with periodic runs of https://github.com/citation-style-language/utilities/blob/master/csl-reindenting-and-info-reordering.py to catch any pretty printing we missed during merging.

It looks like Github Actions can be used for this, but only for PRs originating from the same repo (see https://github.com/stefanzweifel/git-auto-commit-action#unable-to-commit-into-prs-from-forks).

rmzelle:
> I'm wondering if we should remove the "a:defaultValue" declarations?
> 
> Their only functional purpose is for conversion to DTD, which is a non-issue.
> 
> I think I only put them in originally for doc purposes.

Looks like I added them (https://github.com/citation-style-language/schema/commit/5dc8bef13f6a2a64e237b7bfea007e9d85f86cc6). Should we move these defaults back into the comments accompanying the attributes, or should we just rely on the spec for that info?

So none of the modern tools pick up annotations like the following (see https://relaxng.org/compact-tutorial-20030326.html#id2816743, and compare sections 13.1 and 13.3):

```
element addressBook {
  [ a:documentation [ xml:lang="en" "Information about a single address." ] ]
  element card {
    element name { text },
    element email { text }
  }*
}
```

or

```
## Represents an
## address book.
element addressBook {
  element card {
    element name { text },
    element email { text }
  }*
}
```

bwiernik:
I think listing the default value in the schema is useful. Perhaps putting them in comments instead of annotations would be easier for folks unfamiliar or out of practice with rnc?

bdarcus:
> Looks like I added them ([5dc8bef](https://github.com/citation-style-language/schema/commit/5dc8bef13f6a2a64e237b7bfea007e9d85f86cc6)). Should we move these defaults back into the comments accompanying the attributes, or should we just rely on the spec for that info?

I think it depends on whether it would be helpful for someone editing a CSL style in a validating editor to know the default values.

As in, the question is purely practical.

Developers that are coding can refer to the spec.

If we do move to comments, I'm not sure how though.

At first I was thinking we could borrow the convention in the spec, which adds "(default)" after the value. But I don't think that works here. This?

```
        ## Limit the date parts rendered (default is "year-month-day" )
        attribute date-parts {
          
          ## Year, month and day
          "year-month-day"
          | 
            ## Year and month
            "year-month"
          | 
            ## Year only
            "year"
        }?,
```

tools/rnc-prettyprint.sh
bdarcus: Isn't the problem with this that, by definition, CI actions take place after commit?

That's why I was wondering about a pre-commit hook.

bdarcus:
BTW, my latest commit uses the trang-based reformatting. With the defaultValue pieces, that part doesn't get screwed up.

So we're not sure, we could always defer that question.

tools/rnc-prettyprint.sh
rmzelle: Yes. (for the styles repo, pre-commit hooks don't make much sense, since we have so many infrequent/one-off contributors, and a lot of contributions that come in via the GitHub website instead of a local client, but I guess you could use them here)

PR #212, opened 2020-05-29
Restructure repo

### LINKED ISSUE #216: create stable URIs for json schemas at citationstyle.org
> Perhaps as a separate change in the future, I wonder if it makes sense to define a URL like `csl-data-schema.citationstyles.org` that can be changed for `$id` and can be updated if the source code path or repo changes.

I think @dstillman was recommending something like this, though I think he was recommending hosting a copy of the schema files on that site.

Is it feasible for us to do both changes together? That would ensure only one breaking change, rather than two.

If we do, do we need to version the json schema URI ids?

I don't know know who maintains that site, though, or how!

_Originally posted by @bdarcus and responding to @dhimmel in https://github.com/citation-style-language/schema/pull/212#issuecomment-636482128_

bdarcus:
Restructure repo, update json schema ids
                                                              
This restructures the repo to make it easier to manage, and also to  
make room for possibly merging back content now stored in separate 
repos.                               
                                                                
Because the paths have changed on most files, we have also changed the 
ids on the json schemas to the following stable URLs:
                                                                         
https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json
https://resource.citationstyles.org/schema/latest/input/json/csl-data.json

closes #216 

New structure looks like this:

``` console
├── LICENSE.txt
├── README.md
├── schemas
│   ├── input
│   │   ├── csl-citation.json
│   │   └── csl-data.json
│   └── styles
│       ├── csl-categories.rnc
│       ├── csl-data.rnc
│       ├── csl-relaxed.rnc
│       ├── csl-repository.rnc
│       ├── csl.rnc
│       ├── csl.sch
│       ├── csl-terms.rnc
│       ├── csl-types.rnc
│       └── csl-variables.rnc
├── tests
│   └── schemas
│       ├── input
│       │   ├── requirements.txt
│       │   └── test_json.py
│       └── styles
└── tools
```

If we agree to merge the tests here, they could go in a `tests/processor` subdirectory.

If we agree to merge one-or-more of the spec docs here, they would go in a `spec` directory.


denismaier:
Looks good. (That the tests fail is expected with such a fundamental change, right?)

bdarcus:
> Looks good. (That the tests fail is expected with such a fundamental change, right?)

Yes. The test rely on paths.

bdarcus:
> * For the folder structure, I'm wondering if we could improve on "data" and "styles".
>   
>   * For "data", "CSL-JSON" seems has become the accepted name to refer to the CSL input metadata format (https://github.com/citation-style-language/schema#csl-json-schema), although I guess we might eventually provide metadata schemas for non-JSON formats as well. How about "input" or "metadata"?

The directory also includes the csl-citation.json. a) does that still belong here, and if yes, b) how does that impact the decision here?

>   * For "styles", the schemas (currently) also cover the locale files. Not sure I have a better name, though.

Anyone else have input on the above?

> * If we're considering moving the specification (and other documentation) to this repo, we should make sure the new directory structure will play nice with Read the Docs (https://readthedocs.org/) so https://docs.citationstyles.org isn't affected.

We can do in stages; first restructure, then see about spec/docs.

rmzelle:
> The directory also includes the csl-citation.json. a) does that still belong here, and if yes, b) how does that impact the decision here?

Since csl-citation.json is an extension of csl-data.json, yes, they should be kept together.

Note also that Zotero and Mendeley both use `"schema": "https://github.com/citation-style-language/schema/raw/master/csl-citation.json"` in their word-processor-embedded metadata (see https://github.com/rmzelle/ref-extractor/wiki for examples), so that link will break if we move the file. Not sure that's an issue, though. (pinging @dstillman)

> @rmzelle - one other questions: do we still use the `csl-repository.rnc` file?

Nowadays we mostly just rely on our Travis CI tests, but there are a few checks included in this adapted schema that aren't covered by that. There is a script to validate against `csl-repository.rnc` at https://github.com/citation-style-language/utilities/blob/8b4ca0c4545c640305b3a3b220b2a9ca9c5dae0f/style-qc.sh, but I haven't run it in a long time. `csl-repository.rnc` is pretty well documented internally in how it differs from the regular schema. We should probably either run CI on a stricter schema for the repository, or implement these checks separate in CI.

bdarcus:
> Note also that Zotero and Mendeley both use `"schema": "https://github.com/citation-style-language/schema/raw/master/csl-citation.json"` in their word-processor-embedded metadata (see https://github.com/rmzelle/ref-extractor/wiki for examples), so that link will break if we move the file. Not sure that's an issue, though. (pinging @dstillman)

OK, thanks.

My hope would be to just do a clean break, but worst case, maybe we could add a symlink?

bdarcus:
Here's a project that uses readthedocs with a structure similar to what I was thinking.

https://github.com/ericholscher/django-kong

So definitely seems possible; am just not exactly sure how to adapt it ATM.

bwiernik:
> * I guess we might eventually provide metadata schemas for non-JSON formats as well. How about "input" or "metadata"?

pandoc, for example, has translated the CSL JSON format into CSL YAML.

dstillman:
> Note also that Zotero and Mendeley both use `"schema": "https://github.com/citation-style-language/schema/raw/master/csl-citation.json"` in their word-processor-embedded metadata (see https://github.com/rmzelle/ref-extractor/wiki for examples), so that link will break if we move the file. (pinging @dstillman)

Pinging @adomasven to be safe, but I'm pretty sure we don't use that for anything.

> My hope would be to just do a clean break, but worst case, maybe we could add a symlink?

Symlinks don't create redirects for GitHub raw links, if that's what you mean.

I'm not sure what the point is of our including the schema URL in the embedded metadata, but nothing that needs to be programmatically accessible should use a GitHub raw link, since they can never be redirected (other than when a repo is renamed). If you'd like the schema file (or anything else) to be usable in that way, it should be hosted on citationstyles.org where you can guarantee that it will always continue to resolve. (Same goes for styles, which is why we don't use GH directly for distributing those.)

bdarcus:
> Symlinks don't create redirects for GitHub raw links, if that's what you mean.

No, I was just assuming in the case of scripts that might rely on local file paths.

I'll plan to merge this, without symlinks, but wait a bit for feedback from @adomasven.

tests/schemas/input/test_json.py
```diff
@@ -11,7 +11,7 @@
 """
 
 root = pathlib.Path(__file__).parent.parent
-json_paths = list(root.glob('*.json'))
+json_paths = list(root.glob('schemas/input/*.json'))
```
bdarcus, line 14: @dhimmel - I'm doing something wrong here. Trying to adjust the paths for the updated repo structure in this PR.

I'm assuming I actually need to remove this commit, and instead adjust the root, but I'm not sure how.

Can you help please?

dhimmel: Looking through https://github.com/citation-style-language/schema/pull/212/checks?check_run_id=724834161#step:5:53

```
E       FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/schema/schema/tests/schemas/csl-citation.json'
```

I think you probably need to add the new paths in :

https://github.com/citation-style-language/schema/blob/2bbf44f8eb51f9ed5fa2cc84f319d8e4833a4174/tests/test_json.py#L25-L28

dhimmel: `json_paths` is used by

https://github.com/citation-style-language/schema/blob/2bbf44f8eb51f9ed5fa2cc84f319d8e4833a4174/tests/test_json.py#L52-L57

Just tests whether json files in repo are parseable as JSON. So I might change this to:

```suggestion
json_paths = list(root.glob('**/*.json'))
```

Will find any `.json` file in the repo.

dhimmel:
This pull request is likely to break things for many users who are linking to https://github.com/citation-style-language/schema/raw/master/csl-citation.json.

Given that the schema includes

https://github.com/citation-style-language/schema/blob/2bbf44f8eb51f9ed5fa2cc84f319d8e4833a4174/csl-citation.json#L4

And the README never recommended not linking to the schema this way. You can look search [the URL on GitHub](https://github.com/search?q=https%3A%2F%2Fgithub.com%2Fcitation-style-language%2Fschema%2Fraw%2Fmaster%2Fcsl-citation.json&type=Code) to potentially find affected users.

With Manubot, I created a fork and [linked to a specific branch](https://github.com/manubot/manubot/blob/217e51473f1fd1c6427803676b3c70d44314bb93/manubot/cite/citeproc.py#L17-L23) on my fork, as an extra precaution.

That being said, I am still in favor of this change... but given how disruptive it will be, it'd be best to only have to make it once.

Also you probably want to update the schema `$id` so the URL is resolvable.

Perhaps as a separate change in the future, I wonder if it makes sense to define a URL like `csl-data-schema.citationstyles.org` that can be changed for `$id` and can be updated if the source code path or repo changes.

tests/schemas/input/test_json.py
bdarcus: So is the test failing now because the test code is actually in master?  E.g. this code is right now, but it just cannot avoid conflicting with master?

bdarcus:
> Perhaps as a separate change in the future, I wonder if it makes sense to define a URL like `csl-data-schema.citationstyles.org` that can be changed for `$id` and can be updated if the source code path or repo changes.

I think @dstillman was recommending something like this, though I think he was recommending hosting a copy of the schema files on that site.

Is it feasible for us to do both changes together?

I don't know know who maintains that site, or how!

tests/schemas/input/test_json.py
dhimmel: > So is the test failing now because the test code is actually in master?

No I think it's failing because of the issue in https://github.com/citation-style-language/schema/pull/212#discussion_r432952845

bdarcus:
While we're thinking about longer-term stability for the json schema, do we need to anticipate different versions, so different URI ids?

bdarcus:
> While we're thinking about longer-term stability for the json schema, do we need to anticipate different versions, so different URI ids?

I decided to pull this out into a separate issue.

tests/schemas/input/test_json.py
```diff
@@ -11,7 +11,7 @@
 """
 
 root = pathlib.Path(__file__).parent.parent
```
dhimmel, line 13: still failing at https://github.com/citation-style-language/schema/runs/725123572#step:5:53.

I think the issue is that `root` above is no longer correct. We need to add more parent directories like

```suggestion
root = pathlib.Path(__file__).parent.parent.parent.parent
```

bdarcus:
That fixes part of it. Ii'm guessing the remaining error is the uri issue
in the schema itself?

>

dhimmel:
> That fixes part of it. Ii'm guessing the remaining error is the uri issue in the schema itself?

No, based on the error messages, I think there's still an outdated path at:

https://github.com/citation-style-language/schema/blob/2657df6105f22c692f727693db794bc6ae945689/tests/schemas/input/test_json.py#L41

We don't test for `$id` actually resolving AFAIK, but you should probably update it before merging... even if you pick a more stable URL later.

bdarcus:
> > That fixes part of it. Ii'm guessing the remaining error is the uri issue in the schema itself?
> 
> No, based on the error messages, I think there's still an outdated path at:

Doh! Fixed now. Thanks.

> https://github.com/citation-style-language/schema/blob/2657df6105f22c692f727693db794bc6ae945689/tests/schemas/input/test_json.py#L41
> 
> We don't test for `$id` actually resolving AFAIK, but you should probably update it before merging... even if you pick a more stable URL later.

I'll wait to get some feedback on #216.

PR #213, opened 2020-05-30
Ignore rng

denismaier:
I suggest we add `*.rng` to `.gitignore` so we don't have to worry about rng-files after `trang csl.rnc csl.rng`.


bdarcus:
Yes, I was thing about that yesterday.

But the script should also delete them when done.

denismaier:
> But the script should also delete them when done.

Really, should it?
I see only two commands in the script:

```
trang csl.rnc csl.rng
trang csl.rng csl.rnc
```

rng-files remain there.

denismaier:
I don't know, but perhaps there is a command line parameter.
Or just `rm *.rng`?

bdarcus:
Sorry; my wording was vague.

I meant to say I need to revise it to do that.

denismaier:
Ok then.

bdarcus:
Thanks. I just added it directly, since I got a little confused about the three commits on this PR.

rmzelle:
(done in https://github.com/citation-style-language/schema/commit/0a1f43a38a60e4266c1eccafcc7e1bc7622a5715)

PR #224, opened 2020-06-06
Add issue templates

bdarcus:
Not really sure exactly what these should be in the end, but this is a start.

It's designed to address the biggest thing historically missing from issue requests (missing context, including style guide links). 

Note, though, I only tweaked the default template, so not sure.

Please review this "feature request" template, and also weigh in on if you think we need others (like "bug report"), and if so, what the content should be.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,25 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: enhancement
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. 
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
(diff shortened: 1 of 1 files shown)


bdarcus:
I have two approval to merge as is, without adding a bug report template.

Any other thoughts?

denismaier:
I am not sure we need a bug template. I first thought we could something like this:

---
**Describe the bug**
A clear and concise description of what the bug is.

**To Reproduce**
Steps to reproduce the behavior:
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error

**Expected behavior**
A clear and concise description of what you expected to happen.

**Screenshots**
If applicable, add screenshots to help explain your problem.

**Tested with:**
 - OS: [e.g. Windows 10]
 - Reference Manager [e.g. Zotero, Mendeley]
 - Word processor

**Additional context**
Add any other context about the problem here.
---

But probably, this will give us mostly reports about bugs in certain implementations.

bdarcus:
IF we added a bug template, we would want to make clear that it's only for mistakes in the schemas.

For example, #225.

bwiernik:
I think asking casual contributors to discern between "issues" and "bugs" is likely to be more of a headache than it's worth.

PR #225, opened 2020-06-06
Update csl-citation.json

lukesmurray:
The way the reference fo `itemData` is currently written `csl-data.json` is assumed to be a folder. This breaks dereferencing in json schema tools. Fix is simple, remove the extra `/` and access items relative to `csl-data.json`

### DIFF
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -28,7 +28,7 @@
                         ]
                     },
                     "itemData": {
-                        "$ref": "csl-data.json/#/items" 
+                        "$ref": "csl-data.json#/items" 
                     },
                     "prefix": {
                         "type": "string" 


lukesmurray:
@bdarcus you edited the file maybe you can take a look at this.

bdarcus:
JSON not my thing, but maybe @dhimmel could take a few minutes to weigh in?

@lukesmurray do you know of a good commandline json schema validator? I can't immediately find one that supports draft07.

bdarcus:
Actually, probably doesn't matter; we have the CI!

lukesmurray:
@bdarcus https://www.jsonschemavalidator.net/ is what I used for validation. Not sure about CLI validators.

This tool was unable to resolve the refs. https://github.com/APIDevTools/json-schema-ref-parser

bdarcus:
OK, merged. Thanks.

I still want to know why the CI didn't originally catch this, and if there's anything we do to fix that going forward.

dhimmel:
> I still want to know why the CI didn't originally catch this, and if there's anything we do to fix that going forward.

The tests of the `csl-citation.json` schema are very minimal. Probably did not touch that part of the schema. If someone compiles instances of valid and invalid JSON, they can be added to the tests.

bdarcus:
But doesn't the code here, at least in theory, test the validity of the
schemas?

https://github.com/citation-style-language/schema/blob/9f427f2c73499a53754252aa733749b54085f33f/tests/schemas/input/test_json.py#L69

As in, could be a bug in the python jsonschema package?

bdarcus:
In any case, yeah, we just need to add a test that uses csl-citation.json.

I'll look into that.

PR #231, opened 2020-06-09
Add international identifiers

denismaier:
This adds new international identifiers, as per 
citation-style-language/csl-evolution#39

ISWC (for a musical work itself)
ISMN (for a musical notation/score)
ISRC (for an audio or visual recording)
ISAN (for an audiovisual work itself)
ISCI (an identifier for an archival collection)

These are ISO identifiers analogous to ISSN and ISBN, and appear to be 
important for citation and cataloging in fields like musicology 
(e.g., citation-style-language/zotero-bits#28).

Fixes: citation-style-language/csl-evolution#39


bdarcus:
I'll wait a bit if any other feedback, but in the meantime revised the summary to format it for a squash commit message.

PR #234, opened 2020-06-09
Add translated-title

bwiernik:
# Description

`translated-title` is a translation of a different-lanugage title (e.g., to provide an English translation of a Spanish title in an English-locale bibliography). For example, this is the APA format for an item with a non-English title:
Chaves-Morillo, V., Gómez Calero, C., Fernández-Muñoz, J. J., Toledano-Muñoz, A., Fernández-Huete, J., Martínez-Monge, N., Palacios-Ceña, D., & Peñacoba-Puente, C. (2017). La anosmia neurosensorial: Relación entre subtipo, tiempo de reconocimiento y edad [Sensorineural anosmia: Relationship between subtype, recognition time, and age]. _Clínica y Salud, 28_(3), 155–161. https://doi.org/10.1016/j.clysa.2017.04.002

This is not intended to provide full multilingual support ala CSLm.

Addresses the `translated-title` discussion here: https://github.com/citation-style-language/zotero-bits/issues/50

## Type of change

- [x] Variable Addition (adds a new variable or type string)

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -379,6 +379,12 @@
       "title-short": {
         "type": "string"
       },
+      "translated-title": {
+        "type": "string"
+      },
+      "translated-title-short": {
+        "type": "string"
+      },
       "URL": {
         "type": "string"
       },
--- a/schemas/styles/csl-variables.rnc
+++ b/schemas/styles/csl-variables.rnc
@@ -106,6 +106,8 @@ div {
     | "status"
     | "title"
     | "title-short"
+    | "translated-title"
+    | "translated-title-short"
     | "URL"
     | "version"
     | "volume-title"


denismaier:
While the simplicity of just adding `translated-title` certainly has its merits, I'd like to ask whether we could add a more flexibel solution, perhaps not full CSLm, but something closer. I don't really like the idea of having to change item metadata when you cite your example item in a German publication.

Some time ago @fbennett has proposed a simplified method for Jurism metadata sync: https://juris-m.readthedocs.io/en/latest/dev-sync-simplification.html
Here, language codes are appended to ordinary variables with two hyphens, so we could do `title--de` or `title--en` in `extra`.

What do you think?

Obviously, the question would be, how you'd use this with e.g. Zotero, Pandoc, etc. But I'm sure we could work something out...

bwiernik:
What about this?
  - The CSL variable is `translated-title`. 
  - Entry of that variable could be locale-specific:
     - `translated-title--en`
     - `translated-title--de`
  - If there is a variant that matches the document's locale, that variant is used, otherwise the generic  `translated-title` is used.
  - If there is no matching locale or generic `translated-title`, the variable is treated as empty.

bdarcus:
> Here, language codes are appended to ordinary variables with two hyphens, so we could do `title--de` or `title--en` in `extra`.

Reminder: these discussions need to be agnostic on client. So if we're talking about data, we're talking csl json.

And if we're talking that, what's the proposal that works there, including in json schema?

denismaier:
> Reminder: these discussions need to be agnostic on client. So if we're talking about data, we're talking csl json.

Sure, but GUI constraints are serious constraints... So I just wanted to emphasize that this could be used even without changes here.

denismaier:
> So for @bwiernik's example, "translated-title--en" would be a json key? Can you express that in json schema?

Perhaps we can add an addition for processors: Make sure that you pick up language codes appended to ordinary variables...???

bdarcus:
The schema defines what's valid, and what's not, though.

This suggests it might be possible:

https://stackoverflow.com/questions/16222633/how-would-you-design-json-schema-for-an-arbitrary-key

... but this is the kind of thing it seems we need input from implementers are, including developers working with the json.

bdarcus:
I will add, though, if we go down this path with titles, where do we stop?

bwiernik:
Personally, I don't know that the locale-specification is fully necessary, and I think that it might best be left to a potential future multilingual module with expanded multilingual features. A single normal field would meet the most common use case of researchers who primarily work in one language. Sticking with a flat field would also let this variable be merged into v1.0.2.

> I will add, though, if we go down this path with titles, where do we stop?

To a degree, I think translated-title` is the major field that would ever need to be translated outside of full multilingual bibliographies, so we could stop here. 

As an aside, that JSON structure you gave could be used for addressing things like adding `original-` and `reviewed-` versions of all variables without needing to enumerate them.

bdarcus:
> A single normal field would meet the most common use case of researchers who primarily work in one language. Sticking with a flat field would also let this variable be merged into v1.0.2.

That's what I was going to ask; is there value in us just accepting this now, and deferring more complex use cases tied to it to later?

Sounds like yes.

bwiernik:
That would be my vote.

bwiernik:
(And this is also something that a program like Zotero could do on their own GUI-wise in terms of how they construct their CSL JSON files.)

denismaier:
> Personally, I don't know that the locale-specification is fully necessary, and I think that it might best be left to a potential future multilingual module with expanded multilingual features. 

Sounds good in theory. But what exactly does "multilingual module" mean? Are we talking about Zotero, right? If yes, what about other implementations? Or could we think of modularizing CSL? That would be a really interesting idea...

> A single normal field would meet the most common use case of researchers who primarily work in one language. Sticking with a flat field would also let this variable be merged into v1.0.2.
Being able to add this right away is a convincing argument, I agree. And yes, it will cover many use cases. As you've written in the [other thread,](https://github.com/citation-style-language/zotero-bits/issues/50#issuecomment-633630127) "the most common case is authors publishing in two languages—English and their local language." If that is correct, and I think it is, then we could just solve that problem by adding two `translated-title` variables: `translated-title--en`, `translated-title--other`, both flat.

denismaier:
> (And this is also something that a program like Zotero could do on their own GUI-wise in terms of how they construct their CSL JSON files.)

Right. That could even be defined on a per document basis, just like in Jurism.

PR #242, opened 2020-06-11
Modify id to encourage UUIDs

### LINKED ISSUE #239: Modify id doc and/or pattern to encourage use of uuid
I worked with Dan on the spec language [here](https://github.com/citation-style-language/documentation/pull/82) for [this issue](https://discourse.citationstyles.org/t/proposal-make-style-ids-immutable/1478/34).

That language retains `cs:id`, but recommends new styles set that value to a uuid.

Two questions:

1. beyond changing the docstring, what, if any, change on the schema pattern?
2. which release?

On 1, there are a few options:

1. do nothing; xsd:anyURI will alllow a uuid
2. explicitly define a uuid pattern option (see below)
3. I contemplated an attribute for the type of id (like `<id type="uuid">`), but maybe overkill

I think these details primarily matter if you're a style author using a validating editor, and obviously when we want to be strict on styles for public release.

```
uuid = xsd:string { pattern = "[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}" }
```

bdarcus:
## Description

Modify the docstring to recommend UUIDs for new styles.

Closes #239 

Coupled with citation-style-language/documentation#82 doc change.

## Type of change

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -126,8 +126,9 @@ div {
     }
   info.id =
     
-    ## Specify the URI to establish the identity of the style. The URI
-    ## should be stable, unique and dereferenceable URI.
+    ## Specify the unique and stable identifier for the style. A URI
+    ## is valid, but new styles should use a UUID to ensure stability 
+    ## and uniqueness.
     element cs:id { xsd:anyURI }
   info.issn =
     


bdarcus:
I'm thinking to just merge this to master, since it only changes the docstring.

PR #245, opened 2020-06-14
Resolve merge conflicts on v1.1

bwiernik:
Was basically a bunch of line break references that got mangled somehow. 

I think this should work to merge. If not, you could temporarily give me commit permissions and I can push myself to the upstream v1.1


bdarcus:
Thanks much.

Yeah, I could see the pervasive whitespace issues. Perhaps this was introduced because we commits before we switched to trang to avoid these going forward?

bdarcus:
> Was basically a bunch of line break references that got mangled somehow.

Wait; I misread you.

Is it possible one of us has an editor set differently than the others?

bwiernik:
No, I don't mean like `CRLF`, I meant, like one edit had inserted a line break and another had deleted one.

Issue #241, opened 2020-06-11
Redundant id key in json schema?


bdarcus:
Why is there this additional `id` key under `citationItems`? Isn't that redundant?

Or redundant with a purpose; easier processing?

Example:

``` json
{
   "schema":"https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json",
   "citationID":"item1",
   "citationItems":[
      {
         "id":"item1",
         "itemData":{
            "id":"item1",
            "type":"book",
            "title":"test title"
         }
      }
   ]
}
```

_Originally posted by @bdarcus in https://github.com/citation-style-language/schema/pull/233_

denismaier:
Had a look here: https://citeproc-js.readthedocs.io/en/latest/csl-json/markup.html
But it's not explained there.

bdarcus:
@fbennett - can you answer this?

Also, while I have you, note we did some updating to the schema recently; including it schema id.

denismaier:
I've had a look at this again: I think the reason is quite simple: 

```
{
   "schema":"https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json",
   "citationID":"item1",
   "citationItems":[
      {
         "id":"item1",
         "itemData":{
            "id":"item1",
            "type":"book",
            "title":"test title"
         }
      }
   ]
}
```

Under `itemData` you really just have the complete item data, so the `id` is included here as well. One level above, you have all the information about a particular citation, locators and so. I imagine it works more or less like this: users specify the details of a citation, this gets stored directly under `citationItems`. Details about the reference item is then pulled from the database application and stored as is under `itemData`.

I am not sure, but I can imagine that tools like @rmzelle's [ref-extractor](https://rintze.zelle.me/ref-extractor/) probably rely on this. Also, this might be useful for collaboration or when your database goes corrupt.

rmzelle:
Both Zotero and Mendeley seem to store the same value in these two "id" fields in Word citation fields, so it does appear redundant. See https://github.com/rmzelle/ref-extractor/wiki for examples.

denismaier:
So that means we can drop one of them? Which one?
@dstillman @adomasven do you have additional input here?

bwiernik:
What about citations with multiple items?

denismaier:
@bwiernik what do you mean?

bwiernik:
Oh, I see, we are talking about the `id` under `citationItems` and the `id` under `itemData`, which are redundant. Got it.

[...]

PR #246, opened 2020-06-15
adds citekey variable to csl-data.json and csl-variables.rnc

### LINKED ISSUE #243: add citekey variable
Switching from word-processors to plain text based systems (pandoc, latex) is currently problematic since CSL currently doesn't know a user-definable citekey.

Additional Details: https://discourse.citationstyles.org/t/citekey-variable/1651/32

Suggestion per @bwiernik: 
An option could be to add a citekey variable to the CSL-JSON spec, with the indication that it could be used to in the same manner as a BibTeX citekey with id as a fallback if citekey is unavailable.

That would involve primarily minor changes with two players:

- pandoc adding citekey as primary source for the citation key. This would be easy enough as existing documents relying on id would still work.
- citation.js switching citation-label to citekey, which they should anyway.

denismaier:
## Description

This adds `citekey` to csl-data.json and csl-variables.rnc.

This should make switching from word-processors to plain text based systems (pandoc, latex) easier.

Closes #243

## Question
Do we need to add this also to csl-citation.json in addition to `id` [here](https://github.com/citation-style-language/schema/blob/68dff7c3d9906a7dc915b06409353e5ad28b8506/schemas/input/csl-citation.json#L24).

## Type of change

- [X] Variable Addition (adds a new variable or type string)

## Checklist

- [X] If I modified any of the schema files, I have run trang and/or prettier on the files, per [CONTRIBUTING](https://github.com/citation-style-language/schema/blob/master/CONTRIBUTING.md)
- [ ] I have included suggested corresponding changes to the documentation (if relevant)

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -58,6 +58,9 @@
       "id": {
         "type": ["string", "number"]
       },
+      "citekey": {
+        "type": "string"
+      },
       "categories": {
         "type": "array",
         "items": {
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -58,7 +58,7 @@
       "id": {
         "type": ["string", "number"]
       },
-      "citekey": {
+      "citation-key": {
         "type": "string"
       },
       "categories": {
--- a/schemas/styles/csl-variables.rnc
+++ b/schemas/styles/csl-variables.rnc
@@ -3,13 +3,13 @@ namespace a = "http://relaxng.org/ns/compatibility/annotations/1.0"
 
 ## Variables
 div {
-  
+
   ## All variables
   variables = variables.dates | variables.names | variables.standard
-  
+
   ## Standard variables
   variables.standard = variables.numbers | variables.strings
-  
+
   ## Date variables
   variables.dates =
     "accessed"
@@ -19,7 +19,7 @@ div {
     | "issued"
     | "original-date"
     | "submitted"
-  
+
   ## Name variables
   variables.names =
     "author"
@@ -43,7 +43,7 @@ div {
     | "recipient"
     | "reviewed-author"
     | "translator"
-  
+
   ## Number variables
   variables.numbers =
     "chapter-number"
@@ -62,7 +62,7 @@ div {
     | "printing"
     | "supplement"
     | "volume"
-  
+
   ## String variables
   variables.strings =
     "abstract"
@@ -74,6 +74,7 @@ div {
     | "authority"
     | "call-number"
     | "citation-label"
+    | "citekey"
     | "collection-title"
     | "container-title"
     | "container-title-short"
--- a/schemas/styles/csl-variables.rnc
+++ b/schemas/styles/csl-variables.rnc
@@ -3,13 +3,13 @@ namespace a = "http://relaxng.org/ns/compatibility/annotations/1.0"
 
 ## Variables
 div {
-
+  
   ## All variables
   variables = variables.dates | variables.names | variables.standard
-
+  
   ## Standard variables
   variables.standard = variables.numbers | variables.strings
-
+  
   ## Date variables
   variables.dates =
     "accessed"
@@ -19,7 +19,7 @@ div {
     | "issued"
     | "original-date"
     | "submitted"
-
+  
   ## Name variables
   variables.names =
     "author"
@@ -43,7 +43,7 @@ div {
     | "recipient"
     | "reviewed-author"
     | "translator"
-
+  
   ## Number variables
   variables.numbers =
     "chapter-number"
@@ -62,7 +62,7 @@ div {
     | "printing"
     | "supplement"
     | "volume"
-
+  
   ## String variables
   variables.strings =
     "abstract"
--- a/schemas/styles/csl-variables.rnc
+++ b/schemas/styles/csl-variables.rnc
@@ -74,7 +74,7 @@ div {
     | "authority"
     | "call-number"
     | "citation-label"
-    | "citekey"
+    | "citation-key"
     | "collection-title"
     | "container-title"
     | "container-title-short"


bdarcus:
> Do we need to add this also to csl-citation.json in addition to `id` [here](https://github.com/citation-style-language/schema/blob/68dff7c3d9906a7dc915b06409353e5ad28b8506/schemas/input/csl-citation.json#L24).

This is related to this question I had: #241?

bdarcus:
Do you mean to have this in draft?

denismaier:
Yes, I was thinking we need to settle this here first:

> Do we need to add this also to csl-citation.json in addition to `id` [here](https://github.com/citation-style-language/schema/blob/68dff7c3d9906a7dc915b06409353e5ad28b8506/schemas/input/csl-citation.json#L24).

bdarcus:
To make it easier to see the changes, could you please rebase, per [here](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#:~:text=To%20modify%20a%20commit%20that,moving%20them%20to%20another%20one.)?

You want to do `git rebase -i HEAD~2` in your case, though, since you only have two commits on this branch.

denismaier:
Question was: do we need to add the citekey variable to csl-citation.json?

bdarcus:
> Question was: do we need to add the citekey variable to csl-citation.json?

Yes, but who are we tagging to plea for input?

denismaier:
@adam3smith @fbennett @dstillman @jgm 
What do you think here? Do we need `citekey` in addition to `id` in `csl-citation.json`?

schemas/styles/csl-variables.rnc
```diff
@@ -74,6 +74,7 @@ div {
     | "authority"
     | "call-number"
     | "citation-label"
+    | "citekey"
```
bdarcus, line 77:
```suggestion
```

I don't understand why we need this in csl styling, when citation-label was intended for this case. Here's the definition from the spec:

> label identifying the item in in-text citations of label styles (e.g. "Ferr78"). May be assigned by the CSL processor based on item metadata

As in, adding `citekey` to the input data simply means that if the CSL processor does not assign, then it uses `citekey`.

Right?

bdarcus: Or is idea a style would need to test for it?

bwiernik: No, these really are completely different things. Ferr78 is not the citekey looked up in a data file, but rather a formatted citation in a style used primarily in math, statistics, and engineering.

The reason that a _style_ might have the citekey variable would be to facilitate transfer between GUI and plain text citation systems. For example, to go from a Word document with Zotero or Mendeley citations to an a document edited using pandoc, a style could be written that converts formatted citations into citekeys, affixes, and locators in the pandoc style.

This could be accomplished by instructing applications and processors to include a function to "unformat" citations, but a style is simple application-agnostic approach to that as well.

bdarcus: > Ferr78 is not the citekey **looked up in a data file**, but rather a formatted citation in a style used primarily in math, statistics, and engineering.

That's not what I mean above, which is only from the standpoint of the _formatted output_. 

From that standpoint, whether generated or not, the label is (has only ever been intended to be) a local human-readable token id for the reference item; eg:

```
foo bar [doe99]
```

... that then matches the equivalent in the bibliography:

```
[doe99] Doe, J ...
```

It's just, in a style with that output, the label is not (necessarily) generated.

So I'm still not understanding ;-)

If we add this to `cs:style`, we need a clear and simple explanation how it really differs from `citation-label`.

bdarcus: > we add this to `cs:style`, we need a clear and simple explanation how it really differs from `citation-label`, and maybe to change the definition of the latter.

Also, if we need both, maybe should use the same naming convention; say `citation-key`?

Not sure; just a thought.

bdarcus: PS - some history:

https://discourse.citationstyles.org/t/citation-label/761

bwiernik: Yes, I think we are agreeing that `citekey` and `citation-label` are not intended to mean the same thing at all.

The question we are having is whether the data-input-related `citekey` should be available for rendering in styles or only in the data and citation json schemas. I'm saying there is a specific use case for having it in styles: to facilitate transfer between GUI and text-based editors. 

I think we can make this clear in the documentation.

bdarcus:
> I'm saying there is a specific use case for having it in styles: to facilitate transfer between GUI and text-based editors.

Can you briefly explain this, here, or on the linked issue?

denismaier:
> > I'm saying there is a specific use case for having it in styles: to facilitate transfer between GUI and text-based editors.
> 
> Can you briefly explain this, here, or on the linked issue?

The idea is basically this: You edit in, say Word, with a regular citation style that gives you formatted citations. At some point you decide to switch to a plain-text system, e.g. pandoc . So, you'll switch to a citation style that renders the citation in pandoc's citation syntax: `[@doe, 34]`. For this, you need to access the citekey that will be used by pandoc.

bwiernik:
Yes, so for example. I am writing a paper in Word using Zotero. Zotero embeds CSL-JSON into each of its citations. Then, I decide to switch to pandoc. That's currently not easily possible and requires at least a fair number of hacks. What this would let me do is to reformat all of my citations to show the plain-text citation syntax `[@doe, 34]` and then extract the CSL JSON into one data file (e.g., using https://rintze.zelle.me/ref-extractor/). Then I can copy-paste my text into a text editor or convert the Word document to markdown using pandoc, etc.

bdarcus:
Thanks.

But couldn't `pandoc` simply read the embedded citations and convert it to markdown? 

In any case, at this point I'm less worried about this than about my last question on the linked issue; the need for both this and `citation-label`. The only reason I can see ATM for both is if GUI apps need to be able to pre-render citations, and want to store those in that variable.

bwiernik:
citation-label has nothing to do with citekey at all. There is no connection.

adam3smith:
The way I understand Bruce's question is that he's asking: if citation-label can be a customizable, human-readable label for a citation, why couldn't it be customizable to be equal to the citekey and thus satisfy the use case.

Am I understanding this correctly? 

I think that might technically work, but it's conceptually confusing and requires upstream apps to do some very weird gymnastics in generating the citation label. E.g. if you just allow people to manually specify the citation label, this won't work at all. CSL-JSON will include the manually specified citation label, not the otherwise existing citekey. 
If you have auto-generated citation labels that allow for different metadata-based rules (possibly specifiable in the citation style proper), then you'd have to have a completely separate setting for "ignore all rules and just use the citekey". 
Finally, to the extent that we are using CSL JSON as some sort of exchange format (and it is being used for that), using a field for two conceptually completely different things seems fraught with issues. I think conflating the two (even if technically possible) would be a bad idea.

denismaier:
> But couldn't `pandoc` simply read the embedded citations and convert it to markdown?

Maybe pandoc could do it, but I don't know if that's simple. This would most likely be implemented in the reader. There seems to be a [function](https://github.com/jgm/pandoc/blob/9672ff827194ddf5565351b2fdde942c03942b2d/src/Text/Pandoc/Readers/Docx.hs#L447) that performs some sort of normalization for fields.
@jgm What do you think?

But, of course, it's not only about pandoc. There's org, latex, etc.

bdarcus:
@bwiernik:

> citation-label has nothing to do with citekey at all. There is no connection.

From the standpoint of CSL design history, this is simply not true. 

As I said in that 2009 thread, we added the "citation-label" to put citekey data in. _That's_ the connection.

If you want to say that was a bad decision; that's fine, and in retrospect, I'd agree with you. But we can't ignore that history and legacy.

@adam3smith:

> The way I understand Bruce's question is that he's asking: if citation-label can be a customizable, human-readable label for a citation, why couldn't it be customizable to be equal to the citekey and thus satisfy the use case.
> 
> Am I understanding this correctly?

Basically, yes. 

But per above, I'm also saying that the input `citation-label` variable IS, already, an implicit alias for `citekey`.

> I think that might technically work, but it's conceptually confusing and requires upstream apps to do some very weird gymnastics in generating the citation label ...

So I think we all agree the confusion is using `citation-label` to refer to different things on input vs output.

But I don't believe anyone has yet to answer why we can't just do this

1. add `citation-key` or `citekey` to the input json
2. remove `citation-label` from the input json (since it would then serve no purpose for input)
3. add `citation-key` to the output style

Then documentation would be something like:

- `citation-key`: the  locally-unique human-readable id associated with the entry input data
- `citation-label`: the human-readable id generated by the CSL processor to associate a formatted citation with its reference entry

bwiernik:
> But I don't believe anyone has yet to answer why we can't just do this
> 
> add citation-key or citekey to the input json
> remove citation-label from the input json (since it would then serve no purpose for input)
> add citation-key to the output style
> Then documentation would be something like:
> 
> citation-key: the locally-unique human-readable id associated with the entry input data
> citation-label: the human-readable id generated by the CSL processor to associate a formatted citation with its reference entry

That's the proposal on the table no, with the understanding that `id` remains in place as a fallback if `citation-key` is absent from input?

bdarcus:
But (I think) you said earlier `citation-label` must stay, as is.

That's been the remaining thing that's holding this up for me.

Did I misunderstand, or did you change your mind?

And the PR doesn't currently remove it on input.

schemas/styles/csl-variables.rnc
bdarcus, line 77:
```suggestion
    | "citation-key"
```
For symmetry with `citation-label`.

schemas/input/csl-data.json
```diff
@@ -58,6 +58,9 @@
       "id": {
         "type": ["string", "number"]
       },
+      "citekey": {
```
bdarcus, line 61:
```suggestion
      "citation-key": {
```
See comment below.

I'm just proposing; not insisting ;-)

denismaier:
> And the PR doesn't currently remove it on input.

The question is if `citation-label` will always and without exception be generated by the processor. If so, there's indeed no need for this on the input side of things.

denismaier:
Concerning:

> Do we need to add this also to csl-citation.json in addition to `id` [here](https://github.com/citation-style-language/schema/blob/68dff7c3d9906a7dc915b06409353e5ad28b8506/schemas/input/csl-citation.json#L24).

I wouldn't add it at the moment. In citations I don't really see a need for both. Pandoc or other text-based systems will use `citation-key` as `id`, other tools might use `id`, but in both cases we are just talking about an identificator.

bdarcus:
> > And the PR doesn't currently remove it on input.
> 
> The question is if `citation-label` will always and without exception be generated by the processor. If so, there's indeed no need for this on the input side of things.

That's right.

As I may have mentioned, the only reason I can see for it is in a GUI app that wants to store the rendered label for some reason.

bdarcus:
We can remove label on input on a separate PR, if needed.

PR #247, opened 2020-06-15
CI: Add a check for schema formatting

bdarcus:
This addition to the CI tests simply checks if changes to the rnc and 
json schema files conform to our formatting expectations.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -18,7 +18,7 @@ jobs:
     - name: Install prettier
       run: npx prettier@2.0.5 . --write
     - name: Check json formatting
-      run: prettier --check schemas/input/*.json
+      run: npx prettier --check schemas/input/*.json
     - name: Run Tests
       run: pytest --verbose --color=yes tests
   csl-xml:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -23,6 +23,7 @@ jobs:
       uses: actions/checkout@v2
     - name: Install dependencies
       run: sudo apt-get install --yes trang
-    - name: Run Tests
-      run: trang schemas/styles/csl.rnc csl.rng
-
+    - name: Check schema validity
+      run: trang schemas/styles/csl.rnc csl.rng && trang csl.rng csl.rnc
+    - name: Check schema formatting
+      run: for f in csl*.rnc; do diff -q $f schemas/styles/$f ; done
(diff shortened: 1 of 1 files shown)


denismaier:
Nice. So no CI needed, right? Will this be installed automatically? Also, if there is no trang this will throw errors. And of course, it needs bash...

bdarcus:
> Nice. So no CI needed, right? Will this be installed automatically? Also, if there is no trang this will throw errors. And of course, it needs bash...

We still need CI, because this relies on people to install the hook themselves.

It will either throw an error, or simply not do the check. I didn't check.

It's not a very smart script.

The pre-commit project seems like a better option, as I mentioned, where we could do something, say, with python, and so some additional stuff.

But my coding skills are minimal.

bdarcus:
I decided to remove the hook and put it in #247.

dhimmel:
> Could do the same for the json schema, but that may be less important. Do we want this? Argument in favor: should avoid any funky diffs or whitespace-related merge conflicts.

I think so. It makes sense to enforce a style when multiple people are collaborating on a single JSON or XML file. If I remember from the past, we had some messy diffs due to trailing whitespace... would have been best to prevent that whitespace from ever getting merged.

bdarcus:
> > Could do the same for the json schema, but that may be less important. Do we want this? Argument in favor: should avoid any funky diffs or whitespace-related merge conflicts.
> 
> I think so. It makes sense to enforce a style when multiple people are collaborating on a single JSON or XML file. If I remember from the past, we had some messy diffs due to trailing whitespace... would have been best to prevent that whitespace from ever getting merged.

Yeah, my thought as well.

On the pre-commit hook, I'm using the default prettier hook for formatting json files.

Do you know of a similarly simple analog for what [I'm doing to check the rnc files](https://github.com/citation-style-language/schema/pull/247/files#diff-afb7705aa4398060eff1c01d7007947dR29)?

Need to install prettier, run it on the json files, and then diff the output.

dhimmel:
> Need to install prettier, run it on the json files, and then diff the output.

`prettier --check` looks like it's designed for this application https://prettier.io/docs/en/cli.html#--check

bdarcus:
> > Need to install prettier, run it on the json files, and then diff the output.
> 
> `prettier --check` looks like it's designed for this application https://prettier.io/docs/en/cli.html#--check

And what's the best way to install prettier itself on the action config?

dhimmel:
> what's the best way to install prettier itself on the action config

I think you can add the step:

```yaml
- uses: actions/setup-node@v1
```

Then probably just install prettier via one of these methods https://prettier.io/docs/en/install.html

You could do this in a new job or one of the existing jobs.

bdarcus:
OK, thanks.

Looks like I got it working.

PR #248, opened 2020-06-16
Add pre-commit config hook

bdarcus:
Coupled with #247, this adds pre-commit hook config to limit messy git diffs (and I think unnecessary merge conflicts).

Uses prettier for json (and yaml) linting, a couple of other little things.

Also defines our own script for the trang stuff; one bash (which works and is the one specified in the configuration), and the other python (what I'd prefer, I think, but I wasted far more time on this and it still doesn't work correctly).

Note: typically with `pre-commit`, the idea is the check runs, and if it fails (including if the linter modifies the file), the user manually fixes the files, then does `git add`.

For the typical case, this makes sense (given wide support for linting tools for most languages, including integrated in editing environments). But I don't think it does for the rnc files. So the script does this for you; it auto fixes your formatting "problems" during commit.

To install them, do the following from your local repo:

``` console
chmod +x hooks/rnc-format-validate.sh
pip install pre-commit
pre-commit install
```

TBD here: 

1. bash vs python
2. any other checks we want to do? add markdown to prettier? add proselint for the (pretty minimal here) text?

@dhimmel


bdarcus:
This now works with the bash script, but it should only run if the csl files are touched.

Would prefer to use the python script instead, but not done yet.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
+---
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.5.0
+    hooks:
+      - id: check-merge-conflict
+      - id: trailing-whitespace
```
bdarcus, line 7: I should probably remove this, since we're leaving this ti to the linters.

hooks/rnc-format-validate.py
```diff
@@ -0,0 +1,17 @@
...
+import subprocess
+import sys
+
+def main():
+    changed_files = subprocess.check_output([
+        "git", "diff", "--cached", "--name-only", "--diff-filter=.rnc"
+    ]).decode().strip().split('\n')
+
+    print("Changed schema files:", changed_files)
+
```
bdarcus, line 11: if we go with python version, need to add the core logic here

hooks/rnc-format-validate.sh
```diff
@@ -0,0 +1,19 @@
...
+    exit 1
+fi
+
+# iterate through changed rnc files
+for rncfile in $(git diff --cached --name-only | egrep '\.rnc$')
+do
+  rngfile="${rncfile//+(*\/|.*)}.rng"
+  trang $rncfile $rngfile && trang $rngfile $rncfile
+  rm -f $rngfile
+  git add $rncfile
```
bdarcus, line 17: Typically with `pre-commit`, the idea is the check runs, and if it fails, the user manually fixes the files, then does `git add`.

For the typical case, this make sense. But I don't think it does for this case. So the script does for you.

In essence, it auto fixes your formatting problems.

hooks/rnc-format-validate.sh
```diff
@@ -0,0 +1,19 @@
...
+  exit 1
+fi
+
+# iterate through changed rnc files
+for rncfile in $(git diff --cached --name-only | grep -E '\.rnc$')
+do
+  rngfile="${rncfile//+(*\/|.*)}.rng"
+  trang "$rncfile" "$rngfile" && trang "$rngfile" "$rncfile"
+  rm -f "$rngfile"
+  git add "$rncfile"
```
bdarcus, line 17: Here we add the commit so it doesn't fail, and so auto-fix the formatting of the file(s).

hooks/rnc-format-validate.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
+import subprocess
+import subprocess, sys
+from pathlib import Path
+
+def main():
+    changed_files = subprocess.check_output([
+        "git", "diff", "--cached", "--name-only", "--diff-filter=.rnc"
```
bdarcus, line 8: this is wrong; need to fix

dhimmel:
> this adds pre-commit hook config to limit messy git diffs (and I think unnecessary merge conflicts).

Good idea. The JSON and XML files should be linted. A pre-commit hook will help frequent contributors automate linting. However, it won't prevent users from submitting PRs with messy formatting. Therefore, you might also want to enforce linting on CI... or you could just keep an eye out for malformed PRs.

bdarcus:
> However, it won't prevent users from submitting PRs with messy formatting. Therefore, you might also want to enforce linting on CI... or you could just keep an eye out for malformed PRs.

I address this (at least for the rnc files) on #247.

bdarcus:
I just merged the CI updates that add formatting checks on the rnc and json schemas (thanks @dhimmel!).

I plan to merge this later or tomorrow. Unless there's any major objection in favor of the python, I think I'll remove that in favor of the bash. We can always switch later.

hooks/rnc-format-validate.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def main():
+
+    proc = subprocess.Popen(
+        ["git", "diff", "--name-only", "--cached"], stdout=subprocess.PIPE
+    )
+    changed_files = proc.stdout.readlines()
+    changed_files = [f.decode("utf-8") for f in changed_files]
+    changed_files = [f.strip() for f in changed_files]
```
dhimmel, line 13: You can probably achieve the same thing with:

```suggestion
    cmd = ["git", "diff", "--name-only", "--cached"]
    changed_files = subprocess.check_output(cmd, text=True).splitlines()
```

hooks/rnc-format-validate.py
```diff
@@ -0,0 +1,24 @@
...
+
+    # I don't understand why, but this currently isn't working
+    for rncfile in changed_files:
+        if rncfile.endswith(".rnc"):
+            rngfile = Path(rncfile).stem
+            subprocess.call(["trang", rncfile])
+            subprocess.call(["trang", rngfile, rncfile])
+            subprocess.call(["git", "add", rncfile])
```
dhimmel, line 21: `subprocess.check_call` to ensure there's a zero exit status.... not sure if that's what you want for the hook

bdarcus: The idea is that the script fixes the formatting and finishes the commit without failing.

You're suggesting here changing all `call` -> `check_call`?

dhimmel: If you want the script to fail if the subprocess calls fail, then I'd use `.check_call`. Otherwise, I'd use `.run`.

bdarcus: The issue I was having with this was the checks were passing, and the commit made, but the trang formatting wasn't applied.

dhimmel: probably git adding `Path(rncfile).stem` doesn't work. Should add full filename including suffixes

bdarcus: I'm just trying to the basename there, but I guess we don't need human readable filenames for this!

bdarcus: Fixed.

bdarcus:
It seems to me the python and bash scripts are now equivalent.

bdarcus:
OK, merged the python-based version. Please test, and if you see problems, or room for enhancement, submit a PR.

PR #252, opened 2020-06-16
Deprecate sentence case

bwiernik:
We expect data to be input in sentence case because reliably converting to sentence case isn't possible without BibTeX-style escaping. Will remove in v1.1.

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -1093,7 +1093,7 @@ div {
         ## Renders text in title case.
         "title"
       | 
-        ## Renders text in sentence case.
+        ## Renders text in sentence case. Deprecated.
         "sentence"
     }?
 }


bdarcus:
Adding @fbennett

jgm:
I'm just wondering what the specific problem is with converting to sentence case.
CSL JSON seems to support spans with class `nocase` which can be used to mark things that should not be lowercased on conversion to sentence case.  I've got a function that seems to do this reliably, but maybe there are things I'm missing?

denismaier:
Sentence case to title case is supposed to be easier as you can exclude stop words so there us no need for manual mark up.
Title case to sentence case means proper names have to be escaped.

bwiernik:
@jgm Basically it's that bibliographic data are never supplied with `<span class="nocase">` tags in it, so protecting things that shouldn't be lowercased requires user editing of data. Generally, CSL has taken the philosophy of limiting the expectations placed on users for generating correct citations. CSL is also used in a lot of automated contexts (e.g., the citations on osf.io), so ensuring accurate citations without user markup is important. 

All styles in the CSL repository are written assuming that titles are in sentence case, and repository policy prohibits its use.

bdarcus:
Crap; I accidentally introduced a typo in the commit summary on the merge.

It's minor, so guess not worth fixing, given it would rewrite the commit?

jgm:
Standard practice in bibtex/biblatex is to store titles in title case.
See https://tex.stackexchange.com/questions/20335/proper-casing-in-citation-bibliography-titles-using-biblatex-biber
"biblatex only has a macro to convert text to sentence case. There is no command to convert anything to Title Case. The implicit assumption is that people store titles in Title Case and that a style can convert titles to sentence case if need be. (BibTeX works similarly.)"

It looks like you're going for the opposite:  you have no macro to convert to sentence case, and you expect titles to be stored in sentence case.

I wonder if the decision to deprecate sentence-case may make it harder for people who want to use bibtex bibliographies with citeproc tools.

bwiernik:
Every CSL citation style is written with the assumption that data are stored in sentence case (`text-case="sentence"` is never used in any style in the CSL repository), so this is already functionally the situation. The practice of, for example, the Better BibTeX translator in Zotero is to convert titles from title case to sentence case on import. I'm not sure what pandoc does when it imports Bib(La)TeX data currently, but if it is stored in title case, any CSL style would need it to be converted to converted to sentence case anyway.

jgm:
Yes, pandoc-citeproc also does an unTitleCase transformation.

Issue #178, opened 2020-04-20
add term for every item type


denismaier:
as per https://github.com/citation-style-language/zotero-bits/issues/70

Issue #240, opened 2020-06-11
Syntax to use terms in variable contents


bwiernik:
**Is your feature request related to a problem? Please describe.**
Two problems come up in a fair number of styles.
  - Styles have specific wording required for item descriptions, statuses, formats, etc. (cf. [here](https://github.com/citation-style-language/csl-evolution/issues/34)). For example, APA wants items published early online to be labeled "Advance online publication" before an volume/issue number is assigned. Currently, these either need to be hardcoded into a style (breaking localization) or rely on the user to enter them in their data (again breaking localization and potentially requiring style-specific data entry). 
  - There have been requests for some types of creators (e.g., cartographer, podcast host) that are conceptually just a type-specific form of author (e.g., a cartographer is author of a map). It's onerous for style creation and maintenance to add such specific author types. It would be nice to treat these creators as `author` while still having the more specific labels.

**Describe the solution you'd like**
Support for this syntax should be added. In <text variable="variable"/>`, if the data for the variable contains `term:term`, the localized term specified is rendered, rather than the literal text.

For example, if a style contains:
`<text variable="status"/>`

And the item contains:
`"status": "term:advance-online-publication"`

Then the localized value of the `advance-online-publication` term is rendered.

A client application could serve such data using radio buttons or dropdown menus giving discrete options, then filling the corresponding term on the backend.

denismaier:
Sounds good. Should that work in all variables?

In terms of impact: That won't break existing styles, right? But it would add major functionality. So it should be included in a x.x release, right? So 1.1 or 1.2?

bwiernik:
Yeah, it would be a major release, 1.1 or 1.2.

My initial thought is that it should work for any variable in `cs:text` or `cs:label`, but I haven't thought through any potential downsides there.

denismaier:
Ok. Sounds like we're heading towards something like a macro system. Interesting...

denismaier:
This would also be a solution for https://github.com/citation-style-language/csl-evolution/issues/4

PR #257, opened 2020-06-17
Item type terms

denismaier:
## Description

Some style require descriptive terms added for certain item types. This poses challenges for localization. This adds a term for every item type to `csl-terms.rnc`.

Fixes https://github.com/citation-style-language/schema/issues/178

See also https://github.com/citation-style-language/zotero-bits/issues/70

## Type of change

- [X] Term addition (adds a new terms

## Checklist

### DIFF
--- a/schemas/styles/csl-terms.rnc
+++ b/schemas/styles/csl-terms.rnc
@@ -7,11 +7,14 @@ div {
     terms.gender-assignable
     | terms.gender-variants
     | terms.locator
-    | 
+    |
+      ## Item types
+      item-types
+    |
       ## Contributor roles
       variables.names
     | "editortranslator"
-    | 
+    |
       ## Miscellaneous terms
       "accessed"
     | "ad"
@@ -42,26 +45,26 @@ div {
     | "retrieved"
     | "scale"
     | "version"
-    | 
+    |
       ## Punctuation
       "open-quote"
     | "close-quote"
     | "open-inner-quote"
     | "close-inner-quote"
     | "page-range-delimiter"
-    | 
+    |
       ## Seasons
       "season-01"
     | "season-02"
     | "season-03"
     | "season-04"
-    | 
+    |
       ## (legacy; remove in CSL 1.1)
       category.field
-  
+
   ## Terms to which a gender may be assigned
   terms.gender-assignable =
-    
+
     ## Months
     "month-01"
     | "month-02"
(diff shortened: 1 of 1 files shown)


schemas/styles/csl-terms.rnc
```diff
@@ -7,11 +7,14 @@ div {
     terms.gender-assignable
     | terms.gender-variants
     | terms.locator
-    | 
+    |
+      ## Item types
+      item-types
```
bdarcus: I don't think there's any benefit to adding the doc string there.

```suggestion
    | item-types
```

denismaier: Done

bdarcus:
This seems a simple and uncontroversial change. Ready to merge.

Any other feedback?

bwiernik:
We had also discussed adding `type` as a special term where `<text term="type"/>` would render the term corresponding to the item's `type` (e.g., Map for a `map` item, Journal Article for an `article-journal`). @cormacrelf suggested that it would be a miniscule change for processors to add such a hash table and that it could go into 1.0.2.

denismaier:
> We had also discussed adding `type` as a special term where `<text term="type"/>` would render the term corresponding to the item's `type` (e.g., Map for a `map` item, Journal Article for an `article-journal`). @cormacrelf suggested that it would be a miniscule change for processors to add such a hash table and that it could go into 1.0.2.

If I remember correctly, we've agreed to do this in two steps. First terms, lookup logic later.
(I'll add another PR later or tomorrow with target 1.1)

bwiernik:
That's fine by me.

denismaier:
Ok. So how do we proceed. I'd prepare the next pr once this one is merged. Or can I already do that now? (Not sure how the branching strategy is supposed to work in this case.)

bwiernik:
You can open a pull request against v1.1

denismaier:
Hmm, maybe the extended logic will be covered by https://github.com/citation-style-language/schema/issues/240
So, actually we won't need a special PR for this. WDYT?

bwiernik:
No, these are unrelated. `<term = "type"/>` is something that would be written in a style.

denismaier:
Ok. Got it.

bdarcus:
so this is thus ready to merge?

denismaier:
Yes. It's ready to be merged. I'll open another issue with target 1.1 later. That will cover the remains bit.

PR #260, opened 2020-06-18
Add new cs:condition element to choose

### LINKED ISSUE #255: extend condition syntax

Condition statements in official CSL take a single “match” attribute, which determines how the tests will be combined. The match attribute (all, any, none) applies to all tests within the statement: grouping of tests with separate match values is not possible. This makes combining different conditions unnecessarily complicated and often leads to highly nested and unreadable structures inside `cs:if`.

CSL-M has introduced an optional extended syntax for condition statement:
> The alternative syntax may be applied to `cs:if` or `cs:else-if` elements (the “parent node” in this description). The parent node must have no attributes, and a single cs:conditions node as its first child element. The `cs:conditions` node must have one or more cs:condition children. The `cs:condition` children each define a conditional statement with attributes specified in the CSL 1.0.1 schema and in this Supplement. The `cs:condition` statements are joined according to a mandatory “match” attribute on `cs:conditions`.

Example:
``` xml
<choose>
  <if>
    <conditions match="any">
      <condition type="chapter"/>
      <condition variable="container-title collection-title" match="nand"/>
    </conditions>
    <text macro="some-chapter-mac"/>
  </if>
</choose>
```

CSL-M also adds `match="nand"` which tests true if at least one of the tests or condition statements to which it applies returns false.

I suggest we implement this solution, or something similar.

** Current capabilities

Apart from `match="nand"`, most (all?) tests can be done with standard CSL, but they may be less readable. This

``` xml
<choose>
  <if>
    <conditions match="any">
      <condition type="chapter"/>
      <condition variable="container-title collection-title" match="nand"/>
    </conditions>
    <text macro="some-chapter-mac"/>
  </if>
</choose>
```

needs a construction like this:

```xml
<choose>
  <if type="chapter"/>
    <text macro="some-chapter-mac"/>
  </if>
  <else-if variable="container-title collection-title" match="nand">
    <text macro="some-chapter-mac"/>
  <else-if>
</choose>
```

Now a real example (taken from chicago-note-bibliography.csl):

```xml
  <macro name="collection-title">
    <choose>
      <if match="none" type="article-journal">
        <choose>
          <if match="none" is-numeric="collection-number">
            <group delimiter=", ">
              <text variable="collection-title" text-case="title"/>
              <text variable="collection-number"/>
            </group>
          </if>
          <else>
            <group delimiter=" ">
              <text variable="collection-title" text-case="title"/>
              <text variable="collection-number"/>
            </group>
          </else>
        </choose>
      </if>
    </choose>
  </macro>
```

With the extended syntax that's a bit simpler:

```xml
<macro name="collection-title">
<choose>
  <if>
    <conditions match="all">
      <condition match="none" type="article-journal"/>
      <condition match="none" is-numeric="collection-number">
    </conditions>
    <group delimiter=", ">
      <text variable="collection-title" text-case="title"/>
      <text variable="collection-number"/>
    </group>
  </if>
  <else>
    <group delimiter=" ">
      <text variable="collection-title" text-case="title"/>
      <text variable="collection-number"/>
    </group>
  </else>
<choose>
</macro>
```

Note that this extended syntax is not necessarily shorter. But it will be with a third condition, and I'd also say it is more readable and maintainability is better due to DRY (the example above needs to call the same macro twice). 

**Additional context**
see [CSL-M supplement](https://citeproc-js.readthedocs.io/en/latest/csl-m/#cs-conditions-extension) from where I have copied this request almost verbatim.

** Addendum: Possible extension of the extension:
We could go even further and allow multiple `cs:conditions`.
Think of: 
if ((x = 5 or y = 10) and ((a = 1 not b = 2)  not (c = 3 or d = 4))) 

``` xml
<choose>
  <if match="all">
    <conditions match="any">
      <condition type="chapter"/>
      <condition variable="container-title collection-title" match="nand"/>
    </conditions>
    <conditions match="any">
      <condition has-day="issued"/>
      <condition date-range-variable"'issued" date-range-end='1900'/>
    </conditions>
    <text macro="some-other-mac"/>
  </if>
</choose>
```

@bdarcus @bwiernik @adam3smith Thoughts?

bdarcus:
## Description

This adds a modified version of the CSL-M condition element as an option
to configure more complex conditional rules.

Also splits off the cs:choose section into a separate file.

Closes #255

## RFC

Could people please test this, both the actual code, and also the docstrings, in a validating edtior (like Atom), and post feedback?

Will it be totally clear for style authors how this would work if we merge this?

## Type of change

### DIFF
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -1,0 +1,85 @@
+namespace a = "http://relaxng.org/ns/compatibility/annotations/1.0"
+namespace cs = "http://purl.org/net/xbiblio/csl"
+
+
+## cs:choose - Conditional Statements"
+div {
+  rendering-element.choose =
+    
+    ## Use to conditionally render rendering elements.
+    element cs:choose { choose.if, choose.else-if*, choose.else? }
+  
+  ## When specifying conditions, we allow either the attribute syntax or the element-based one.
+  choose.condition =
+    condition.atts
+    | element cs:conditions { match, condition.elem+ }
+  choose.if = element cs:if { choose.condition, rendering-element* }
+  choose.else-if =
+    element cs:else-if { choose.condition+, rendering-element* }
+  choose.else = element cs:else { rendering-element+ }
+  
+  ## This element allows for more complex conditional logic; for 
+  ## example, if you need to specify mixed "match" rules.
+  condition.elem = element cs:condition { match?, condition.atts }
+  condition.atts =
+    
+    ## If used, the element content is only rendered if it disambiguates two
+    ## otherwise identical citations. This attempt at disambiguation is only
+    ## made after all other disambiguation methods have failed.
+    [ a:defaultValue = "true" ] attribute disambiguate { "true" }
+    | 
+      ## Tests whether the given variables contain numeric text.
+      attribute is-numeric {
+        list { variables+ }
+      }
+    | 
+      ## Tests whether the given date variables contain approximate dates.
+      attribute is-uncertain-date {
+        list { variables.dates+ }
+      }
+    | 
+      ## Tests whether the locator matches the given locator types.
+      attribute locator {
+        list { (terms.locator.testable | "sub-verbo")+ }
+      }
+    | 
+      ## Tests whether the cite position matches the given positions.
+      attribute position {
+        list {
+          ("first"
+           | "subsequent"
+           | "ibid"
+           | "ibid-with-locator"
+           | "near-note")+
+        }
+      }
+    | 
+      ## Tests whether the item matches the given types.
+      attribute type {
+        list { item-types+ }
+      }
+    | 
+      ## Tests whether the default ("long") forms of the given variables
+      ## contain non-empty values.
+      attribute variable {
+        list { variables+ }
+      }
+  match =
+    
+    ## Set the testing logic.
+    [ a:defaultValue = "all" ]
+    attribute match {
+      
+      ## Element only tests "true" when all conditions test "true" for all
+      ## given test values.
+      "all"
+      | 
+        ## Element tests "true" when any condition tests "true" for any given
+        ## test value.
+        "any"
+      | 
+        ## Element only tests "true" when none of the conditions test "true"
+        ## for any given test value.
+        "none"
+    }?
+}
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -21,6 +21,7 @@ dc:description [
 ]
 
 ## Subparts of the CSL schema
+include "csl-choose.rnc"
 include "csl-terms.rnc"
 include "csl-types.rnc"
 include "csl-variables.rnc"
@@ -1062,85 +1063,6 @@ div {
       [ a:defaultValue = "ascending" ]
       attribute sort { "ascending" | "descending" }?
     }
-}
-# ==============================================================================
-
-## cs:choose - Conditional Statements"
-div {
-  rendering-element.choose =
-    
-    ## Use to conditionally render rendering elements.
-    element cs:choose { choose.if, choose.else-if*, choose.else? }
-
-  choose.if = element cs:if { condition+, match, rendering-element* }
-
-  choose.else-if =
-    element cs:else-if { condition+, match, rendering-element* }
-
-  choose.else = element cs:else { rendering-element+ }
-
-  condition =
-    
-    ## If used, the element content is only rendered if it disambiguates two
-    ## otherwise identical citations. This attempt at disambiguation is only
-    ## made after all other disambiguation methods have failed.
-    [ a:defaultValue = "true" ] attribute disambiguate { "true" }
-    | 
-      ## Tests whether the given variables contain numeric text.
-      attribute is-numeric {
-        list { variables+ }
-      }
-    | 
-      ## Tests whether the given date variables contain approximate dates.
-      attribute is-uncertain-date {
-        list { variables.dates+ }
-      }
-    | 
-      ## Tests whether the locator matches the given locator types.
-      attribute locator {
-        list { (terms.locator.testable | "sub-verbo")+ }
-      }
-    | 
-      ## Tests whether the cite position matches the given positions.
-      attribute position {
-        list {
-          ("first"
-           | "subsequent"
-           | "ibid"
-           | "ibid-with-locator"
-           | "near-note")+
-        }
-      }
-    | 
-      ## Tests whether the item matches the given types.
-      attribute type {
-        list { item-types+ }
-      }
-    | 
-      ## Tests whether the default ("long") forms of the given variables
-      ## contain non-empty values.
-      attribute variable {
-        list { variables+ }
-      }
-
-  match =
-    
-    ## Set the testing logic.
-    [ a:defaultValue = "all" ]
-    attribute match {
-      
-      ## Element only tests "true" when all conditions test "true" for all
-      ## given test values.
-      "all"
-      | 
-        ## Element tests "true" when any condition tests "true" for any given
-        ## test value.
-        "any"
-      | 
-        ## Element only tests "true" when none of the conditions test "true"
-        ## for any given test value.
-        "none"
-    }?
 }
 # ==============================================================================
 


bdarcus:
I have two commits here ATM: the base commit, and then one that makes `@match` required.

Could people test this, both the actual code, and also the docstrings, in a validating edtior (like Atom), and post feedback?

bdarcus:
PS - I'm getting confused by working with the different base branches.

bwiernik:
@bdarcus This would go into v1.1, no?

bdarcus:
> @bdarcus This would go into v1.1, no?

Yep; this is what I mean above by getting confused; when I switched the base branch.

But we can figure that out later.

schemas/styles/csl-choose.rnc
```diff
@@ -0,0 +1,84 @@
...
+  choose.condition = condition.atts | condition.elem
+  choose.if =
+    element cs:if { choose.condition+, match, rendering-element* }
+  choose.else-if =
+    element cs:else-if { choose.condition+, match, rendering-element* }
+  choose.else = element cs:else { rendering-element+ }
+  
+  ## This element allows for more complex conditional logic; for 
+  ## example, if you need to specify mixed "match" rules.
+  condition.elem = element cs:condition { match?, condition.atts }
```
bwiernik: I think there is usability value in forcing all `condition` elements to be grouped together by making them children of a single `condition` element. That would prevent something like, where a style writer misunderstands that all conditions apply to the single `if` and that `else-if` is needed for other cases (this `if` will always return false because an item cannot have two types):
```
<if>
  <condition type="article-journal"/>
  <text variable="title"/>
  <condition type="book"/>
  <text variable="title" text-case="title" font-face="italic"/>
</if>
```

Or this, where the style will work, but the exact logic is unclear. This will be a problem for others to maintain the style:
```
<if>
  <condition variable="editor" match="none"/>
  <text variable="collection-title" text-case="title" font-face="italic"/>
  <condition variable="editorial-director"/>
  <names variable="editorial-director">
    <label form="verb"/>
    <name form="long"/>
  </names>
</if>
```

It would be better to group all conditions together:
```
<if>
  <conditions>
    <condition variable="editor" match="none"/>
    <condition variable="editorial-director"/>
  </conditions>
  <text variable="collection-title" text-case="title" font-face="italic"/>
  <names variable="editorial-director">
    <label form="verb"/>
    <name form="long"/>
  </names>
</if>
```

schemas/styles/csl-choose.rnc
```diff
@@ -0,0 +1,84 @@
...
+
+## cs:choose - Conditional Statements"
+div {
+  rendering-element.choose =
+    
+    ## Use to conditionally render rendering elements.
+    element cs:choose { choose.if, choose.else-if*, choose.else? }
+  
+  ## When specifying conditions, we allow either the attribute syntax or the element-based one.
+  choose.condition = condition.atts | condition.elem
```
bwiernik, line 13: Does this permit _both_ attribute syntax or element syntax?

schemas/styles/csl-choose.rnc
bdarcus: > I think there is usability value in forcing all `condition` elements to be grouped together

I agree with this 100%, and have no strong opinions about adding a containing element, but have you confirmed the current code doesn't already group the elements together in sequence? 

That was my intention, and I think they do, but I'm having problems with my editors ATM, so haven't tested it.

schemas/styles/csl-choose.rnc
bdarcus: My intention is only allows one or the other. That's what we want; right?

But to repeat, I haven't tested it. Working on it.

denismaier: Yes, one or the other.

schemas/styles/csl-choose.rnc
denismaier: Yes, I think this forces grouping. This [RNC cheatsheet](https://gist.github.com/miikka/b137a8705e88b0b6c164#file-cheatsheet-rnc-L35) says: "if the order of the elements does not matter, use "&" instead of ","" => "," means order is important and all `condition` elements must come first. Right?

schemas/styles/csl-choose.rnc
```diff
@@ -19,7 +19,7 @@ div {
   
   ## This element allows for more complex conditional logic; for 
   ## example, if you need to specify mixed "match" rules.
-  condition.elem = element cs:condition { match?, condition.atts }
+  condition.elem = element cs:condition { match, condition.atts }
```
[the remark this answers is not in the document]
bdarcus: I mean, can do whatever you guys think best, and I'll defer to you, since you deal more with style authoring and editing than I ever have. 

I was just hoping to make the rules clear in the validation specification, and thinking that this wouldn't require the element:

``` xml
<choose>
  <if>
    <condition type="book"/>
    <condition variable="title"/>
  </if>
</choose>
```

... which would just be:

``` xml
<choose>
  <if type="book" variable="title">
    ...
  </if>
</choose>
```

But there may not be value in restricting.

schemas/styles/csl-choose.rnc
bdarcus: Yes.

schemas/styles/csl-choose.rnc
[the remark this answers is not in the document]
denismaier: Well, but this

> ```
> <choose>
>   <if>
>     <condition type="book chapter" match="none"/>
>     <condition variable="title"/>
>   </if>
> </choose>
> ```

Can't be transformed into the current syntax without adding nested `choose` elements.

schemas/styles/csl-choose.rnc
bdarcus: I just tested on Atom, and it works as I expected. 

For the first choice after `if` or `if-then`, the only choice is `condition`. 

If you then add a `text` element, `condition` is no longer a choice.

But OTOH, I think if we're in doubt, it's better to err on the side of adding back the `containers` wrapper.

Thoughts before I resolve this?

schemas/styles/csl-choose.rnc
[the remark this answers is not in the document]
bdarcus: Correct, because the match conditions differ.

I wonder if we made a mistake allowing a default match value.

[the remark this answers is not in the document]
bdarcus: I'll revert this commit.

bdarcus:
I resolve your requested changes @bwiernik, but I'm not sure how to formally do that in the review system ATM.

In any case, should be ready to test.

bdarcus:
> > @bdarcus This would go into v1.1, no?
> 
> Yep; this is what I mean above by getting confused; when I switched the base branch.
> 
> But we can figure that out later.

@bwiernik - if you have time in the next few days, can you please figure out how to properly get this setup for merging in v1.1?

I'm hoping to we can close 1.0.2 work ASAP to avoid some of this confusion as we wrap up 1.1.

bwiernik:
Just set the base branch to v1.1 (I just did that).

bdarcus:
How weird; when I did that earlier, it resulted in some funky results. Maybe subsequently rebase resolved that.

schemas/styles/csl-choose.rnc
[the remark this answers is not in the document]
bwiernik: No, leave it. To be clear, I **do think** that the default should be `match="all"` for `if`, `else-if`, and `condition`. I just am not so familiar with the rnc syntax to know if that's the case here.

schemas/styles/csl-choose.rnc
bwiernik: Yes, one or the other. Just want to be sure that's enforced.

bdarcus:
> How weird; when I did that earlier, it resulted in some funky results. Maybe subsequent changes resolved that.

Actually, no i see what I earlier saw, which confuses me.

See all those commits on this PR?

schemas/styles/csl-choose.rnc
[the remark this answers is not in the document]
bdarcus: RelaxNG, by design, has no notion of default values and such. That's only something specified in documentation, etc.

What I changed was requiring `@match`.

I think we'll only know for sure on these details with testing.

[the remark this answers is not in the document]
bwiernik: Okay great, then do leave it as optional

bwiernik:
> > How weird; when I did that earlier, it resulted in some funky results. Maybe subsequent changes resolved that.
> 
> Actually, now i see what I earlier saw, which confuses me.
> 
> See all those commits on this PR?
> 
> It looks to me like this has become the new 1.1 branch!

No, I don't see the commits in this PR on the v1.1 branch: https://github.com/citation-style-language/schema/commits/v1.1

bdarcus:
So what _should_ I do here?

Checkout the v1.1 branch, then checkout a new branch from there, then cherry-pick commits specific to this PR?

bdarcus:
Anyone have a chance to test this? 

I did, but am not the best person to test it.

Hoping to merge later today or tomorrow.

denismaier:
Will be able to test on Monday.
  
  
   
    bdarcus <notifications@github.com> hat am 20.06.2020 17:20 geschrieben:
   
   
    
   
   
    
   
   
   Anyone have a chance to test this?
   I did, but am not the best person to test it.
   Hoping to merge later today or tomorrow.

bwiernik:
Bruce, when you find that your fork or local copy of a repo is screwed up, it’s often best to test from the upstream master:
```
git reset --hard upstream/master
```

bdarcus:
I decided to merge it. We'll have to test 1.1 anyway.

denismaier:
I tried to test v1.1, but I have errors here. Does that branch work for you?

bdarcus:
I see an error now. Trying to figure it out, and how it got there.

bdarcus:
I just pushed a bug fix. Try again.

Was basically mistakes that we made earlier, that trang did not pick up.

Trang seems to only give errors when there are syntax mistakes.

bdarcus:
I think there's another bug.

bdarcus:
Now fixed.

denismaier:
Ok, it seems to work now.

denismaier:
But: we still need to wrap the the conditions under `cs:conditions`, right? Haven't we agreed that this is not strictly necessary?

bdarcus:
Auto-complete should require the cs:citations wrapper.

BTW, for a variety of reasons, it would be very helpful if you guys could
figure out how to describe, in natural language, how to transform from
attribute to element form.

I can then add that to the upgrade xslt, and we can in turn use it to add
content to the test suite, docs, etc.

denismaier:
> Auto-complete should require the cs:citations wrapper.

Currently, yes. But is that on purpose? I think we've said we can do without.

bdarcus:
It's a decision we can revisit. It's not required, but I'm agnostic,
but I figured better to err on the side of caution.

denismaier:
His main point was that the conditions should appear an bloc at the beginning.  That would also be required without the additional element. But I can live with that as well.

bwiernik:
Yeah, I don't care one way or the other so long as all `condition` elements appear together.

PR #261, opened 2020-06-19
add build script for variable list

### LINKED ISSUE #223: Add a script to update json schema strings from rnc files
> How would that look like and how would that solve the problem with schema verbosity? Would it?

A python script in GitHub Actions could compile the csl-data.json at commit time. It would have a list of all types and variables (separated by category) and dynamically construct the JSON. The main benefit would be ease of maintenance and updating, not needing to manually keep four nearly identical lists aligned manually.

_Originally posted by @bwiernik in https://github.com/citation-style-language/schema/issues/167#issuecomment-639037322_

denismaier:
## Description

Updating variable lists is a tedious task, especially since each variable has multiple variants (e.g. `-short`, `original-`, `reviewed-`), and we have to keep multiple lists in sync (`csl-variables.rnc` and `csl-data.json`)

This scripts aims to simply this task by creating multiple variants for each variable from a single source file. Variables are added to a `csl-variables.yaml` under `/schemas/variables`. For each variable, add only the basic form, i.e., only `title`, not `title-short`, `original-title`, `reviewed-title`, as these derivative forms will be generated automatically.

A couple of remarks:
- In its current form, this script only produces csl-variables.rnc. JSON output will be added later.
- The list of variables in `csl-variables.yaml` does not reflect the current status. (At least, I'm not sure it does.)
- The script can already add variants that are planned for 1.1 (`-main` and `-sub` variants on title variables.) This is currently disabled.

Fixes #223  

(Note: I'm a bit confused by the additional changes that seem to be part of that PR. What's wrong here? I actually did merge upstream/master before creating a new branch.)


bdarcus:
You seem to have the same git problem I have on #260

denismaier:
> 14 changed files?

I don't know. It's the python script, the yaml, and the output csl-variables.rnc. But I certainly did not touch any of those other files.

PR #262, opened 2020-06-19
tools: Merge rnc formatting/validations scripts

bdarcus:
### Description

This merges the previous bash formatting and validation script with the
python-based pre-commit hook script.

Run without argument from the base directory of the repo to format the
schema files.

### Notes and Questions

I was just playing with this to see how it'd work. 

Should we merge?

If yes, should I delete the bash script, or keep it?

It works, but there are still a few minor issues I need to address.


denismaier:
Should we add a makefile to simplify those script calls? (could be also useful for the variable list script)

bdarcus:
Probably. 

So you're thinking something like `make schemas` could build the schema files, and format them?

denismaier:
Yep.

bdarcus:
Also feasible to merge this and the build script.

Does make work on Windows?

denismaier:
Yes. There are a couple of options. Easiest is to install with chocolatey.

denismaier:
> Also feasible to merge this and the build script.

Right. Good idea.

bdarcus:
Or if not formally merge, use python to do the integration, so could do something like `tools/makeschemas`.

bdarcus:
You have any thoughts on this @dhimmel? See also #263.

tools/rnc-validate-format.py
```diff
@@ -0,0 +1,60 @@
...
+    except Exception:
+        print("trang commnd not found; please install")
+
+    rng_new = rncfile + "_new.rng"
+    rnc_new = rncfile + "_new.rnc"
+
+    # round-trip the rnc file through trang
+    subprocess.run(["trang", rncfile, rng_new])
+
+    if check:
```
bdarcus, line 29: not sure this is the right approach; it's assuming moving back to the pre-commit mere;y checking the formatting, and I'm not even sure this is the best way to do that. should probably remove the `if`.

tools/rnc-validate-format.py
```diff
@@ -0,0 +1,60 @@
...
+parser = argparse.ArgumentParser(description="process csl schema files")
+parser.add_argument("-c", action="store_true", help="commit the file(s)")
+args = parser.parse_args()
+
+
+def rnc_format(rncfile, check=False):
+    try:
+        which("trang") is not None
+    except Exception:
+        print("trang commnd not found; please install")
```
dhimmel, line 21: this catches the error and continues. You probably want to quit if trang is not installed? Like

```suggestion
    except Exception:
        raise SystemExit("trang commnd not found; please install")
```

tools/rnc-validate-format.py
```diff
@@ -0,0 +1,60 @@
...
+import subprocess
+import argparse
+from shutil import which
+from glob import glob
+
+# to validate and pretty print the rnc schema files before commit:
+# tools/rnc-validate-format.py
+
+parser = argparse.ArgumentParser(description="process csl schema files")
+parser.add_argument("-c", action="store_true", help="commit the file(s)")
```
dhimmel, line 13: i think long args help with readability, especially for stuff that is not going to be frequently typed but read a lot:

```suggestion
parser.add_argument("--commit", action="store_true", help="commit the file(s)")
```

bdarcus:
I have a couple minor things to fix, and will then merge.

I won't delete the bash script for now, and we can keep this separate from the other script, and integrate them through a third script.

PR #263, opened 2020-06-19
tools: Add build script for variable list

### LINKED ISSUE #223: Add a script to update json schema strings from rnc files

denismaier:
## Description

Updating variable lists is a tedious task, especially since each variable has multiple variants (e.g. `-short`, `original-`, `reviewed-`), and we have to keep multiple lists in sync (`csl-variables.rnc` and `csl-data.json`)

This scripts aims to simply this task by creating multiple variants for each variable from a single source file. Variables are added to a `csl-variables.yaml` under `/schemas/variables`. For each variable, add only the basic form, i.e., only `title`, not `title-short`, `original-title`, `reviewed-title`, as these derivative forms will be generated automatically.

A couple of remarks:
- The list of variables in `csl-variables.yaml` does not reflect the current status. (At least, I'm not sure it does.)
- The script can already add variants that are planned for 1.1 (`-main` and `-sub` variants on title variables.) This is currently disabled.

Fixes #223


tools/build-variable-schemas.py
```diff
@@ -0,0 +1,135 @@
...
+    titles = titles_stringified,
+    strings= strings_stringified,
+    )
+
+# Write to csl-variables.rnc
+
+rnc_filename = '../schemas/styles/csl-variables.rnc'
+rnc_variables_file= open(rnc_filename,'w')
+rnc_variables_file.write(rnc_template)
+rnc_variables_file.close()
```
bdarcus, line 135: there's newline warning here; could you run the script though, say, black?

denismaier: Better now?

bdarcus:
When you're ready, share this with dhillman?

schemas/variables/csl-variables.yaml
```diff
@@ -0,0 +1,97 @@
...
+variables.titles:
+  - collection-title
+  - container-title
+  - part-title
+  - volume-title
+  - title
+  - translated-title
+variables.json:
+  - shortTitle
+  - journalAbbreviation
```
bdarcus, line 97: Another warning from github.

I'm trying to setup CI integration to use prettier to check the yaml files, so you might run that on this?

denismaier: done

denismaier:
Update: Json schema now build as well. Perhaps not super elegant, but it should be a start. We can refactor later.

tools/build-variable-schemas.py
```diff
@@ -0,0 +1,172 @@
...
+## Dates
+dates = create_original_reviewed_variants(dates_input)
+
+## Names
+names = create_original_reviewed_variants(names_input)
+
+## Numbers
+numbers = create_original_reviewed_variants(numbers_input)
+
+## Stings
```
bwiernik, line 54: `## Strings`

bdarcus:
> Update: Json schema now build as well. Perhaps not super elegant, but it should be a start. We can refactor later.

Do you want to undraft?

@bwiernik - should we merge?

denismaier:
Ok. I'll remove draft.
People should nevertheless do some tests. Hope I haven't missed something.

denismaier:
Other questions:
1. How will people use that? Run manually? Github actions? 
2. Are file locations ok?

bdarcus:
@dhimmel - you ok with this?

With this, we'd move to maintaining variable lists in a tank file, and use
this script to generate the json and rnc files, and potentially also parts
of the specification.

bdarcus:
I don't know the right technical solution, but I'd like to only worry about
updating the YAML file, including adding the docstrings to it.

As in, we merge, and everything gets updated for us.

bdarcus:
BTW, I merged the update to the CI workflow.

Can you replace the current prettier section command for "*.json" with
"{*.json,*.yaml}" on your PR?

denismaier:
@bdarcus You mean like updating the appendix in the specs as well? Could be possible with docs and schema in one repo. But there are costs, of course.

dhimmel:
> This scripts aims to simply this task by creating multiple variants for each variable from a single source file. Variables are added to a `csl-variables.yaml` under `/schemas/variables`. For each variable, add only the basic form, i.e., only `title`, not `title-short`, `original-title`, `reviewed-title`, as these derivative forms will be generated automatically.

What do these prefixes/suffixes mean? I understand short, but what's the purpose of original and reviewed? Are these currently part of the spec, or does this PR create them? It seems like a lot of fields to add to the spec.

Will this PR add `keyword`, `keyword-short`, `original-keyword`, `reviewed-keyword` to the spec? What do those mean?

Rather than duplicating fields to such an extent, couldn't you change the syntax to allow:

```yaml
keyword:
  full: xxxx
  short: yyyy
  original: zzzz
  reviewed: xxxx
```

This is less backwards compatible.

bdarcus:
Couldn't the doc repo remain separate, but pull in these changes some way?

denismaier:
Original and short are coming from here: https://github.com/citation-style-language/schema/issues/167#issuecomment-646311574

You are right about keyword, those variants don't make sense. Getting rid of short is easy here, just move it to another section in the yaml.

Concerning original and reviewed: maybe I'll add a new array to the yaml for such variables. 
I have thought your syntax as well, but won't that require that we add each variable multiple times? I was hoping not having to do that.

bdarcus:
Pretty sure you can create a definition once, and then reference it.

denismaier:
Hmm maybe something like this could work:

```yaml
title: 
  - prefixes:
    - reviewed
    - original
  - suffixes 
    - -short
```
But isn't that not more complicated on the input side?

bdarcus:
Spitballing; what about:

All current variables stay the same, but we add -ext variants for the new
stuff?

title
title-ext
    original

This is getting complicated though.

I was also worrying about the explosion of strings, even as I earlier
implied we shouldn't worry about it.

bdarcus:
Spitballing; what about something like this, at least as a first step?

All current variables stay the same, but we add `-ext` or `-alt` or `-variant` variants for the new
stuff? This would be the output json.

``` js
{
  "title": "title",
  "title-ext": {
    "original": "original title"
  }
}
```

Could be coupled with a new `cs:text@variant` or similar on csl proper?

This is getting complicated though.

I was also worrying about the explosion of strings, even as I earlier
implied we shouldn't worry about it.

bdarcus:
> Hmm maybe something like this could work:
> 
> ```yaml
> title: 
>   - prefixes:
>     - reviewed
>     - original
>   - suffixes 
>     - -short
> ```

The prefixes are really relations, and the suffixes are alternate forms.

bwiernik:
I think we are overthinking this. `original-` variables describe the original version of an item. Currently, there are `original-` versions of title, author, publisher, and publisher-place, which works for reprinted books but isn't enough for other item types (like a reprinted article or book chapter). Similarly, `reviewed-` variables describe the item being reviewed (e.g., a book, film, or article). These are needed, for example, to give different formats for reviews of books versus films versus articles. 

For these, I think we should explicitly not create lots of extra lists that have to be managed—I don't know if `keyword` is ever actually used, but what difference does it make if `original-keyword` or `reviewed-keyword` exists? If someone has a use for `keyword`, then they probably have a use for the original-language versions of those keywords too?

My original proposal was just to give processor instructions that any CSL variable might have `original-` or `reviewed-` appended to them. The downside there would be no autocompletion in editors. The upside would be shorter rnc and json schemas. I don't care one way or the other, but I think `original-` and `reviewed-` should be handled very simply--"every CSL variable [and also 'type'] should be able to have a `reviewed-` or `original-` prefix added to it."

Given the choice, my preference would be to skip the autocompletion and keep shorter schemas.

denismaier:
Does it really matter how many variables we have in the schema if we build it with a script?

bwiernik:
I don't think it impacts readability of the schema if all of the "reviewed-" variables are grouped together and labeled as "any variable may have reviewed- appended to it to refer to the item being reviewed". Similarly for "original-". I think that general rule is much easier to manage and implement than discrete lists of these variables.

dhimmel:
> Does it really matter how many variables we have in the schema if we build it with a script?

Yes, because users still look at the schema to understand the format. Also once you've officially added a field, it becomes harder to remove it or rename it. So by adding all these fields to the spec, you're complicating the user experience considerably and making your job as maintainers harder in the future.

> I think that general rule is much easier to manage and implement than discrete lists of these variables.

I think the general rule is okay, especially as a convention. There are near infinite extensions to CSL JSON that may make sense for specific use cases. The schema doesn't need to list every possible extension as long as it doesn't preclude them.

denismaier:
I am not sure I undersrand: so you are saying it's posdible to use variables that are not in the spec? How is that?

dhimmel:
> you are saying it's posdible to use variables that are not in the spec?

For the CSL JSON Data schema, if you switch the following to true:

https://github.com/citation-style-language/schema/blob/0a6bcb9d15d23f64bca7a4d212ebdb9e6879c798/schemas/input/csl-data.json#L492

That would allow additional fields.

> autocompletion? And validation?

Those would not occur, which could be somewhat of a problem. But the other end of the spectrum is also a problem: where the spec is cluttered with fields that are rarely used and that don't necessarily make sense in the future. So I think the solution is to make the spec flexible to accommodate custom fields. Once a custom field proves its usage in the wild, it can be added to the spec.

I am not so against adding these fields to the schema if they already have substantial usage... but I do think it opens the floodgates. Some users might need `title-translation-en`, `title-translation-fr`, etcetera. So I think I am just worried about adding some fields that would justify adding almost any field. Especially when there is so much repetition.

Don't want to hold up progress, since I don't have the time to offer an actual solution at the moment.

denismaier:
> Those would not occur, which could be somewhat of a problem. But the other end of the spectrum is also a problem: where the spec is cluttered with fields that are rarely used and that don't necessarily make sense in the future. So I think the solution is to make the spec flexible to accommodate custom fields. Once a custom field proves its usage in the wild, it can be added to the spec.

I see your point. The problem is that currently it is impossible to upload non-validating styles to the Zotero styles repo. Also, Zotero will complain when you install such a style. Using non-spec variables will work to a certain degree with citeproc-js (it will treat unknown fields as strings though. If you need names, dates or numbers---don't know if that is possible. Pandoc will just ignore non-spec variables

denismaier:
> I don't think it impacts readability of the schema if all of the "reviewed-" variables are grouped together and labeled as "any variable may have reviewed- appended to it to refer to the item being reviewed". Similarly for "original-".

So, your are saying I should add original and reviewed variables to dedicated groups? Right now, they just appear at the end of the category that they belong to (titles, strings, numbers, dates, names). I could change that. Don't know what is better.

Also, if we will use this, we will probably need to check where this produces useless variants.

tools/build-variable-schemas.py
denismaier: Done

bdarcus:
You want to make the script executable, I think.

``` console
$ git update-index --chmod=+x tools/build-variable-schemas.py
```

bdarcus:
> The problem is that currently it is impossible to upload non-validating styles to the Zotero styles repo. Also, Zotero will complain when you install such a style. Using non-spec variables will work to a certain degree with citeproc-js (it will treat unknown fields as strings though. If you need names, dates or numbers---don't know if that is possible. Pandoc will just ignore non-spec variables.

Technically, it should be pretty easy to add extension options to the schemas. 

We just haven't done it, largely to avoid the compatibility headaches it would cause.

I think that's a larger conversation in any case.

denismaier:
Original-DOI and reviewed-ISBN? Really?

Anyway, looks like we really need to discuss the conceptual issue first before we talk about the technical implementation (the script).

bdarcus:
Yes, the script is good.

Maybe we should move the substantive discussion to the linked issue.

>

denismaier:
By the way, enabling -sub and -main variables for titles will give us again a lot more variables.
  
  
   
    bdarcus <notifications@github.com> hat am 20.06.2020 14:35 geschrieben:
   
   
    
   
   
    
   
   Yes, the script is good.
   
   Maybe we should move the substantive discussion to the linked issue.

bwiernik:
> Original-DOI and reviewed-ISBN? Really?

Sure. The ISBN of the original book. Or the DOI of the original version of the article (e.g., for the original article published in a Spanish language journal and now translated and reprinted in English).

denismaier:
I think we should first reach a consensus on #167 and then discuss automated solutions, but sub and main as forms instead of variables sounds good.

bdarcus:
Given recent merges and decisions, we should probably close this?

PR #264, opened 2020-06-19
CI: refactor

bdarcus:
Defines a separate prettier job and runs on both json and yaml files.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,8 +1,20 @@
-name: Python Test Suite
+name: CSL CI Test Suite
 on:
 - push
 - pull_request
 jobs:
+  prettier:
+    name: prettier
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout 
+      uses: actions/checkout@v2
+    - name: Set up npm
+      uses: actions/setup-node@v1
+    - name: Install prettier
+      run: npx prettier@2.0.5 . --write
+    - name: check json and yaml formatting
+      run: npx prettier --check schemas/**/*.{json,yaml}
   csl-json:
     name: CSL JSON Tests
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -1,8 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout 
+      uses: actions/checkout@v2
+    - name: Set up npm
+      uses: actions/setup-node@v1
+    - name: Install prettier
+      run: npx prettier@2.0.5 . --write
+    - name: check json and yaml formatting
+      run: npx prettier --check schemas/**/*.{json,yaml}
```
bdarcus, line 16: @dhimmel - github is complaining about a yaml syntax error here, but I've tried everything I can think of, and nothing works. Any ideas?

.github/workflows/test.yaml
```diff
@@ -1,8 +1,19 @@
-name: Python Test Suite
+name: CSL CI Test Suite
 on:
 - push
 - pull_request
 jobs:
+    name: prettier
```
dhimmel, line 6: you are missing a key

```suggestion
jobs:
  prettier:
    name: prettier
```

.github/workflows/test.yaml
dhimmel: see https://github.com/citation-style-language/schema/pull/264/files#r442869965

bdarcus: OMG; what am embarrassing oversight!

Thanks.

.github/workflows/test.yaml
```diff
@@ -1,8 +1,20 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout 
+      uses: actions/checkout@v2
+    - name: Set up npm
+      uses: actions/setup-node@v1
+    - name: Install prettier
+      run: npx prettier@2.0.5 . --write
+    - name: check json and yaml formatting
+      run: npx prettier --check schemas/**/*.{json,yaml}
```
bdarcus, line 17: this currently fails because prettier complains about missing yaml files

Issue #190, opened 2020-05-22
Add title and description annotations to JSON Schema elements


dhimmel:
From https://json-schema.org/understanding-json-schema/reference/generic.html#annotations

> JSON Schema includes a few keywords, `title`, `description`, `default`, `examples` that aren't strictly used for validation, but are used to describe parts of a schema.
> 
> None of these "annotation" keywords are required, but they are encouraged for good practice, and can make your schema "self-documenting".
> 
> The `title` and `description` keywords must be strings. A "title" will preferably be short, whereas a "description" will provide a more lengthy explanation about the purpose of the data described by the schema.

Currently, there is confusion over what certain variables in CSL JSON are supposed to containing. Setting `title` and `description` would be helpful in addressing this. Also it would allow for a single authoritative source of CSL JSON field documentation.

`examples` might also be helpful, but probably not as pressing:

> New in draft 6 The `examples` keyword is a place to provide an array of examples that validate against the schema. This isn't used for validation, but may help with explaining the effect and purpose of the schema to a reader. Each entry should validate against the schema in which is resides, but that isn't strictly required. There is no need to duplicate the `default` value in the `examples` array, since `default` will be treated as another example.

However, the test suite could extract and test the examples, so examples could also help improve test coverage.

bdarcus:
Would it be possible/wise to have a single source for definitions, and then programmatically propagate them to the different places they need to be?

For example, the RNC schema should have annotations for everything, but they need to be in the documentation too, and also (per this issue) would be helpful in the json schemas.

For that matter, could we programmatically update the csl-data.json file from the RNC schema?

bdarcus:
Well, now that I think about, I may be overstating how well the rnc schema is commented currently. We don't have annotations for the terms and such, which we should.

dhimmel:
@bdarcus I'm not familiar with what the RNC schemas are used for and what their overlap is with the CSL Data schema. Ideally, we wouldn't be defining the same schema in two formats unless one is entirely auto-generated from the other.

dhimmel:
> I may be overstating how well the rnc schema is commented currently.

The RNC schema should be describing the XML used to specify a CSL Style. So if a title/definition concerns an element of an XML style, then we should put the description in the RNC. But if it primarily concerns CSL JSON Data, we should use the JSON Schema. Does that delineation make sense?

bdarcus:
I'm just observing that in both the json and the csl, strings like variables and types are just lists, with slightly different syntax.

[...]

Issue #115, opened 2014-08-18
proposal: Add more names to support Council of Science Editors


serpodrick:
I would like to have these name variables added to CSL:

volume-editor
compiler
editor-translator

denismaier:
Shouldn't we add a `volume-editor`.
Or, should this be covered with `@relation`?
@bwiernik @bdarcus

bwiernik:
volume-editor is fine.

bdarcus:
> Or, should this be covered with `@relation`?

It's a relation.

bwiernik:
I’m thinking of things like big multivolume encyclopedias where there is an editor in chief and an editor for each volume. Right now, we just enter both as editor.

denismaier:
> I’m thinking of things like big multivolume encyclopedias where there is an editor in chief and an editor for each volume. Right now, we just enter both as editor.

Yes. That's the use case. The question is just if we introduce a new variable `volume-editor`, or if we do `names variable="editor" relation="volume"/>`. I think we should do the latter if we introduce the data-model with relations. Otherwise, `volume-editor` seems to be an obvious addition.

bwiernik:
Let's add `volume-editor`. I don't think we should rely on the hierarchical data structure more than is necessary.

bdarcus:
I think we should first settle the relation discussion, and the impact on
the data model, before deciding this.

denismaier:
> Let's add `volume-editor`. I don't think we should rely on the hierarchical data structure more than is necessary.

Let's say we did it like this. What about translators? We will need `container-translator`and `volume-translator`. That will get pretty messy...

bwiernik:
Honestly, I think any of these beyond vanilla `editor` and `translator` are rare enough to be edge cases. Translators especially so. I think we should not overthink this.

denismaier:
> Honestly, I think any of these beyond vanilla `editor` and `translator` are rare enough to be edge cases. Translators especially so. I think we should not overthink this.

That's why I think they are a good fit for the relation syntax. So we don't have to argue over `volume-compiler`, but the variable would still be available if someone needs it. It seems to me that this is the major advantage of this syntax.

bwiernik:
Yes. volume-editor is one that is fairly common and should be available regardless of whether there is a hierarchical data model

bdarcus:
I'm going to repeat, I want us to resolve the data model question first.

Because as a part of that, I think we should remove exactly these sorts of
variables.

[...]

PR #268, opened 2020-06-22
input: Add related-item modelling #268

bdarcus:
The CSL 1.0 style schema adopted a flat key-value data model for standard variables like titles and identifiers.

The 1.0 JSON schema was designed to provide a machine-readable way to feed CSL processors with a model that matched the style model.

Since then, CSL, and the use of CSL JSON, has exploded, as have the number of feature requests. 

Recent development work put us in a situation where we could have ended up with well over 100 variables to support new features (see #167), which does not seem practical.

This PR aims to make the input schema simpler and more flexible, without needing to add new variables whenever there's a feature request.

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -4,451 +4,7 @@
   "$id": "https://resource.citationstyles.org/schema/latest/input/json/csl-data.json",
   "type": "array",
   "items": {
-    "type": "object",
-    "properties": {
-      "type": {
-        "type": "string",
-        "enum": [
-          "article",
-          "article-journal",
-          "article-magazine",
-          "article-newspaper",
-          "bill",
-          "book",
-          "broadcast",
-          "chapter",
-          "classic",
-          "collection",
-          "dataset",
-          "document",
-          "entry",
-          "entry-dictionary",
-          "entry-encyclopedia",
-          "figure",
-          "graphic",
-          "hearing",
-          "interview",
-          "legal_case",
-          "legislation",
-          "manuscript",
-          "map",
-          "motion_picture",
-          "musical_score",
-          "pamphlet",
-          "paper-conference",
-          "patent",
-          "performance",
-          "periodical",
-          "personal_communication",
-          "post",
-          "post-weblog",
-          "regulation",
-          "report",
-          "review",
-          "review-book",
-          "software",
-          "song",
-          "speech",
-          "standard",
-          "thesis",
-          "treaty",
-          "webpage"
-        ]
-      },
-      "id": {
-        "$ref": "#/definitions/id-variable"
-      },
-      "categories": {
-        "type": "array",
-        "items": {
-          "type": "string"
-        }
-      },
-      "language": {
-        "type": "string"
-      },
-      "journalAbbreviation": {
-        "type": "string"
-      },
-      "shortTitle": {
-        "type": "string"
-      },
-      "author": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "chair": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "collection-editor": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "compiler": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "composer": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "container-author": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "contributor": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "curator": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "director": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "editor": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "editorial-director": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "executive-producer": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "interviewer": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "illustrator": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "organizer": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "original-author": {
-        "description": "[Deprecated - Use `original` related `author` property instead]",
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "performer": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "producer": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "recipient": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "reviewed-author": {
-        "description": "[Deprecated - Use `reviewed` related `author` property instead]",
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "translator": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
-        }
-      },
-      "accessed": {
-        "$ref": "#/definitions/date-variable"
-      },
-      "available-date": {
-        "$ref": "#/definitions/date-variable"
-      },
-      "container": {
-        "$ref": "#/definitions/date-variable"
-      },
-      "event-date": {
-        "$ref": "#/definitions/date-variable"
-      },
-      "issued": {
-        "$ref": "#/definitions/date-variable"
-      },
-      "original-date": {
-        "$ref": "#/definitions/date-variable"
-      },
-      "submitted": {
-        "$ref": "#/definitions/date-variable"
-      },
-      "abstract": {
-        "type": "string"
-      },
-      "annote": {
-        "type": "string"
-      },
-      "archive": {
-        "type": "string"
-      },
-      "archive_collection": {
-        "type": "string"
-      },
-      "archive_location": {
-        "type": "string"
-      },
-      "archive-place": {
-        "type": "string"
-      },
-      "authority": {
-        "type": "string"
-      },
-      "call-number": {
-        "type": "string"
-      },
-      "chapter-number": {
-        "type": ["string", "number"]
-      },
-      "citation-number": {
-        "type": ["string", "number"]
-      },
-      "citation-label": {
-        "type": "string"
-      },
-      "collection-number": {
-        "type": ["string", "number"]
-      },
-      "collection-title": {
-        "type": "string"
-      },
-      "container-title": {
-        "type": "string"
-      },
-      "container-title-short": {
-        "type": "string"
-      },
-      "dimensions": {
-        "type": "string"
-      },
-      "division": {
-        "type": "string"
-      },
-      "DOI": {
-        "type": "string"
-      },
-      "edition": {
-        "type": ["string", "number"]
-      },
-      "event": {
-        "type": "string"
-      },
-      "event-place": {
-        "type": "string"
-      },
-      "first-reference-note-number": {
-        "type": ["string", "number"]
-      },
-      "genre": {
-        "type": "string"
-      },
-      "ISAN": {
-        "type": "string"
-      },
-      "ISBN": {
-        "type": "string"
-      },
-      "ISCI": {
-        "type": "string"
-      },
-      "ISMN": {
-        "type": "string"
-      },
-      "ISRC": {
-        "type": "string"
-      },
-      "ISSN": {
-        "type": "string"
-      },
-      "ISWC": {
-        "type": "string"
-      },
-      "issue": {
-        "type": ["string", "number"]
-      },
-      "jurisdiction": {
-        "type": "string"
-      },
-      "keyword": {
-        "type": "string"
-      },
-      "locator": {
-        "type": ["string", "number"]
-      },
-      "medium": {
-        "type": "string"
-      },
-      "note": {
-        "type": "string"
-      },
-      "number": {
-        "type": ["string", "number"]
-      },
-      "number-of-pages": {
-        "type": ["string", "number"]
-      },
-      "number-of-volumes": {
-        "type": ["string", "number"]
-      },
-      "original": {
-        "title": "Original Related Item",
-        "$ref": "#/definitions/related-item"
-      },
-      "original-publisher": {
-        "description": "[Deprecated - Use `original` related `publisher` property instead]",
-        "type": "string"
-      },
-      "original-publisher-place": {
-        "description": "[Deprecated - Use `original` related `publisher-place` property instead]",
-        "type": "string"
-      },
-      "original-title": {
-        "description": "[Deprecated - Use `original` related `title` property instead]",
-        "type": "string"
-      },
-      "page": {
-        "type": ["string", "number"]
-      },
-      "page-first": {
-        "type": ["string", "number"]
-      },
-      "part": {
-        "type": ["string", "number"]
-      },
-      "part-title": {
-        "type": "string"
-      },
-      "PMCID": {
-        "type": "string"
-      },
-      "PMID": {
-        "type": "string"
-      },
-      "printing": {
-        "type": ["string", "number"]
-      },
-      "publisher": {
-        "type": "string"
-      },
-      "publisher-place": {
-        "type": "string"
-      },
-      "references": {
-        "type": "string"
-      },
-      "reviewed": {
-        "title": "Reviewed Related Item",
-        "$ref": "#/definitions/related-item"
-      },
-      "reviewed-title": {
-        "description": "[Deprecated - Use `reviewed` related `title` property instead]",
-        "type": "string"
-      },
-      "scale": {
-        "type": "string"
-      },
-      "section": {
-        "type": "string"
-      },
-      "source": {
-        "type": "string"
-      },
-      "status": {
-        "type": "string"
-      },
-      "supplement": {
-        "type": ["string", "number"]
-      },
-      "title": {
-        "type": "string"
-      },
-      "title-short": {
-        "type": "string"
-      },
-      "translated-title": {
-        "type": "string"
-      },
-      "translated-title-short": {
-        "type": "string"
-      },
-      "URL": {
-        "type": "string"
-      },
-      "version": {
-        "type": "string"
-      },
-      "volume": {
-        "type": ["string", "number"]
-      },
-      "year-suffix": {
-        "type": "string"
-      },
-      "custom": {
-        "title": "Custom key-value pairs.",
-        "type": "object",
-        "description": "Used to store additional information that does not have a designated CSL JSON field. The note field can also store additional information, but custom is preferred for storing key-value pairs.",
-        "examples": [
-          {
-            "short_id": "xyz",
-            "other-ids": ["alternative-id"]
-          },
-          {
-            "metadata-double-checked": true
-          }
-        ]
-      }
-    },
-    "required": ["type", "id"],
-    "additionalProperties": false
+    "$ref": "#/definitions/refitem"
   },
   "definitions": {
     "edtf-string": {
@@ -587,14 +143,6 @@
     "id-variable": {
       "type": ["string", "number"]
     },
-    "related-item": {
-      "type": "object",
-      "$ref": "#/definitions/related-item-variable"
-    },
-    "related-item-variable": {
-      "description": "The id for a related item, to be included in the output.",
-      "$ref": "#/definitions/id-variable"
-    },
     "name-variable": {
       "anyOf": [
         {
@@ -630,6 +178,453 @@
           "additionalProperties": false
         }
       ]
+    },
+    "refitem": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string",
+          "enum": [
+            "article",
+            "article-journal",
+            "article-magazine",
+            "article-newspaper",
+            "bill",
+            "book",
+            "broadcast",
+            "chapter",
+            "classic",
+            "collection",
+            "dataset",
+            "document",
+            "entry",
+            "entry-dictionary",
+            "entry-encyclopedia",
+            "figure",
+            "graphic",
+            "hearing",
+            "interview",
+            "legal_case",
+            "legislation",
+            "manuscript",
+            "map",
+            "motion_picture",
+            "musical_score",
+            "pamphlet",
+            "paper-conference",
+            "patent",
+            "performance",
+            "periodical",
+            "personal_communication",
+            "post",
+            "post-weblog",
+            "regulation",
+            "report",
+            "review",
+            "review-book",
+            "software",
+            "song",
+            "speech",
+            "standard",
+            "thesis",
+            "treaty",
+            "webpage"
+          ]
+        },
+        "id": {
+          "$ref": "#/definitions/id-variable"
+        },
+        "categories": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "language": {
+          "type": "string"
+        },
+        "journalAbbreviation": {
+          "type": "string"
+        },
+        "shortTitle": {
+          "type": "string"
+        },
+        "author": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "chair": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "collection-editor": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "compiler": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "composer": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "container-author": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "contributor": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "curator": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "director": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "editor": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "editorial-director": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "executive-producer": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "interviewer": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "illustrator": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "organizer": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "original-author": {
+          "description": "[Deprecated - Use `original` related `author` property instead]",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "performer": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "producer": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "recipient": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "reviewed-author": {
+          "description": "[Deprecated - Use `reviewed` related `author` property instead]",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "translator": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "accessed": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "available-date": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "container": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "event-date": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "issued": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "original-date": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "submitted": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "abstract": {
+          "type": "string"
+        },
+        "annote": {
+          "type": "string"
+        },
+        "archive": {
+          "type": "string"
+        },
+        "archive_collection": {
+          "type": "string"
+        },
+        "archive_location": {
+          "type": "string"
+        },
+        "archive-place": {
+          "type": "string"
+        },
+        "authority": {
+          "type": "string"
+        },
+        "call-number": {
+          "type": "string"
+        },
+        "chapter-number": {
+          "type": ["string", "number"]
+        },
+        "citation-number": {
+          "type": ["string", "number"]
+        },
+        "citation-label": {
+          "type": "string"
+        },
+        "collection-number": {
+          "type": ["string", "number"]
+        },
+        "collection-title": {
+          "type": "string"
+        },
+        "container-title": {
+          "type": "string"
+        },
+        "container-title-short": {
+          "type": "string"
+        },
+        "dimensions": {
+          "type": "string"
+        },
+        "division": {
+          "type": "string"
+        },
+        "DOI": {
+          "type": "string"
+        },
+        "edition": {
+          "type": ["string", "number"]
+        },
+        "event": {
+          "type": "string"
+        },
+        "event-place": {
+          "type": "string"
+        },
+        "first-reference-note-number": {
+          "type": ["string", "number"]
+        },
+        "genre": {
+          "type": "string"
+        },
+        "ISAN": {
+          "type": "string"
+        },
+        "ISBN": {
+          "type": "string"
+        },
+        "ISCI": {
+          "type": "string"
+        },
+        "ISMN": {
+          "type": "string"
+        },
+        "ISRC": {
+          "type": "string"
+        },
+        "ISSN": {
+          "type": "string"
+        },
+        "ISWC": {
+          "type": "string"
+        },
+        "issue": {
+          "type": ["string", "number"]
+        },
+        "jurisdiction": {
+          "type": "string"
+        },
+        "keyword": {
+          "type": "string"
+        },
+        "locator": {
+          "type": ["string", "number"]
+        },
+        "medium": {
+          "type": "string"
+        },
+        "note": {
+          "type": "string"
+        },
+        "number": {
+          "type": ["string", "number"]
+        },
+        "number-of-pages": {
+          "type": ["string", "number"]
+        },
+        "number-of-volumes": {
+          "type": ["string", "number"]
+        },
+        "original": {
+          "title": "Original Related Item",
+          "$ref": "#/definitions/refitem"
+        },
+        "original-publisher": {
+          "description": "[Deprecated - Use `original` related `publisher` property instead]",
+          "type": "string"
+        },
+        "original-publisher-place": {
+          "description": "[Deprecated - Use `original` related `publisher-place` property instead]",
+          "type": "string"
+        },
+        "original-title": {
+          "description": "[Deprecated - Use `original` related `title` property instead]",
+          "type": "string"
+        },
+        "page": {
+          "type": ["string", "number"]
+        },
+        "page-first": {
+          "type": ["string", "number"]
+        },
+        "part": {
+          "type": ["string", "number"]
+        },
+        "part-title": {
+          "type": "string"
+        },
+        "PMCID": {
+          "type": "string"
+        },
+        "PMID": {
+          "type": "string"
+        },
+        "printing": {
+          "type": ["string", "number"]
+        },
+        "publisher": {
+          "type": "string"
+        },
+        "publisher-place": {
+          "type": "string"
+        },
+        "references": {
+          "type": "string"
+        },
+        "reviewed": {
+          "title": "Reviewed Related Item",
+          "$ref": "#/definitions/refitem"
+        },
+        "reviewed-title": {
+          "description": "[Deprecated - Use `reviewed` related `title` property instead]",
+          "type": "string"
+        },
+        "scale": {
+          "type": "string"
+        },
+        "section": {
+          "type": "string"
+        },
+        "source": {
+          "type": "string"
+        },
+        "status": {
+          "type": "string"
+        },
+        "supplement": {
+          "type": ["string", "number"]
+        },
+        "title": {
+          "type": "string"
+        },
+        "title-short": {
+          "type": "string"
+        },
+        "translated-title": {
+          "type": "string"
+        },
+        "translated-title-short": {
+          "type": "string"
+        },
+        "URL": {
+          "type": "string"
+        },
+        "version": {
+          "type": "string"
+        },
+        "volume": {
+          "type": ["string", "number"]
+        },
+        "year-suffix": {
+          "type": "string"
+        },
+        "custom": {
+          "title": "Custom key-value pairs.",
+          "type": "object",
+          "description": "Used to store additional information that does not have a designated CSL JSON field. The note field can also store additional information, but custom is preferred for storing key-value pairs.",
+          "examples": [
+            {
+              "short_id": "xyz",
+              "other-ids": ["alternative-id"]
+            },
+            {
+              "metadata-double-checked": true
+            }
+          ]
+        }
+      },
+      "required": ["type"],
+      "additionalProperties": false
     }
   }
 }
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -164,6 +164,7 @@
         }
       },
       "original-author": {
+        "description": "[Deprecated - Use `original` related `author` property instead]",
         "type": "array",
         "items": {
           "$ref": "#/definitions/name-variable"
@@ -188,6 +189,7 @@
         }
       },
       "reviewed-author": {
+        "description": "[Deprecated - Use `reviewed` related `author` property instead]",
         "type": "array",
         "items": {
           "$ref": "#/definitions/name-variable"
@@ -338,12 +340,15 @@
         "type": ["string", "number"]
       },
       "original-publisher": {
+        "description": "[Deprecated - Use `original` related `publisher` property instead]",
         "type": "string"
       },
       "original-publisher-place": {
+        "description": "[Deprecated - Use `original` related `publisher-place` property instead]",
         "type": "string"
       },
       "original-title": {
+        "description": "[Deprecated - Use `original` related `title` property instead]",
         "type": "string"
       },
       "page": {
@@ -377,6 +382,7 @@
         "type": "string"
       },
       "reviewed-title": {
+        "description": "[Deprecated - Use `reviewed` related `title` property instead]",
         "type": "string"
       },
       "scale": {
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -339,6 +339,10 @@
       "number-of-volumes": {
         "type": ["string", "number"]
       },
+      "original": {
+        "title": "Original Related Item",
+        "$ref": "#/definitions/related-item"
+      },
       "original-publisher": {
         "description": "[Deprecated - Use `original` related `publisher` property instead]",
         "type": "string"
@@ -380,6 +384,10 @@
       },
       "references": {
         "type": "string"
+      },
+      "reviewed": {
+        "title": "Reviewed Related Item",
+        "$ref": "#/definitions/related-item"
       },
       "reviewed-title": {
         "description": "[Deprecated - Use `reviewed` related `title` property instead]",
@@ -423,10 +431,6 @@
       },
       "year-suffix": {
         "type": "string"
-      },
-      "related": {
-        "type": "array",
-        "items": [{ "$ref": "#/definitions/related-item" }]
       },
       "custom": {
         "title": "Custom key-value pairs.",
@@ -585,18 +589,11 @@
     },
     "related-item": {
       "type": "object",
-      "properties": {
-        "original": {
-          "$ref": "#/definitions/related-item-variable"
-        },
-        "reviewed": {
-          "$ref": "#/definitions/related-item-variable"
-        }
-      }
+      "$ref": "#/definitions/related-item-variable"
     },
     "related-item-variable": {
       "description": "The id for a related item, to be included in the output.",
-      "oneOf": [{ "$ref": "#/definitions/id-variable" }]
+      "$ref": "#/definitions/id-variable"
     },
     "name-variable": {
       "anyOf": [
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -393,6 +393,7 @@
           "$ref": "#/definitions/date-variable"
         },
         "original-date": {
+          "description": "[Deprecated - Use `original` related date property instead]",
           "$ref": "#/definitions/date-variable"
         },
         "submitted": {
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -56,7 +56,7 @@
         ]
       },
       "id": {
-        "type": ["string", "number"]
+        "$ref": "#/definitions/id-variable"
       },
       "categories": {
         "type": "array",
@@ -418,6 +418,10 @@
       "year-suffix": {
         "type": "string"
       },
+      "related": {
+        "type": "array",
+        "items": [{ "$ref": "#/definitions/related-item" }]
+      },
       "custom": {
         "title": "Custom key-value pairs.",
         "type": "object",
@@ -517,13 +521,27 @@
           "id": "range-1",
           "type": "book",
           "title": "The Title",
-          "issued": [{ "year": 2000 }, { "year": 2001 }]
+          "issued": [
+            {
+              "year": 2000
+            },
+            {
+              "year": 2001
+            }
+          ]
         },
         {
           "id": "range-2",
           "type": "book",
           "title": "The Title",
-          "issued": [{ "year": 2000 }, { "undefined": true }]
+          "issued": [
+            {
+              "year": 2000
+            },
+            {
+              "undefined": true
+            }
+          ]
         }
       ],
       "items": {
@@ -555,6 +573,24 @@
           "$ref": "#/definitions/date-structured"
         }
       ]
+    },
+    "id-variable": {
+      "type": ["string", "number"]
+    },
+    "related-item": {
+      "type": "object",
+      "properties": {
+        "original": {
+          "$ref": "#/definitions/related-item-variable"
+        },
+        "reviewed": {
+          "$ref": "#/definitions/related-item-variable"
+        }
+      }
+    },
+    "related-item-variable": {
+      "description": "The id for a related item, to be included in the output.",
+      "oneOf": [{ "$ref": "#/definitions/id-variable" }]
     },
     "name-variable": {
       "anyOf": [


denismaier:
So, this will be only about `@related` and `form="sub"` will be covered by a separate PR?

bdarcus:
Yes; makes it easier all around, I think.

>

denismaier:
Will you add one?

bdarcus:
I will.

>

bdarcus:
Given the other activity on the input side, and my recognition that csl input data has room to grow, I think that strengthens the case for relations on the input side.

bwiernik:
Just want to link to this discussion of related issues: https://github.com/citation-style-language/schema/issues/115#issuecomment-651855588

schemas/input/csl-data.json
```diff
@@ -4,425 +4,811 @@
...
+        "issued": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "original-date": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "submitted": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "original": {
```
bdarcus, line 578: Here's where I added the two new related item properties.

schemas/input/csl-data.json
```diff
@@ -4,425 +4,811 @@
...
-          "$ref": "#/definitions/name-variable"
+      "required": ["type", "id"],
+      "additionalProperties": false
+    },
+    "related-bibitem": {
+      "title": "Related bib item",
+      "description": "Used to define related bibliographic data; value can either be a refid string, or an embedded item object",
+      "anyOf": [
+        {
+          "type": "string"
```
bdarcus, line 805: The citekey confusion potentially screws this up

schemas/input/csl-data.json
```diff
@@ -4,425 +4,533 @@
...
+          }
+        },
+        "organizer": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "original-author": {
+          "description": "[Deprecated - Use `author` on the `original` related item instead]",
```
bdarcus, line 174: I added two kinds of deprecation annotations; here's an example of a suggestion I am making.

schemas/input/csl-data.json
```diff
@@ -4,425 +4,533 @@
...
+          "type": "string"
+        },
+        "archive": {
+          "type": "string"
+        },
+        "archive_collection": {
+          "type": "string"
+        },
+        "archive_location": {
+          "description": "[Deprecated - Q: what TODO?]",
```
bdarcus, line 246: This is the other type of deprecation annotation: where I am unsure.

bdarcus:
@bwiernik @denismaier - OK, I just pushed this commit for the json schema. Please review when you get a chance.

BTW, related to #190, I did the annotations with [this](https://coveooss.github.io/json-schema-for-humans/) in mind.

PS - there's conflicts, but I'm unclear why, and don't want to worry about that ATM.

bdarcus:
BTW, per discussion on a related issue, now would be a good time to test the changes we're contemplating using an auto--completing and validating editor, at least on the style end. It's important that someone using this be able to easily understand the UX.

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
```
bdarcus, line 243: Here's the other kind of deprecation annotation; just me asking a question of what, if anything, we should do.

bwiernik:
I think this is generally a good structure. I like it. 

With respect to the deprecation questions, generally, I think it would be good to maintain the existing variables as "first-class" variables that can be entered directly. Even if _all_ variables can't be entered as flat variables, the existing set is reasonably complete enough to permit simpler entry for many cases. This would especially be the case if `variable="reviewed-title"` is maintained as the syntax in styles.

Some of the specific questions I don't think we should put into the relation system at all (e.g., archive, event); I'll give individual line comments for those.

bdarcus:
> With respect to the deprecation questions, generally, I think it would be good to maintain the existing variables as "first-class" variables that can be entered directly.

Even ones we added in the past week?

In any case, maybe add a suggested line comment for how you would suggest explaining that? It could be confusing.

Also, for this, I think we need to get as much feedback as possible. Who should we tag?

bwiernik:
I think we've just added "reviewed-genre" in the past week right? I'll think about and make line comments/suggestions. 

Might be good to draft up some example items giving related items in the two formats to show how this would work/that it is feasible.

> Also, for this, I think we need to get as much feedback as possible. Who should we tag?
Dan Stillman, Frank, Sebastian, Rintze, John definitely. Perhaps Lars Willighagen and Emiliano? I'm not sure what the status of Cormac/citeproc-rs is. Anyone from Mendeley or ReadCube we should tag?

bdarcus:
Yes, the annotations allow examples, and that was the plan.

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+        "publisher-place": {
+          "type": "string"
+        },
+        "references": {
+          "type": "string"
+        },
+        "reviewed": {
+          "title": "Reviewed - Related Item",
+          "description": "In a review, defines metadata for the reviewed item.",
+          "$ref": "#/definitions/bibitem"
```
bwiernik, line 401:
```suggestion
          "$ref": "#/definitions/related-bibitem"
```

This and `original` shoudl be `related-bibitem`, no?

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+          "type": ["string", "number"]
+        },
+        "title": {
+          "type": "string"
+        },
+        "title-short": {
+          "type": "string"
+        },
+        "translated-title": {
+          "description": "[Deprecated - Q: what TODO?]",
```
bwiernik, line 429:
```suggestion
```

This one `translated-*` variable will suffice for the vast vast majority of cases. I don't think standard CSL needs to include `translated` as a related element (though this structure provides a great hook for making the multilingual parts of CSLm/Jurism a separate module).

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+        "number-of-pages": {
+          "type": ["string", "number"]
+        },
+        "number-of-volumes": {
+          "type": ["string", "number"]
+        },
+        "original": {
+          "title": "Original - Related Item",
+          "description": "Used to define metadata for an original publication.",
+          "$ref": "#/definitions/bibitem"
```
bwiernik:
```suggestion
          "$ref": "#/definitions/related-bibitem"
```

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+        },
+        "original-publisher": {
+          "description": "[Deprecated - Use `publisher` on the `original` related item instead]",
```
bwiernik, line 357:
```suggestion
          "description": "Can also be be entered as `publisher` on the `original` related item",
```

I think we could leave these `original-*` variables available for direct entry into item data.

schemas/input/csl-data.json
bdarcus, line 354:
```suggestion
          "$ref": "#/definitions/related-bibitem"
```

schemas/input/csl-data.json
bwiernik:
```suggestion
          "description": "Can also be entered as `author` on the `original` related item",
```

I think we could leave existing `original-*` variables available for direct entry into item data. `original-author` specifically might not necessarily need to be accompanied by additional original information (e.g., if it refers to the historical author of a work, the original songwriter of a musical piece, or the original developer of a software).

cf. the documentation
> The original creator of a work (e.g. the form of the author name listed on the original version of a book; the historical author of a work; the original songwriter or performer for a musical piece; the original developer or programmer for a piece of software; the original author of an adapted work such as a book adapted into a screenplay)

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+          "$ref": "#/definitions/date-variable"
+        },
+        "event-date": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "issued": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "original-date": {
+          "description": "[Deprecated - Use a date property on the `original` related item instead]",
```
bwiernik, line 224:
```suggestion
          "description": "Can also be entered as `issued` on the `original` related item",
```

schemas/input/csl-data.json
bwiernik: I don't think `archive` needs to fit into this related model--it's analogous in structure to `publisher`, rather than to `reviewed`

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+          "type": "string"
+        },
+        "DOI": {
+          "type": "string"
+        },
+        "edition": {
+          "type": ["string", "number"]
+        },
+        "event": {
+          "description": "[Deprecated - Q: what TODO?]",
```
bwiernik, line 290:
```suggestion
```

I don't think `event` needs to fit into this related model--it's analogous in structure to `publisher`, rather than to `reviewed`

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+          "type": "string"
+        },
+        "event-place": {
+          "description": "[Deprecated - Q: what TODO?]",
```
bwiernik, line 294:
```suggestion
```

I don't think `event` needs to fit into this related model--it's analogous in structure to `publisher`, rather than to `reviewed`

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+          "type": "string"
+        },
+        "ISSN": {
+          "type": "string"
+        },
+        "ISWC": {
+          "type": "string"
+        },
+        "issue": {
+          "type": ["string", "number"]
```
bwiernik, line 325: General comment--perhaps we should make the structure of the JSON and RNC schemas more similar and make definitions for "standard" and "number" and "title" variables, each variable can refer to, rather than have number variables just include both "string" and "number"? This would be a good opportunity since it's a big diff anyway.

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+        },
+        "title-short": {
+          "type": "string"
+        },
+        "translated-title": {
+          "description": "[Deprecated - Q: what TODO?]",
+          "type": "string"
+        },
+        "translated-title-short": {
+          "description": "[Deprecated - Q: what TODO?]",
```
bwiernik, line 433:
```suggestion
```

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
-      "translator": {
-        "type": "array",
-        "items": {
-          "$ref": "#/definitions/name-variable"
+      "required": ["type", "id"],
+      "additionalProperties": false
+    },
+    "related-bibitem": {
+      "title": "Related bib item",
+      "description": "Used to define related bibliographic data; value can either be a refid string, or an embedded item object",
```
bwiernik, line 468:
```suggestion
      "description": "Used to define related bibliographic data; value can either be an item id or citekey string, or an embedded item object",
```

Do we ever use the term `refid` elsewhere?

schemas/input/csl-data.json
bwiernik: Does the suggested edit work? Otherwise, I'm not sure what you are referring to with respect to "citekey confusion".

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+        "citation-label": {
+          "type": "string"
+        },
+        "collection-number": {
+          "type": ["string", "number"]
+        },
+        "collection-title": {
+          "type": "string"
+        },
+        "container-title": {
```
bwiernik, line 271: With respect to `contianer-` variables, the principle reason to connect it with the `@related` structure is to enable a bibliography entry like this:

Jones, A. A. (2010). This is a book chapter. In Wilson (2010), pp. 100–150.
Smith, B. B. (2010). This is another chapter in the same book. In Wilson (2020), pp. 180–210.
Wilson, E. E. (Ed.). (2020). _This is the book holding the chapters._ Publisher. https://doi.org/10.1234/56789

Given this fairly rare need for a hierarchical structure, I'm inclined to do this:

```
        "container": {
          "title": "Container holding the item",
          "description": "Used to link an item to its container item.",
          "$ref": "#/definitions/related-id"
        },
```

That is, `container` only contains an `id` or `citekey`, not a full `bibitem`. The `container-title`, etc. for the item are entered as flat variables.

bwiernik: The other situation that has been raised is for things like distinguishing if a `translator` is the translator of the item (e.g., a specific chapter) or of the whole book. I'm skeptical that this is a distinction we should try to make. Bibliographic metadata seem to rarely make this distinction (the translator of either the chapter or whole book is just listed as "translator"), so I don't think that styles can rely on a `translator` variable to mean the item translator and `container-translator` to mean the translator of the whole book. Same with editor. We would almost always want to assume that editor/translator refers to the role for the whole container.

A potentially more reliable translator heuristic might be: "If the there is no editor or if editor and translator are the same [i.e., `if variable="editor-translator"`], then `translator` is the translator of the whole container; if there is an editor who is different from the translator, then `translator` is the translator of the specific item." But even that seems too unreliable.

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+        },
+        "archive_collection": {
+          "type": "string"
+        },
+        "archive_location": {
+          "description": "[Deprecated - Q: what TODO?]",
+          "type": "string"
+        },
+        "archive-place": {
+          "description": "[Deprecated - Q: what TODO?]",
```
bwiernik, line 247:
```suggestion
```

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+          "title": "Original - Related Item",
+          "description": "Used to define metadata for an original publication.",
+          "$ref": "#/definitions/bibitem"
+        },
+        "original-publisher": {
+          "description": "[Deprecated - Use `publisher` on the `original` related item instead]",
+          "type": "string"
+        },
+        "original-publisher-place": {
+          "description": "[Deprecated - Use `publisher-place` on the `original` related item instead]",
```
bwiernik, line 361:
```suggestion
          "description": "Can also be be entered as `publisher-place` on the `original` related item",
```

I think we could leave these `original-*` variables available for direct entry into item data.

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+          "type": "string"
+        },
+        "original-title": {
+          "description": "[Deprecated - Use `title` on the `original` related item instead]",
```
bwiernik, line 365:
```suggestion
          "description": "Can also be be entered as `title` on the `original` related item",
```

I think we could leave these `original-*` variables available for direct entry into item data.

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+        "references": {
+          "type": "string"
+        },
+        "reviewed": {
+          "title": "Reviewed - Related Item",
+          "description": "In a review, defines metadata for the reviewed item.",
+          "$ref": "#/definitions/bibitem"
+        },
+        "reviewed-title": {
+          "description": "[Deprecated - Use `title` on the `reviewed` related item instead]",
```
bwiernik, line 404:
```suggestion
          "description": "Can also be be entered as `title` on the `reviewed` related item",
```

I think we could leave these `reviewed-*` variables available for direct entry into item data.

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+        "volume": {
+          "type": ["string", "number"]
+        },
+        "year-suffix": {
+          "type": "string"
+        },
+        "custom": {
+          "title": "Custom key-value pairs.",
+          "type": "object",
+          "description": "Used to store additional information that does not have a designated CSL JSON field. The note field can also store additional information, but custom is preferred for storing key-value pairs.",
```
bwiernik, line 451:
```suggestion
          "description": "Used to store additional information that does not have a designated CSL JSON field. The custom field is preferred over the note field for storing custom data, particularly for storing key-value pairs, as the note field is used for user annotations in annotated bibliography styles.",
```

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
-          "$ref": "#/definitions/name-variable"
+      "required": ["type", "id"],
+      "additionalProperties": false
+    },
+    "related-bibitem": {
+      "title": "Related bib item",
+      "description": "Used to define related bibliographic data; value can either be a refid string, or an embedded item object",
+      "anyOf": [
+        {
+          "type": "string"
```
bwiernik, line 471:
```suggestion
          "$ref": "#/definitions/bibitem-id"
```

`bibitem-id` just defined as a string, but this could make the schema more human-readable to indicate what type of string this is intended to be?

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+          }
+        },
+        "recipient": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/name-variable"
+          }
+        },
+        "reviewed-author": {
+          "description": "[Deprecated - Use `author` on the `reviewed` related item instead]",
```
bwiernik, line 196:
```suggestion
          "description": "Can also be be entered as `author` on the `reviewed` related item",
```

I think we could leave these `reviewed-*` variables available for direct entry into item data.

bdarcus:
This is a related change I think we need to seriously consider to the input and style models for 1.1. 

Feedback please?

PS - I'm not sure who's still at Mendeley.

@robertknight @jgm @fbennett @adam3smith @rmzelle @dstillman @dhimmel @cormacrelf @retorquere @larsgw @njbart

schemas/input/csl-data.json
```diff
@@ -4,425 +4,477 @@
...
+            "software",
+            "song",
+            "speech",
+            "standard",
+            "thesis",
+            "treaty",
+            "webpage"
+          ]
+        },
+        "id": {
```
bdarcus, line 62: Technical issue: `id` is required, but we don't need it for the embedded form.

bwiernik:
So, as an idea for a graphical interface: In Zotero for example, a `reviewed` item could either be specified using Zotero's Related Items feature, with a directional "reviewed" type applied to the relation. When Zotero generates CSL-JSON for the item, it would include the id and data for the related item in the data. Alternatively, the interface could have a "Add Reviewed Item" button or similar that would add the original item fields to the current item's interface.

schemas/input/csl-data.json
bwiernik: Hmm, should we make a `standard-bibitem` definition that is `bibitem` plus required `id`, whereas `related-bibitem` is either `id` _or_ `bibitem`?

bdarcus:
Oh yes, that should be the (in retrospect) obvious solution.

I'll give it a try.

schemas/input/csl-data.json
dhimmel: nice change. we should consider implementing it separately from this PR, since it's unrelated.

schemas/input/csl-data.json
dhimmel: Hmm I think we should take the opportunity to simplify the schema and remove redundant fields.

bwiernik: My major concerns are always:
  1. Backwards compatibility with existing CSL-JSON data in the wild (especially embedded in word processor documents) (e.g., why `journalAbbreviation` and `shortTitle` stick around)
  2. Implementation ease in applications; I don't want breakage around the existing `reviewed-author` structure in Zotero, for example, to get in the way of it adopting CSL 1.1 entirely.
  3. Ease of use for entering flat key-value pairs when users are entering information manually (e.g., in Mendeley's Note field).

bdarcus:
But keep in mind I didn't remove these; I marked them deprecated, where the
audience is mostly developers.

We could remove them in 1.2 or 1.3.

BTW, we need to add a note to the spec that we don't follow the semantic
versioning spec, for good reason (in our case).

schemas/input/csl-data.json
bdarcus: Should be fixed now. Thanks for the tip.

schemas/input/csl-data.json
bdarcus: > With respect to `contianer-` variables, the principle reason to connect it with the `@related` structure is to enable a bibliography entry like this:

If we had started with this structured, "container" and "collection" would likely be relations.

Consider the example of a chapter in a book, where you cite both. For that case, there's value in allowing the full relation.

It's worth discussing, but I did not include it, because it would make this more complicated.

schemas/input/csl-data.json
bdarcus: But keep in mind I didn't remove these; I marked them deprecated, where the audience is mostly developers.

We could hold off on removing to 1.2. or 1.3; even later if needed be.

But saying nothing will enshrine the status quo for a long time.

fbennett:
> [A] `reviewed` item could either be specified using Zotero's Related Items feature, with a directional "reviewed" type applied to the relation.

I think you would get more flexibility and keep things simpler on the calling-application side by leaving relations as a simple bidirectional reference, and making a controlled list of subtypes available on individual items. Then you could produce any of four renderings at the author's discretion:
* <original-work>
* <review>
* <original-work> *reviewed by* <review>; or
* <review> *reviewing* <original-work>

In the two-item forms, you could optionally use parallel field suppression to filter out things that needn't appear in the subsequent item. If that would work, the only additional work for the calling application would be to flag a simple relation between the items.

bwiernik:
@fbennett I’m not following. I’m confused by your notation. 

Let’s use `base-item` as the label for the main item in the data. 

I need to be able to specify:
`base-item` _reviews_ `reviewed-item`
`base-item` _was originally_ `original-item`

Some possible extensions:
`base-item` _is contained by_ `container-item`
`base-item` _affirms_ `affirming-item`
`base-item` _retracted by_ `retraction-item`

That’s going to require a directional relationship necessarily.

larsgw:
Regarding `archive`, `event` and `publisher`: I think they can be made related, but the related item won't be a bibitem which complicates the schema a bit. Large bibliographies with common publishers could be a lot nicer that way.

Regarding actual bibitems: I don't really see why not to make every bibitem-like property related is a problem. I think the main problem is that in the linked format, there needs to be some way to specify which items are rendered. Then we could e.g. make a `original` bibitem with just an author (as @bwiernik mentioned) without that being any sort of problem (as far as I can tell anyway).

In both cases, I would personally like to go all-in on linked data, with the note that not every item is an item in the bibliography.

---

I would think the flat format (in styles) could still be resolved against linked-data input, though variable names with a form at the end make that a bit more difficult. Right now, the suggestion seems to be to do this the other way around?

---

Can linked and embedded formats be mixed?

---

Regarding `custom` and `edtf-datatype` — that seems a bit unrelated to related. I don't know if we need a separate PR for everything, but I would like to discuss EDTF a bit (if there hasn't been a discussion already).

fbennett:
@bwiernik I'm pretty sure that what I suggested will handle those cases. Maybe it will help to work backward from expected output. If you can give an rendered example of each of the five cases, I can check my intuition.

bdarcus:
Lars - I merged EDTF support yesterday. It's an option alongside the
current date model.

But the documentation for that is not yet written, so if you want to help
with that ...

Also, there's an issue here that proposes to move publisher to names.

bwiernik:
@fbennett I just have no idea what you were trying to say/illustrate. Could you rewrite your comment using `base-item` to avoid conflating original and regular items and explain your logic?

bwiernik:
@larsgw The primary audiences I think we need to bear in mind in terms of accessibility of the data structure I think are:

1. Applications like Zotero or a Mendeley that need to construct CSL data, presumably based on their user-facing data structure.
2. Users manually curating their data in CSL YAML-type data. 

I really don’t see a lot of users or applications being interested in, for example, cross-linking to `publisher` or `archive` or `event` items. For example, a publisher’s listed name or location often varies across items (SAGE vs Sage Publications)—what should be cited is the form listed on the item. Similarly for archives or events, for most users, only a few items are going to share an event or archive at most. 

I think we should be cautious about mandating a much more complex data structure into applications and users.

fbennett:
@bwiernik We're speaking at cross purposes because we're adopting different assumptions about how the processor input should be structured. You are seeing the problem as one of making the semantic relation between items (`base-item` and the other) explicit in the input. I'm suggesting that you might be able to achieve the desired output with atomic items that carry hints resolved at render-time, in which case the concept of `base-item` would not apply. I can clarify with a sample of desired output.

bwiernik:
@fbennett Styles are still defined around a base item. The “original” being described here is not related to the “reviewed” here at all. They are different types of relations (to a base item) entirely. 

So for example, this is a base item-reviewed item relation:
Jones. “Title of review—the base item” [Review of “Title of reviewed item” by Smith]

This is a base item-original item relation:
Adams. “Title of base item” [Originally published as “Title of original item”]

schemas/input/csl-data.json
bwiernik: I appreciate that, but the flat data model allows a lot of valuable flexibility. For example, bibitem requires a type field. What is the type for `collection`—book series, lecture series, etc.? We don’t currently have types for these and probably don’t need them except as types for this internal hierarchical node. For containers, currently container-title can be used for all manner do containers (presentation sessions, books, journals)—constraining them to be specific enumerated types is less valuable than regarding the container as an container that need to be included on the cited item metadata—something that is specific to the item.

fbennett:
@bwiernik Right, I understand the relationships to be expressed. How would those citation samples look in their full form (with date, publisher, etc.) for both partners?

bwiernik:
Jones. “Title of review—the base item” [Review of “Title of reviewed item” by Smith]

Smith. “Title of reviewed item”

This is a base item-original item relation:
Adams. “Title of base item” [Originally published as “Title of original item”]

Adams. “Title of original item”

The reviewed or original items don’t need to refer to the base item at all.

bdarcus:
@fbennett  - just highlighting a couple of places where I'm not following.

> I think you would get more flexibility and keep things simpler on the calling-application side by leaving relations as a simple bidirectional reference, and making a controlled list of subtypes available on individual items. 

First, what's the object of the above statement? Is this the client app (Zotero, for example), or are you talking about this PR. If the latter, input data, style, or both?

E.g. where is that "list of subtypes" defined?

...

> In the two-item forms, you could optionally use parallel field suppression ...

Second, you mean some additional spec logic that would make sure, say, one didn't print the publisher name twice?

bwiernik:
@fbennett perhaps let’s take an (absurdly) more complex case to illustrate the difference between base and original:

Jones. “Title of book review” [Review of “Title of reviewed book” by Smith] (Originally published as “Original title of book review”)

There are three items: the base item, the reviewed item, and the original item.

fbennett:
> First, what's the object of the above statement?

Simply put, I'm concerned that added complexity of input will prove a fatal barrier to implementation. Processors are already insanely complicated inside, demanding years of effort to cover the CSL spec in full. Adding the burden of navigating nested inputs on top of that is a pretty deep papercut.

If the aim is to produce a particular pattern of output for related items, I have a notion that it can be achieved with minor extensions to the content of atomic items, and with relatively minor adjustments in processor code; but without complete citation examples to work back from, all I can offer is the intuition.

bdarcus:
Ok, got it.

Can someone provide @fbennett with some complete examples?

fbennett:
@bwiernik 
> @fbennett perhaps let’s take an (absurdly) more complex case to illustrate the difference between base and original:

This pales in comparison with requirements in the legal domain. For a set of worked examples, please take a look at [R14. History of the Case](https://juris-m.github.io/indigobook/#R14) in the Jurism version of Indigo Book. If you click on the citation for *Energy & Env’t Legal Inst. v. Epel*, the popup shows the three atomic items of which it is composed. The only item data not illustrated in the view is an array on each item that flags the relations among them. The terms *aff'd* and *cert. denied* are set as simple prefixes on their respective items.

dstillman:
I've only skimmed this, so apologies if I'm misunderstanding something.

On a technical level, we could fairly easily implement directional relations like @bwiernik describes in Zotero. (While related items in Zotero are always bidirectional, we do use directional relations for other things.) But I have other concerns:

1) This just isn't how data comes into Zotero. People go to a website and save a book review. They might expect to tweak some fields, but they don't expect to somehow separate out a reviewed book into a separate item from the saved data. And automatically building two items from a book review page (at least when not served as CSL-JSON) would likely be nearly impossible.

2) Dependencies in data models tend to make everything more complicated. What happens if the reviewed book is deleted? Does it need to warn you that there's a dependency? What if there's a sync error and the reviewed book isn't yet downloaded to a computer? If you drag the book review to another library, does it need to automatically transfer the reviewed book to the other library? Whenever Zotero displays the title of the book review, everywhere that it does that, does it, too, need to format it using the reviewed book? If you export a book review in RIS, does it try to build a flat single item that would import into another program, or does it unexpectedly export two items? Et cetera.

As I said in a related thread where we discussed the advantages of the flat key-value model, I'm just not sure these problems are worth avoiding a bit of extra verbosity.

dstillman:
(Though this may be unnecessarily conflating input model and data model. Practically speaking, there's no real difference between a nested input model and lots of `reviewed-` variables. A calling application could still have flat items and generate a nested input model if that was what the processor required. But a nested data model creates lots of new problems.)

larsgw:
> Practically speaking, there's no real difference between a nested input model and lots of reviewed- variables. A calling application could still have flat items and generate a nested input model if that was what the processor required.

That's what I wanted to get to :). I get how the flat model is beneficial (or essential) in a lot of use cases, but I also like the linked data a lot. However, I think the current flat model is pretty well-suited for conversion between the two. I made a small demo (it isn't proper JSON-LD yet, nor does it handle nested linking and authors): https://jsfiddle.net/LarsGW/zt96gxvu/.

fbennett:
@larsgw In the jsfiddle, I see two items in the terminal in the lower-right corner. If I click on **run**, the console shows code running without error, and the terminal content refreshed (I think), but the content remains the same. Is it meant to convert to a linked format? Should I be clicking somewhere else?

bwiernik:
Thanks @dstillman

I agree that I am generally in favor of not mandating related item ids. 

The only thing being proposed that would really required separate items in a database that related would be multiple children sharing a common container (multiple chapters in a book, which might be cited all with reference to the shared book)—even that could be handled with some headaches via text matching across `chapter` items. 

The proposal to for reviewed, original, retracted, etc., would permit all of this information to be supplied as children of the base item. 

I think that the proposal to enter reviewed, original, retracted, etc., which represent _entirely distinct items_, in arrays is useful to avoid creating many variants of variables `reviewed-title`, `reviewed-container-title`, etc. That is useful from a style design perspective.

But I don’t see particular value and a lot of drawbacks to imposing a hierarchical data structure on all item components.

bwiernik:
@fbennett It’s the end of my day, so apologies if I’m not the clearest. 

Does this illustration make more sense to illustrate the relationships in both directions (I think we might be using “directional” differently):

Jones (2020). I hated this book. _Journal of Book Reviews_. [Review of “A book” by Smith, 2019, Smith’s Publisher]

Smith (2019). A book. Smith’s Publisher. 

Entered into data as:

Jones _reviews_ Smith
Smith _is reviewed by_ Jones 

But from the perspective of CSL for this relationship, the second relationship doesn’t matter.

denismaier:
> But from the perspective of CSL for this relationship, the second relationship doesn’t matter.

Except if you change the relationship from "reviewed" to "original".

Then, you'll have

Jones (2020). A reprinted book . Jones's new publisher [Reprint of “An original book” by Jones, 2019, Jones's old publisher].
Jones (2019). A original book. Jones's old publisher. [Reprinted as "A reprinted book", by Jones, 2020, Jones's new publisher].

Entered into data as:
Jones 2019 reprinted of Jones 2020 = Jones 2020 is reprint of Jones 2019

Depending on your needs and disciplinary customs, you might want to treat one or the other as `base-item`.

denismaier:
> > With respect to the deprecation questions, generally, I think it would be good to maintain the existing variables as "first-class" variables that can be entered directly.
> 
> Even ones we added in the past week?

Given that these variables aren't yet used much/at all in the wild yet, I'd say we can remove them if we can come up with a better and more flexible solution. After all, 1.0.2 has not even been released yet.

Concerning biblatex: First, they keep deprecated fields/features available some time, often as aliases to the new recommended way. But they have a mechanism where users and style authors be warned when deprecated fields are used (p. 179 in the biblatex manual). The can even map a deprecated field to a new field.

denismaier:
> @larsgw In the jsfiddle, I see two items in the terminal in the lower-right corner. If I click on **run**, the console shows code running without error, and the terminal content refreshes (I think), but the content remains the same. Is it meant to convert to a linked format? Should I be clicking somewhere else?

@fbennett In the lower right corner, there's a "convert" button. Clicking this converts the CSL-JSON to CSL-JSON-LD.

denismaier:
> As I said in a related thread where we discussed the advantages of the flat key-value model, I'm just not sure these problems are worth avoiding a bit of extra verbosity.

This leads to the question of what exactly is "a bit of extra verbosity".  Is multiplying the existing number of variables by 3 just a bit?

fbennett:
@denismaier Okay, thanks. Maybe it's only the view supplied to this region, but the Convert button was overlaid by a jsfiddle-supplied button for Console. Setting that to `display:none:` in the browser's Inspector revealed the button underneath, and it works. I can see how this linked structure would be useful on the application side, to share a journal name or publisher among multiple records. In a citation processor, it looks to me like it just adds complexity. My temptation on receiving records in this form would be to resolve the field links, flatten the records, and just deal with everything as key:val pairs as we do now. Am I missing something?

fbennett:
@bwiernik Thanks, it's starting to come into focus. To extend one of the examples, I think what you have in mind is this?

**Citation**
> (Jones 2020)

**Bibliography**
> Jones (2020). I hated this book. Journal of Book Reviews. [Review of “A book” by Smith, 2019, Smith’s Publisher]

So the author cites a single item, the review by Jones, and that appears in citations as a single item, but in the bibliography it pulls in its related partner the book by Smith. I'll pause there for confirmation, to be sure I understand correctly.

denismaier:
> So the author cites a single item, the review by Jones, and that appears in citations as a single item, but in the bibliography it pulls in its related partner, the book by Smith. I'll pause there for confirmation, to be sure I understand correctly.

Not necessarily in the bibliography, of course. It can also pull in related information in a footnote. That should just depend on the citation style.

larsgw:
> Maybe it's only the view supplied to this region, but the Convert button was overlaid by a jsfiddle-supplied button for Console.

Oh sorry, I wasn't sure where I could host it quickly and I thought this would work.

> I can see how this linked structure would be useful on the application side, to share a journal name or publisher among multiple records. In a citation processor, it looks to me like it just adds complexity.

Yes, I feel like we are working to three different goals right now, and they don't really combine.

  1. A format that is easy to write and read for humans (YAML, EDTF)
  2. A format that can represent information necessary for citation rendering without too much complication (flat format but with 'chained' properties, as @fbennett suggest)
  3. A format that can be mapped to other formats easily (linked formats means less individual fields)

Combining those goals into a single format adds complexity, I feel. Checking for and parsing both EDTF and the old `date-parts`, writing `related="reviewed-id" variable="title"` instead of `variable="reviewed"` all adds to the list of edge cases. Instead of accommodating the increased use of CSL-JSON, this requires every library and application dealing with CSL-JSON to support everything, no matter which goal they have themselves.

I (carefully) suggest making two formats, one flat format with chained properties and EDTF used in citeprocs, and a linked format using the CSL vocabulary, without EDTF but with an improved date format. Converting between the two is still possible (and possibly quite easy), but it hopefully separates the goal of rendering citations from using the CSL vocabulary to represent bibliographical data for other purposes. That's my perspective, I'm open to feedback.

bdarcus:
> Combining those goals into a single format adds complexity, I feel. 

I think your analysis is correct. I was thinking the same thing yesterday.

> I (carefully) suggest making two formats, one flat format with chained properties and EDTF used in citeprocs, and a linked format using the CSL vocabulary, without EDTF but with an improved date format. Converting between the two is still possible (and possibly quite easy), but it hopefully separates the goal of rendering citations from using the CSL vocabulary to represent bibliographical data for other purposes. That's my perspective, I'm open to feedback.

So two json schemas is indeed an option, I think, worth considering: one aimed at humans and yaml (as you say, with EDTF), and the other aimed at machines and json (as you say, with an enhanced date model to match what we adopt of EDTF).

But if that was the path, you'd want them to be losslessly convertible of course, hence need to update the data model.

While there are pros and cons to this path, I'm totally open to this.

But, it is partially orthogonal to this PR, unless we want to say the human schema has the relations, and the machine schema does not?

Also, just FYI everyone, I added this and this other question to the top: do we want both embedded and linked input forms, or only embedded?

fbennett:
Sticking with the bibliography example, if the relation is expressed by embedding the "child" item in the explicitly cited item (or by setting a directional relation between them), that's coding an implicit assumption about the appearance of the bibliography into the input data. I'm not sure that's a good idea. What should be the effect if the author cites the "child" item explicitly? If a style guide requires items to be listed separately, would the "child" item still be pulled in as an uncited item? When switching from such a style to one that requires impliciting combination of elements, would the processor's internal data structures need to be transformed? Another issue is sequencing. If there is an item in Chinese, and a translation, and a review in English, should all three be pulled in automatically and listed in a single bibliography entry? If so, will their sequence be fixed by the calling application, or by the style? 

Coding to combine elements in the bibliography will raise some hard problems. They can be resolved, but I'm not sure that a hierarchical data structure in the input makes their resolution easier.

bdarcus:
There are a lot of issues here, but worth remembering the origin of this, which @denismaier hinted at with his question about whether it would be acceptable to triple the amount of flat variables in CSL (which I don't think it is): #167 covers it, I think. 

When I realized how many new variables would be added with keeping the exclusively key-value structure (beyond names and dates), I thought it was time to have this discussion. 

But  to @fbennett's latest point ...

> Coding to combine elements in the bibliography will raise some hard problems. They can be resolved, but I'm not sure that a hierarchical data structure in the input makes their resolution easier.

My hunch is the flat vs related structure actually doesn't make much difference on the formatting side; that what's hard is the more complicated citations.

Of course, maybe we conclude we don't want to add any hierarchy on the flat values, and we also don't want to support these feature requests now. I do think that's a reasonable conclusion. 

But I do not want to triple the amount of new variables, per @denismaier's question.

bwiernik:
Perhaps let’s set the question of relations-by-item-id aside for now. Because the proposal also allows for a full bibitem to be listed under reviewed/original, Zotero could choose to just not support or use the id-based structure and just use the full nested bibitem structure. 

I think there is a lot of value in being able to do full proper citations of reviews, republished works, retracted items, etc. Hierarchically structuring these data seems like a good way to avoid an tripling the number of enumerated variables.

larsgw:
> Also, just FYI everyone, I added this and this other question to the top: do we want both embedded and linked input forms, or only embedded?

If either is compatible with JSON-LD, there's a way to translate between the two.

bdarcus:
> If either is compatible with JSON-LD, there's a way to translate between the two.

I don't think JSON-LD is an explicit goal for us. But if it's trivial to add (say, adding a `@type` property, or renaming our current `type`?), we might as well consider it.

fbennett:
I've not been following the full discussion in detail, but on the implementation side, the simplest way to explore changes at the various version levels will be test fixtures. If tests for particular changes or extensions can be added to the test suite, or a branch of the test suite and flagged in some way, I can take a look at how far `citeproc-js` can cover them.

bdarcus:
> ... the simplest way to explore changes at the various version levels will be test fixtures. If tests for particular changes or extensions can be added to the test suite, or a branch of the test suite, and flagged in some way, I can take a look at how far `citeproc-js` can cover them.

That's a _great_ that should have been obvious in retrospect!

I just added a new "experiements" [subdir](https://github.com/citation-style-language/test-suite/tree/master/processor-tests/humans/experiments) to the master repo. 

Am not sure if that was the best approach (I wasn't sure on how to manage the branch approach in this case), but feel free to change it if not.

I put two placeholder tests there related to 1.1, and gave each a version number tag of `1.1-experimental`.

Could someone please help me by filling out these two tests (and add any other tests relevant to 1.1), and link the relevant one(s) here? [Here's](https://github.com/citation-style-language/test-suite/blob/master/processor-tests/humans/experiments/related_att.txt) the start of the one for `@related`.

bdarcus:
Actually, I also added a `v1.1-experiements` branch. We can add there.

bdarcus:
I just wanted to call attention to @fbennett comment over [here](https://github.com/citation-style-language/test-suite/commit/e3673e5f16cbb98f87ac7ce3fcd512453e5bd3b2#r40425197).

In it, he basically argues, as I understand it, for:

1. don't allow the embedded object variant
2. don't worry about explicit typed related; just have a generic "seeAlso" or equivalent array.

So the result would be:

``` json
[
  {
    "id": "ITEM-1",
    "type": "article-journal",
    "subtype": "review",
    "title": "A review",
    "seeAlso": [ "ITEM-2" ]
  },
  {
    "id": "ITEM-2",
    "subtype": "book",
    "title": "A reviewed title",
    "seeAlso": [ "ITEM-1" ]
  }
]
```

As I asked there: how would a processor know the cited item ("ITEM-1") is a review? Just by virtue of the metadata on that item?

Seems like that would work, and have the virtue of simplicity.

dhimmel:
> don't allow the embedded object variant

I think only adding the id-reference-syntax is a good idea. It's simpler and satisfies 90% of use cases. Adding the embedded object syntax could always be done at a later time.

> don't worry about explicit typed relations on input; just have a generic "seeAlso" or equivalent array.

I think having a way to indicate the type of the relationship is essential. But it may be nice to have a syntax that supports arbitrary relationships, like:

```yaml
related:
- reviews: id-of-reviewed-paper
- translates: id-of-original-language-version
```

bdarcus:
> It might be that you adopt this format in the future, since it would be backwords compataible with `reviews: id` as a top-level object.

I'm not following you here @dhimmel.

dhimmel:
> I'm not following you here

```
# this could be supported immediately.. only works for predefined relationship types
reviews: id

# arbitrary relationship types and multiple relations of the same type could be supported in the future with:
relations:
 - type: reviews
   id: id-of-reviewed-paper
 - type: translates
   id: id-of-original-language-version
```

bwiernik:
I think @fbennett's comments are a bit of a non-sequitor. That proposal works fine for legal citation, but requiring a user to manually cite a book review and then the reviewed book separately in sequence or reprinted item and its original version separately in sequence just doesn't make sense to me at all. 

My comment from that thread:

The major impetus I had in mind was for citing things like this:
> Bronfenbrenner, U. (2005). The social ecology of human development: A retrospective conclusion. In U. Bronfenbrenner (Ed.), _Making human beings human: Bioecological perspectives on human development_ (pp. 27–40). SAGE Publications. (Reprinted from _Brain and intelligence: The ecology of child development_, pp. 113–123, by F. Richardson, Ed., 1973, National Educational Press)

or this:
> Mirabito, L. A., & Heck, N. C. (2016). Bringing LGBTQ youth theater into the spotlight [Review of the film The year we thought about love, by E. Brodsky, Dir.]. Psychology of Sexual Orientation and Gender Diversity, 3(4), 499–500. https://doi.org/10.1037/sgd0000205

or this:
> Ricaurte, G. A. (2002). Severe dopaminergic neurotoxicity in primates after a common recreational dose regimen of MDMA (“ecstasy”). Science, 297(5590), 2260–2263. https://doi.org/10.1126/science.1074501 (Retraction published September 12, 2003, Science, 301[5639], 1454, https://doi.org/10.1126/science.301.5639.1479b)

In these items, there is important information about the original item, reviewed item, or retraction notice that needs to appear _every time the base item is cited_. It is not the same situation as a legal `seeAlso` where the relevant references are more a part of the author's argument and so they might cite various combinations of them. The reviewed, original, retracted information _is a core part of the cited item_. 

The purpose of this proposal was to allow for complete references of reviews, reprinted items, retracted items, etc. without exploding the number of variables by adding enumerated `reviewed-*`, `original-*`, `retracted-*` versions of every variable.

fbennett:
(Response on the [original thread](https://github.com/citation-style-language/test-suite/commit/e3673e5f16cbb98f87ac7ce3fcd512453e5bd3b2#r40528352))

bwiernik:
The question comes down to how are these data going to be constructed. @dstillman would it be feasible for Zotero to generate CSL-JSON like this from one Zotero item (so not relying on relations between separate items in a Zotero library):

```
id: book-review-item-id
  author:
    given: John
    family: Smith
  title: Review title
  relations:
    reviewed: book-item-id

id: book-item-id
  author: 
    given: Susan
    family: Jones
  title: Book title
```
(I've obviously written it as YAML for readability.)

bdarcus:
I simplified this PR to a) only focus on the input side (I'll add another one for styles once we settle this), and b) only offer the "linked" option.

I partly redid this because this branch was out of sync with 1.1.

bwiernik:
I’d suggest to postpone discussion of deprecating the existing variables until we settle the discussion about input model.

bdarcus:
I think that's reasonable tack, except we can't really document this without some decision on the plan long term. Otherwise, it will be confusing for users and developers.

dstillman:
> @dstillman would it be feasible for Zotero to generate CSL-JSON like this from one Zotero item (so not relying on relations between separate items in a Zotero library)
>
> ```
> id: book-review-item-id
>  author:
>    given: John
>    family: Smith
>  title: Review title
>  relations:
>    reviewed: book-item-id
>
> id: book-item-id
>  author: 
>    given: Susan
>    family: Jones
>  title: Book title
> ```

If the data's all there in a single item, we can generate it in any form, though we would need to generate a random id for the reviewed book, since that wouldn't exist in the database. A second review of the same book would generate a duplicate, identical reviewed book in the JSON.

But it would create additional complexity for anything accepting the input that doesn't use the same relations-based data structure. That includes imports back into Zotero, which would become more complicated, because it would require either processing the entire file and building a map of ids before importing or deferring entries with relations until the linked items were found.

Obviously, if you have a tool designed around elegant hierarchical relations, this would be easier, but given that that's just not the format that metadata exists in the real world — where you export metadata for a book review from a website — that's a pretty narrow constituency.

If this will solve real problems, OK, but I would just suggest keeping the history of bibliographic file formats in mind when evaluating the trade-offs here. There's a reason people use RIS and not RDF.

bwiernik:
@bdarcus BibLaTeX doesn't much support reviews much (the standard styles treat `@review` as an alias for `@article`). Some styles use `titleaddon` or define new "reviewed" versions of fields: https://tex.stackexchange.com/questions/114207/book-reviews-entry-in-biblatex-chicago So, BibLaTeX has two systems: a system of extended item types and also a Related system.

From a date curation point of view, I really strongly prefer the embedded object structure. That makes it much easier for CSL-JSON or CSL YAML to be transportable and shareable because all of the information for an item is stored in one place

It also just much more well-aligns with how bibliographic data exist in the wild for things like retractions and reviews. If we _additionally_ support an id-based structure, that would (1) be convenient users curating YAML by hand and (2) be useful for applications more like what @fbennett describes. I might be missing something obvious, but I really don't see how looking up how `related -> id -> look up data for item by id` is much different or easier than `related -> extract embedded data below`.

bdarcus:
I'm agnostic on the embedded, and see no problem including it.

I'm not sure where others are on the question.

Do you see your concern about this as connected to your concern about
deprecating?

As in, if we have the former are you less concerned about the latter?

dstillman:
There's not much value in adding embedded in order to avoid the complications of linked if you're also going to add linked. If it's specified, tools that claim to take CSL-JSON as input would presumably be expected to handle both.

If CSL-JSON is intended to be used solely as a data format for citation processors, then it's up to @fbennett and a handful of other people in the world whether a linked data model is worth it. If you want other tools to support it as an input format, making it significantly harder to process is not going to help.

bwiernik:
Another option is to expect embedded data, and id-based linking can happen based on an `id` element in the embedded data.

bdarcus:
> If CSL-JSON is intended to be used solely as a data format for citation processors ...

I think some of the changes we've discussed and implemented are based on the recognition that the format is and (particularly it's YAML variant) will be used for more than just citation processing.

Eg there will be some people that use it much like many people use bibtex now; an exchange format, but also as a plain-text database.

dstillman:
> But to be concrete, adding the linked variant would require of tools no new code to parse the items, but some simple change to resolve the linked item, while adding the embedded option would involve trivial changes to parse the items

Embedded items would make no practical difference in terms of handling. That's not the case with linked items, which would change the way the file had to be processed. It would no longer be possible to import items one by one, since tools would need to process the entire file before it could import all items, which has implications for performance and memory usage with large files (not hard to overcome, but there would be better and worse ways). It would also mean that users could no longer naively split up files for debugging broken entries or managing memory usage, since they'd risk creating missing dependencies. The ability to separate entries at random is part of what makes RIS vastly easier to debug and handle than RDF.

This isn't hard stuff, but it is a pretty fundamental change, so I'd encourage you to make sure you really need it before agreeing to the additional complexity.

bdarcus:
I just pushed a commit that replaced the linked option with the embedded one.

schemas/input/csl-data.json
```diff
@@ -591,6 +178,453 @@
...
+        },
+        "event-date": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "issued": {
+          "$ref": "#/definitions/date-variable"
+        },
+        "original-date": {
+          "$ref": "#/definitions/date-variable"
+        },
```
zuphilip, line 397: Why is this not deprecated with the `original` relation?

bdarcus:
Just an oversight. There's still some debate about whether we should
deprecate these.

I take it you're in favor?

schemas/input/csl-data.json
bdarcus: Just an oversight. There's still some debate about whether we should deprecate these.

I take it you're in favor?

zuphilip: Okay, I understand. The debate here is too lengthy for me to have already a clear opinion of whether to deprecate these terms at all. But you want to do it, then it should be IMO consistent.

I guess that deprecating/deleting some fields in the schema is the one thing, but supporting it in the implementations like citeproc-js is another thing, which we cannot change quickly (to much software depend on that).

bwiernik: My preference is to keep the existing variables valid for that reason.

bdarcus:
I'm not understanding the reticence to deprecate.

The input schema is versioned, as are the styles, and 1.0 data and styles
will still work. Even these deprecated terms would work in 1.1.

Meanwhile, we have an XSLT that will update 1.0 styles to 1.1.

What long-term value is there in retaining them?

What am I missing?

dhimmel:
The longer a deprecation notice is in place, the greater the benefit. The whole purpose is to steer users to the new implementation and alert them that the feature might be removed in the future. So the sooner the deprecation notice, the better the user experience. The deprecated fields can remain in the schema as long as we feel is necessary. But no reason not to give notice ASAP?

schemas/input/csl-data.json
bdarcus: I added the same deprecate not to address @zuphilip's comment.

We can make a final decision on whether to use "deprecate" or not separately.

I vote for deprecate, if we follow up with a PR that changes the file to an object with some top-level metadata, including a version property. This would make supporting multiple versions much easier for tools.

bdarcus:
Merged. We can tweak going forward.

georgd:
I see that only review and original relations are defined in the commit. Is there a reason why chapter ~ edited book is left out? 

Also, wouldn’t it make sense to add the juridic parallel citations or should that be left to the mlz variant? Contrary to the chapter relationship I’m not sure about the latter in standard csl.

bdarcus:
I'm personally open to both.

We started with baby steps, of sorts.

We could add the chapter-book (in) relation indeed, but then to where to
stop? Use the same relation for articles in journals?

Papers (presented at) conferences?

Not rhetorical questions; WDYT?

bdarcus:
Basically, in the CSL model, container and collection are implicit
relations.

denismaier:
> Is there a reason why chapter ~ edited book is left out?

Nothing specific I think. We've discussed it somewhere along the way, but the idea was just to start with two relations, and, if that works, to expand the model later. (see: https://github.com/citation-style-language/schema/issues/167#issuecomment-650590297)

georgd:
> We could add the chapter-book (in) relation indeed, but then to where to stop? Use the same relation for articles in journals? Papers (presented at) conferences?

These are indeed legitimate questions and I think handling of journals, series and conferences could be improved in Zotero et al. E.g. some interface for controlling and unifying journal-titles and -abbreviations.

But for CSL I’d look if it has a relevance in citation and bibliography formatting. For these three, I at least haven’t come across requirements in citation styles yet which make the related structure necessary.

bwiernik:
What I think for these:
- Containers
  - Following discussions about practical features, I suggest we we _not_ hierarchical-ize container, collection, volume. Instead, if we want to support the sort of related-container citations, I suggest we add `container-id` as a variable that can induce that. Or if we switch to a hierarchical structure, we stick with the embedded item idea and only have id to accommodate linking across items. 

- Legal related items
  - I suggest we adopt `legal-see-also` as the term here. Jurism/CSLm just uses `seeAlso`, but I think a more precise label would be valuable. @fbennett

bwiernik:
The question of abbreviations, I suggest should be handled by adopting lists of abbreviations for events and container titles. Zotero currently will automatically generate journal abbreviations from the MEDLINE list and has some hack-y methods to specify a different list. We should discuss whether a formally specified abbreviation list mechanism makes sense. I think it does.

denismaier:
> But for CSL I’d look if it has a relevance in citation and bibliography formatting.

That's indeed a good criterium, and would warrant for adding such a relation. But here again, there's the question of how that should be added...



>     * Following discussions about practical features, I suggest we we _not_ hierarchical-ize container, collection, volume. Instead, if we want to support the sort of related-container citations, I suggest we add `container-id` as a variable that can induce that

So, you're saying each item should in fact be self-contained, but we e.g. add a test to check if more item from the same container are cited?

```xml
<if more-from-same-container="true">
<text macro="citation-with-subsequent-container-infos">
<if>
<else>
<text macro="regular-citation"/>
</else>
```

georgd:
>   * Following discussions about practical features, I suggest we we _not_ hierarchical-ize container, collection, volume. Instead, if we want to support the sort of related-container citations, I suggest we add `container-id` as a variable that can induce that. Or if we switch to a hierarchical structure, we stick with the embedded item idea and only have id to accommodate linking across items.

I think, keeping the flat structure if possible while still enabling hierarchical citation requirements sounds good. I like the approach of bib(la)tex where the hierarchical structure is optional. But that approach isn’t possible here, it seems.

denismaier:
> We should discuss whether a formally specified abbreviation list mechanism makes sense. I think it does.

Absolutely! Even more now that we are adding a `classic` item type, and that `citation-label` can be used as `hereinafter` in CSLm.

bdarcus:
> I like the approach of bib(la)tex where the hierarchical structure is optional. But that approach isn’t possible here, it seems.

It was initially, but I removed that, at least for now, because of concern (see above) about complexity.

denismaier:
> I like the approach of bib(la)tex where the hierarchical structure is optional. But that approach isn’t possible here, it seems.

Yeah, biblatex is in a league of its own here. You can use `crossref`, which establishes a parent-child relationship and inherits data,  or `xref` which does establish a relationship, but does not inherit data, and there's also the powerful `related` feature.

PR #271, opened 2020-06-24
Add sub/main form values to cs:text

bdarcus:
## Description

This moves specification of split title components from dedicated variables to new "sub" and "main" `@form` values.

Removes the -short title variable variants as well.

## Type of change

### DIFF
--- a/schemas/styles/csl-variables.rnc
+++ b/schemas/styles/csl-variables.rnc
@@ -3,14 +3,14 @@ namespace a = "http://relaxng.org/ns/compatibility/annotations/1.0"
 
 ## Variables
 div {
-  
+
   ## All variables
   variables = variables.dates | variables.names | variables.standard
-  
+
   ## Standard variables
   variables.standard =
     variables.numbers | variables.strings | variables.titles
-  
+
   ## Date variables
   variables.dates =
     "accessed"
@@ -20,7 +20,7 @@ div {
     | "issued"
     | "original-date"
     | "submitted"
-  
+
   ## Name variables
   variables.names =
     "author"
@@ -44,7 +44,7 @@ div {
     | "recipient"
     | "reviewed-author"
     | "translator"
-  
+
   ## Number variables
   variables.numbers =
     "chapter-number"
@@ -63,22 +63,16 @@ div {
     | "printing"
     | "supplement"
     | "volume"
-  
+
   ## Title variables
   variables.titles =
     "collection-title"
-    | "collection-title-short"
     | "container-title"
-    | "container-title-short"
     | "original-title"
-    | "original-title-short"
     | "reviewed-title"
-    | "reviewed-title-short"
     | "volume-title"
-    | "volume-title-short"
     | "title"
-    | "title-short"
-  
+
   ## String variables
   variables.strings =
     "abstract"
@@ -122,10 +116,7 @@ div {
     | "section"
     | "source"
     | "status"
-    | "title"
-    | "title-short"
     | "translated-title"
-    | "translated-title-short"
     | "URL"
     | "version"
     | "year-suffix"
--- a/schemas/styles/csl-variables.rnc
+++ b/schemas/styles/csl-variables.rnc
@@ -67,28 +67,16 @@ div {
   ## Title variables
   variables.titles =
     "collection-title"
-    | "collection-title-main"
-    | "collection-title-sub"
     | "collection-title-short"
     | "container-title"
-    | "container-title-main"
-    | "container-title-sub"
     | "container-title-short"
     | "original-title"
-    | "original-title-main"
-    | "original-title-sub"
     | "original-title-short"
     | "reviewed-title"
-    | "reviewed-title-main"
-    | "reviewed-title-sub"
     | "reviewed-title-short"
     | "volume-title"
-    | "volume-title-main"
-    | "volume-title-sub"
     | "volume-title-short"
     | "title"
-    | "title-main"
-    | "title-sub"
     | "title-short"
   
   ## String variables
--- a/schemas/styles/csl-variables.rnc
+++ b/schemas/styles/csl-variables.rnc
@@ -69,9 +69,11 @@ div {
     "collection-title"
     | "container-title"
     | "original-title"
+    | "part-title"
     | "reviewed-title"
-    | "volume-title"
     | "title"
+    | "translated-title"
+    | "volume-title"
 
   ## String variables
   variables.strings =
@@ -105,8 +107,6 @@ div {
     | "note"
     | "original-publisher"
     | "original-publisher-place"
-    | "original-title"
-    | "part-title"
     | "PMCID"
     | "PMID"
     | "publisher"
@@ -116,7 +116,6 @@ div {
     | "section"
     | "source"
     | "status"
-    | "translated-title"
     | "URL"
     | "version"
     | "year-suffix"
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -125,7 +125,7 @@ div {
   info.id =
     
     ## Specify the unique and stable identifier for the style. A URI
-    ## is valid, but new styles should use a UUID to ensure stability 
+    ## is valid, but new styles should use a UUID to ensure stability
     ## and uniqueness.
     element cs:id { xsd:anyURI }
   info.issn =
@@ -321,6 +321,11 @@ div {
       attribute name { xsd:NMTOKEN },
       rendering-element+
     }
+}
+# ==============================================================================
+
+## Rendering Elements
+div {
   rendering-element =
     rendering-element.names
     | rendering-element.date
@@ -641,10 +646,11 @@ div {
     | 
       ## Specify verbatim text.
       attribute value { text }
-    | (
-       ## Select a variable.
-       attribute variable { variables.standard },
-       [ a:defaultValue = "long" ] attribute form { "short" | "long" }?)
+    | ((attribute variable { variables.numbers | variables.strings },
+        [ a:defaultValue = "long" ] attribute form { "short" | "long" })
+       | (attribute variable { variables.titles },
+          [ a:defaultValue = "long" ]
+          attribute form { "short" | "long" | "sub" | "main" })?)
 }
 # ==============================================================================
 
@@ -909,7 +915,7 @@ div {
     [ a:defaultValue = "simple" ]
     attribute title-split {
       
-      ## Matches "."", ":", "::", "!", "?" 
+      ## Matches "."", ":", "::", "!", "?"
       "simple"
       | 
         ## Matches the values for "simple" plus ";"
@@ -918,7 +924,7 @@ div {
         ## Matches the values for "simple" plus "—" and ";"
         "full"
       | 
-        ## Matches the values for "simple" plus ";"" 
+        ## Matches the values for "simple" plus ";""
         ## and "[;,] or[,:]" (see CMoS 14.91), e.g.  "Moby-Dick; or, The Whale"
         "chicago"
     }?,


bwiernik:
Good call. `cs:text` is the only place where title variables can currently be rendered, so that's fine. 

Is it possible to validate a style so that the `variable` attribute on `cs:text` must be a `title`variable` in the schema `form="main"` or `form="sub"`? More generally, what is the expected fallback behavior? To `form="long"` if `form="main"` is empty and to empty if `form="sub"` is emptry?

I think that the `rendering-attribute.form.values`, which always contains `"long"` and `"short"`, makes the schema less clear. Is this change just a Don't Repeat Yourself idea in case we ever had another standard form?

bdarcus:
> Is it possible to validate a style so that the `variable` attribute on `cs:text` must be a `title`variable`in the schema`form="main"`or`form="sub"`? 

Yes it is. It's just not easy, so I want confirmation before doing it.

> More generally, what is the expected fallback behavior? To `form="long"`if`form="main"`is empty and to empty if`form="sub"` is empty?

I think so. The case that would be relevant here is a simple title like "A Book."

> I think that the `rendering-attribute.form.values`, which always contains `"long"` and `"short"`, makes the schema less clear. Is this change just a Don't Repeat Yourself idea in case we ever had another standard form?

Yes, but also, these two issues are, I think, related. I was planning to build on this to resolve the first point.

But I can do that other ways; it's not necessary.

bwiernik:
I would like the validator to reject something like `<text variable="volume" form="main"/>`, so I think restricting the main and sub forms to the title variables would be good if it's a reasonable effort. If `rendering-attribute.form.values` helps with that, great.

bdarcus:
> I would like the validator to reject something like `<text variable="volume" form="main"/>`, so I think restricting the main and sub forms to the title variables would be good if it's a reasonable effort. If `rendering-attribute.form.values` helps with that, great.

Should work as expected now.

I removed the additional stuff I did earlier, so this is mostly just adding this.

bdarcus:
I also just removed the sub/main title variables.

Which raises a question I think I should ask at this point: should we move "-short" titles to this mechanism too?

denismaier:
Yes. Short should be moved to this too. But short is relevant for more than just titles.
Also: how will we deal with a this in the json?

bdarcus:
We could deal with those questions separately?

For now, just about titles and text.

bdarcus:
Actually, short variables are only on titles, so I think this only applies to titles.

denismaier:
But the plan was to extend short to other variables as well, right?

bdarcus:
So two options (I don't care):

1. keep as is, and have a separate PR specific to the broader "short" question
2. add "short" specific to titles here, and the rest on a separate PR

Preference?

bdarcus:
In any case, when you guys are ready to GTM, please approve the PR.

bwiernik:
Let’s remove short from rnc. Should stay in some form in JSON. Can’t see How to access the review system on mobile, but this looks good to me

bwiernik:
Let’s remove short from rnc. Should stay in some form in JSON. Can’t see How to access the review system on mobile, but this looks good to me

denismaier:
But question: how can I checkout a PR for testing?

bdarcus:
> Git question: how can I checkout a PR for testing?

Here's one, direct, way.

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally

Can also fetch the branch, then check it out.

bwiernik:
`git fetch origin <PR number>`
`git checkout origin <PR number>`

denismaier:
Concerning json input: @dhimmel has indicated elsewhere that we might find a solution that doesn't involve polluting the json schema with `-short` variables. Maybe that can be of use here? 

I was only able to find [this comment](https://github.com/citation-style-language/schema/pull/263#issuecomment-646837561) by @dhimmel on this topic, but maybe he can give some more hints on what would be involved, and on potential drawbacks.

schemas/styles/csl-variables.rnc
```diff
@@ -122,10 +116,7 @@ div {
     | "section"
     | "source"
     | "status"
-    | "title"
-    | "title-short"
     | "translated-title"
```
denismaier, line 119: `translated-title` should be under "variables.titles", shouldn't it?

bwiernik:
Should `original-title`, `part-title`, `translated-title` be listed under title variables?

bdarcus:
> Should `original-title`, `part-title`, `translated-title` be listed under title variables?

I was assuming no, so in the process of removing them from there.

@denismaier?

dhimmel:
> I was only able to find this comment by @dhimmel on this topic, but maybe he can give some more hints on what would be involved, and on potential drawbacks.

@denismaier I don't understand this PR (and the RNC side of things) sufficiently to weight in here.

It looks like their are four possible forms: `{ "short" | "long" | "sub" | "main" }`. Anywhere I can read what these mean?

> we might find a solution that doesn't involve polluting the json schema with -short variables.

As far as the CSL JSON goes, given that `-short` fields are already part of it, I'm not sure it makes sense to pursue an alternative. Although there are possibly some backwards compatible alternatives, like allowing the following syntax

```yaml
container-title:
  long: PLOS Computational Biology
  short: PLOS Comp Bio
```

As opposed to the traditional:

```yaml
container-title: PLOS Computational Biology
container-title-short: PLOS Comp Bio
```

But keeping this traditional syntax valid (at least for `container-title`). Makes it more complicated for the citeproc processor.

denismaier:
> > Should `original-title`, `part-title`, `translated-title` be listed under title variables?
> 
> I was assuming no, so in the process of removing them from there.

Why no? As they are clearly titles, I guess they should be there as well.

bdarcus:
> > > Should `original-title`, `part-title`, `translated-title` be listed under title variables?
> > 
> > 
> > I was assuming no, so in the process of removing them from there.
> 
> Why no? As they are clearly titles, I guess they should be there as well.

Because they were listed in under the "strings" list; not titles.

See the latest commit and confirm that looks right?

bdarcus:
> > I was only able to find this comment by @dhimmel on this topic, but maybe he can give some more hints on what would be involved, and on potential drawbacks.
> 
> @denismaier I don't understand this PR (and the RNC side of things) sufficiently to weight in here.
> 
> It looks like their are four possible forms: `{ "short" | "long" | "sub" | "main" }`. Anywhere I can read what these mean?

Not written yet. But, examples:

``` yaml
title: "Some Title: And a Subtitle"
main: Some Title
sub: And a Subtitle
```
"Short": when we originally defined it, I thought of it as alias for what we now introducing as "main." But it seems for very long titles, they may differ.

These are style details, BTW; not necessarily translating to the input json.

denismaier:
> "Short": when we originally defined it, I thought of it as alias for what we now introducing as "main." But it seems for very long titles, they may differ.

Yes, they are not the same. `short` sometimes be equivalent to `main`, or a shorter version of `main`. Or, they are almost same, but `short` drops a leading article:

```yaml
title: "An important title: and a subtitle"
main: "An important title"
short: "important title"
sub: and a Subtitle
```

denismaier:
> These are style details, BTW; not necessarily translating to the input json.

Yes, but as `main` and `sub` are derived by citeprocs from the full title we need them in input json to provide ways to override the heuristics.

denismaier:
I was hoping that we could get away with some sort of addition to the the specs: "processors should accept -short variants to all variables." Or something similar? Wdyt @dhimmel? Would that be reasonable?

dhimmel:
> I was hoping that we could get away with some sort of addition to the the specs: "processors should accept -short variants to all variables." Or something similar?

And then also add `title-main` and `title-sub`? Is it just titles that will have main & sub, or other variables like `container-title` might have `container-title-sub`?

On one hand, we could keep the number of CSL JSON top-level fields minimal with a syntax like:

```yaml
title:
  long: "An important title: and a subtitle"
  main: "An important title"
  short: "important title"
  sub: and a Subtitle
```

On the other hand, there seems to be a preference for flat CSL JSON (ignoring the funky date-variables nesting). And some of the flat fields already exist like `container-title-short` and backcompat would be good.

If going with the flat structure, I think it's best to update the CSL JSON schema to include these properties. I don't think it makes sense to add `-short` variants where there is not a documented or theoretical need.

To enable the flexibility to add suffixes to CSL fields, we could look into [JSON Schema pattern properties](https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties). This would allow us to keep the number of schema definitions from exploding:

```json
{
  "type": "object",
  "patternProperties": {
    "^title(-long|-sub|-main|-short)?": { "type": "string" }
  },
  "additionalProperties": false
}
```

bdarcus:
> And then also add `title-main` and `title-sub`?

I don't believe that will be necessary, as these can be derived from the full title.

> To enable the flexibility to add suffixes to CSL fields, we could look into JSON Schema pattern properties.

It also occurred to me recently that we could use this mechanism, and possibly a similar one in rng, to allow prefix extension data; like `law:foobar`. 

That would be a big decision to make, but it seems easy enough technically. And we could couple it, say, with a wiki to collect the extensions.

denismaier:
> > I was hoping that we could get away with some sort of addition to the the specs: "processors should accept -short variants to all variables." Or something similar?
> 
> And then also add `title-main` and `title-sub`? Is it just titles that will have main & sub, or other variables like `container-title` might have `container-title-sub`?

Yes, also `-main` and `-sub`. This should apply to all variables under `variables.titles`, so 

```
collection-title
container-title
original-title
part-title
reviewed-title
title
translated-title
volume-title
```

Each of those will have `-short`, `-main`, and `-sub`

> 
> On one hand, we could keep the number of CSL JSON top-level fields minimal with a syntax like:
> 
> ```yaml
> title:
>   long: "An important title: and a subtitle"
>   main: "An important title"
>   short: "important title"
>   sub: and a Subtitle
> ```
> 
> On the other hand, there seems to be a preference for flat CSL JSON (ignoring the funky date-variables nesting). And some of the flat fields already exist like `container-title-short` and backcompat would be good.
> 
> If going with the flat structure, I think it's best to update the CSL JSON schema to include these properties. I don't think it makes sense to add `-short` variants where there is not a documented or theoretical need.
> 
> To enable the flexibility to add suffixes to CSL fields, we could look into [JSON Schema pattern properties](https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties). This would allow us to keep the number of schema definitions from exploding:
> 
> ```json
> {
>   "type": "object",
>   "patternProperties": {
>     "^title(-long|-sub|-main|-short)?$": { "type": "string" }
>   },
>   "additionalProperties": false
> }
> ```

This looks promising. This will work for `title`, right? 
What's the best way to add the prefix patterns? Does that work?

```json
{
  "type": "object",
  "patternProperties": {
    "^(container-|collection-)?title(-long|-sub|-main|-short)?$": { "type": "string" }
  },
  "additionalProperties": false
}
```

Or should we add one pattern per title?


```json
{
  "type": "object",
  "patternProperties": {
    "^title(-long|-sub|-main|-short)?$": { "type": "string" },
    "^container-title(-long|-sub|-main|-short)?$": { "type": "string" },
    "^collection-title(-long|-sub|-main|-short)?$": { "type": "string" }
  },
  "additionalProperties": false
}
```

Readability is better here, but it is redundant, of course. Maybe something like this?


```json
{
  "type": "object",
  "patternProperties": {
    "^(container-
        |collection-
        |volume-)?
        title(-long|-sub|-main|-short)?$": { "type": "string" }
  },
  "additionalProperties": false
}
```

denismaier:
> > And then also add `title-main` and `title-sub`?
> 
> I don't believe that will be necessary, as these can be derived from the full title.

Sure we need them. Heuristics can fail and we need to provide some way to override those derived variables.

bdarcus:
I'm starting to get lost in all this.

Didn't we just agree to add language to the specification that would mean this is completely unnecessary? 

E.g. if a processor sees `Title: Subtitle` it will split know how to access `main` and `sub`?

denismaier:
> What applications or legacy data formats will allow a user to manually specify these though?

In Zotero you will just add `title-main` to the `extra` field.

The assumption is: Users enter the full title in the `title` field. The processor splits the title up into `title-main` and `title-sub`. Now: If you have a complicated title where that does not work, you will add `title-main` via `extra` so you can indicate where the processor should split.

denismaier:
Concerning biblatex: Biblatex's `title` is `title-main` in CSL.
We can't expect end user oriented applications to switch to this two field solution. That's why we came upt with this splitting mechanism in the first place.

dhimmel:
> Heuristics can fail and we need to provide some way to override those derived variables.

I agree that there should be a way to manually specify `sub-` and `main-` to address situation where the heuristic fails.

> What applications or legacy data formats will allow a user to manually specify this though?

None yet, but perhaps in the future some. For Manubot we edit CSL JSON by hand when automated generation fails. Our reference manager is a bunch of persistent identifiers for which CSL JSON is automatically generated. And when that fails, the user manually provides the CSL JSON.

> Or should we add one pattern per title?
>
> ```json
> {
>   "type": "object",
>   "patternProperties": {
>     "^title(-long|-sub|-main|-short)?$": { "type": "string" },
>     "^container-title(-long|-sub|-main|-short)?$": { "type": "string" },
>     "^collection-title(-long|-sub|-main|-short)?$": { "type": "string" }
>   },
>   "additionalProperties": false
> }
> ```

I like this format because we should be defining JSON Schema's `title` and `description` differently for each of these variables (see https://github.com/citation-style-language/schema/issues/190, we should confirm `title` and `description` work with pattern properties). I think it's okay that `title-long` and `title-short` and the rest of the title family share documentation. Also it's helpful in terms of looking through the schema to see what variables are available.

bwiernik:
> In Zotero you will just add title-main to the extra field.

To elaborate on this, what this behavior is is that citeproc-js has behavior to extract CSL variables from the top of `note` (which is **Extra** in Zotero and **Note** in Mendeley). I've opened an issue to discuss whether we might want to add this behavior formally to the spec. https://github.com/citation-style-language/documentation/issues/97

PR #273, opened 2020-06-24
Add editor-translator as testable variable

bwiernik:
## Description

Add editor-translator to the RNC schema to make it a testable variable. This makes it possible to modify the formatting of editor and translator depending on if they are the same person or not. The variable itself is still populated by the processor collapsing editor and translator.

This would require a minor update to processor code. They already have the editortranslator collapsing code; but they would likely need a small modification to call it during and if test. Given that, probably v1.1, rather than 1.0.2?

Fixes part of # citation-style-language/csl-evolution#7

## Type of change

### DIFF
--- a/schemas/styles/csl-variables.rnc
+++ b/schemas/styles/csl-variables.rnc
@@ -33,6 +33,7 @@ div {
     | "curator"
     | "director"
     | "editor"
+    | "editor-translator"
     | "editorial-director"
     | "executive-producer"
     | "illustrator"


bdarcus:
That linked issue contemplates other solutions. Why this one (which admittedly has the benefit of simplicity).

Just wondering what about kinds of needs there may be like this.

bwiernik:
I think this one is clearer for casual style authors than Frank's `require-match` option, and it has the elegance of the `editortranslator` term already existing. I don't really see other creator combinations (e.g., "Director & Producer") having the same need to be tested.

bdarcus:
Or writer-director?

>

bwiernik:
The unusual case with `editor-translator` is that you might have something like this:
`Editor, A. A. (Ed.). (2019). Book title (B. B. Translator, Trans.). Publisher.`

which should become:
`Editor, A. A. (Ed. & Trans.). (2019). Book title. Publisher.`

if editor and translator are the same. But it should not be:
`Editor, A. A. (Ed.), & Translator, B. B. (Trans.). (2019). Book title. Publisher.`

For the media roles (writer, director, producer), the creators are pretty much always listed together in a group. So collapsing can be handled with the proposed `multiple="combined"` syntax described in https://github.com/citation-style-language/csl-evolution/issues/7

PR #274, opened 2020-06-24
Move "section", "version" strings to numbers

bwiernik:
## Description

What it says: Move `section` to numbers. Makes `label="section"` available.

## Type of change

### DIFF
--- a/schemas/styles/csl-variables.rnc
+++ b/schemas/styles/csl-variables.rnc
@@ -60,6 +60,7 @@ div {
     | "page-first"
     | "part"
     | "printing"
+    | "section"
     | "supplement"
     | "volume"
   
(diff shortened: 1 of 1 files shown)


bdarcus:
I'm forgetting the implications of this, but what's the use cases?

What happens if you have a document with unnumbered sections?

bwiernik:
The main benefit is that it makes `label="section"` available (currently, you have to use `term="section"`. A variable being a number means that it can be used with `cs:label` and can be formatted as ordinal, etc. You can still render a number variable with `cs:text` (e.g., most styles use `<text variable="volume"/>` instead of `<number variable="volume"/>`

Thinking about it, version should probably also be a number for these reasons.

Issue #277, opened 2020-06-26
Provide a way for CSL JSON to store extra / custom fields


dhimmel:
In https://github.com/citation-style-language/documentation/issues/97, there's been much discussion on whether to add a way for CSL JSON to specify extra fields that are not part of the spec currently.

Currently, the CSL JSON Data Schema sets the following for items:

https://github.com/citation-style-language/schema/blob/c2b2eee63e45e964f08dc56494bea85ad2a59d51/schemas/input/csl-data.json#L426

This means that additional fields are not allowed as per the schema.

Some points in favor and against that I'm migrating here:

By @dhimmel at https://github.com/citation-style-language/documentation/issues/97#issuecomment-649782124:

> Custom variables are inevitable and required for certain use cases. We use the note cheater syntax in Manubot to store variants of id ([example](https://github.com/manubot/manubot/blob/47b03e0c202e6a0561e6548ba08a4ad790f56eee/manubot/cite/tests/test_csl_item.py#L110-L126)). Since custom variables are inevitable, I think it's best to consider the most effective and least problematic method for storing them.

By @bwiernik in https://github.com/citation-style-language/documentation/issues/97#issuecomment-650152255:

> I'll reiterate my opinion that CSL should not accommodate non-CSL variables. A major strength of CSL compared to BibTeX, in my opinion, is the way it has not embraced user-specific data entry. I believe that Sebastian has expressed similar sentiments.

From @bdarcus in https://github.com/citation-style-language/documentation/issues/97#issuecomment-650157412:

> Why do we care what people or applications put in the input data, if we don't care what they embed in the note field in some non-standard syntax?
>
> CSL JSON is being used for data exchange, and I certainly wouldn't want my custom data stripped from the output, even if it has no impact on citation processing. I might want to run a script on that output, for example, to generate some other output.

I will follow up with an additional post about possible implementations. But wanted to get this issue started to keep discussion consolidated.

dhimmel:
Possible implementations based on a [Manubot-outputed CSL JSON example](https://github.com/greenelab/connectivity-search-manuscript/blob/70b1da8ac787ed9d8f3ee2aa763a8a16fb66b7f3/references.json#L898-L958):

```yaml
# method 1. current
- id: O21tn8vf
  type: article-journal 
  note: "standard_id: doi:10.7554/elife.26726"
# method 2. extra variables in item root (currently forbidden)
- id: O21tn8vf
  type: article-journal 
[...]

PR #280, opened 2020-06-26
CSL JSON Data schema: add custom field

dhimmel:
closes https://github.com/citation-style-language/schema/issues/277

let me know if any other files need to be changed. I added `title` and `description` as per #190. Hopefully we can eventually auto-generate CSL JSON docs from this... but are there docs that need to be updated manually now?

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -417,6 +417,15 @@
       },
       "year-suffix": {
         "type": "string"
+      },
+      "custom": {
+        "title": "Custom key-value pairs.",
+        "type": "object",
+        "description": "Used to store additional information that does not have a designated CSL JSON field. The note field can also store additional information, but custom is preferred for storing key-value pairs.",
+        "examples" : [
+          {"short_id": "xyz", "other-ids": ["alternative-id"]},
+          {"metadata-double-checked": true}
+        ]
       }
     },
     "required": ["type", "id"],


bdarcus:
Thanks!

This is currently set to merge to master, feasibly for 1.0.2, which I am tagging today.

Should we do this now, or for 1.1?

I'm agnostic.

@bwiernik and @denismaier can answer your questions, as I'm unsure.

dhimmel:
> This is currently set to merge to master, feasibly for 1.0.2, which I am tagging today.

It is backwards compatible (as far as I can tell), so not huge harm. But it is a new feature, so if you want to follow semantic versioning, it would go in the next MINOR release... but I am guessing much of the stuff in 1.0.2 is actually new features and not backcompat bug fixes.

bdarcus:
> I am guessing much of the stuff in 1.0.2 is actually new features and not backcompat bug fixes.

1.0.2 is mostly new variables and types; strings.

denismaier:
As it seems, semantic versioning does not really work as is for xml schemas, due to validation and so. Or at least, we haven't figured out how to do it properly... ;-)
So yeah 1.0.2 adds new features, but requires relatively minor changes in processors.

denismaier:
Concerning the script: we shouldn't worry about this now. I'm not so sure we will need this if we adopt `@related` and "sub/main" as forms. These will solve the same sort of problems the script was trying to solve. But anyway, adapting the script will be easy enough.

dhimmel:
> Concerning the script: we shouldn't worry about this now. I'm not so sure we will need this if we adopt @related and "sub/main" as forms. These will solve the same sort of problems the script was trying to solve.

Ah cool. Just made a related post at https://github.com/citation-style-language/schema/issues/223#issuecomment-650258108.

bdarcus:
Partly for marketing reasons, then, how about we add this to 1.1?

dhimmel:
> Partly for marketing reasons, then, how about we add this to 1.1?

Fine with me. I think as time goes on, you might want to go with just MANOR.MINOR since we don't really have the concept of PATCHES for this project.

No rush on my end to have this official

dhimmel:
Changing the base branch to v1.1 messed stuff up.

@bdarcus is that the correct branch to merge to? If so, I fix this PR so it's only one commit?

bdarcus:
> Changing the base branch to v1.1 messed stuff up.
> 
> @bdarcus is that the correct branch to merge to? If so, I fix this PR so it's only one commit?

Yes. I ran into the same problem a few days ago; did we do something wrong with that branch or something?

dhimmel:
> I ran into the same problem a few days ago; did we do something wrong with that branch or something?

Well currently v1.1 is "27 commits ahead, 23 commits behind master.". But since v1.1 hasn't been released yet, I don't think v1.1 branch needs to exist yet and all new work should be going in master. `v1.0` should accumulate commits that will just go into `v1.0.2`.

But it seems like this won't be the case for 1.1. So will switch to the 1.1 branch.

bdarcus:
So we did it backwards; should have done minor work on the 1.0 branch, and major work on master.

> But it seems like this won't be the case for 1.1. So will switch to the 1.1 branch.

Could/should we fix this now?

I was planning to tag a 1.0.2-pre and give some time for feedback.

So should I merge master to 1.0, and tag there, and then merge 1.1 to master?

dhimmel:
> So we did it backwards; should have done minor work on the 1.0 branch, and more major work on master.

Yes. Oftentimes, you might just be picking commits from master to cherry-pick onto the `1.0.2` branch. Everything goes into master, and then stuff is back-ported that needs to be. In the future, it's probably easiest if you don't have to upkeep old versions, because the process is annoying.

> Could/should we fix this now?

Sounds like 1.0.2 is coming out soon. Maybe wait until 1.0.2 is released. After that you can merge 1.1 into master and then get rid of 1.1 until you are ready to release it.

bdarcus:
OK, I just updated the 1.0 branch and tagged v1.0.2-pre.1 there. Is
that part right, or do I also need to tag on master?

We'll hold off a bit on merging 1.1.

dhimmel:
> OK, I just updated the 1.0 branch and tagged v1.0.2-pre.1 there.

That is good.

> Is that part right, or do I also need to tag on master?

No. I don't think tags point to a branch at all. They point to a commit hash. If that commit is on v.1.0 and master, the tag doesn't specify which particular branch. That being said, a 1.0.* tag should point to a commit that exists on the v.1.0 branch.

bwiernik:
Yeah, think of tags as signposts. "This is the point you should regard as the v1.0.2-pre.1 release."

Issue #103, opened 2012-09-04
Add "op. cit." term


rmzelle:
See http://forums.zotero.org/discussion/24804/some-csl-lithuanian-locale-translation-problems/#Item_21

denismaier:
"Op. cit." and "loc. cit." are evil. And they are not the same as "ibid". See: https://en.wikipedia.org/wiki/Op._cit.

Ibid. => the same citation as cited immediately before.
Op. cit. => the cited work (by a given author); replaces publication details given in a prior full citation, would take effect on `position="subsequent"`.
loc. cit. => basically the same as "op. cit." but replaces also the page number.

The problem with them is that they are/were sometimes used in more complex ways:

```
1. John Doe, Book A (Location: Publisher).
2. John Doe, Book B (Location: Publisher).
3. Doe, Book A, op. cit.
4. Doe, Book B., op. cit.
5. Jane Smith, Book C (Location: Publisher).
6. Doe, op. cit. => This refers to Doe, Book B
7. Doe, Book A, op. cit.
```

So, to support this we'd have to introduce new options for position. (e.g. `position="subsequent-same-title-as-last-title-by-same-author"`)

bdarcus:
When we add this to the documentation, I suggest we recommend against its use, and cite Chicago (quoted in the wikipedia article denis links to). It really doesn't belong in 21st century texts.

denismaier:
And we have to keep in mind that the term is only part of the solution.

PR #283, opened 2020-06-27
Add new terms

bwiernik:
## Description

Add various new terms needed for APA, MLA, or other styles or to match new variables added to CSL 1.0.2.

The change in of `article` to `article-locator` is because we have the `article` type as well. 

A few questions:
1. Thinking about `"no-name"`, since is is used to mean "no-publisher", should we just call it that, even though the Latin literally means "no name"?
2. I'd like to be able to have a term for both "review of" and "review of the" to be able to accomodate both of these situations:
    - `Review of "Strangers on a Train"`
    - `Review of the film "Strangers on a Train"`

    Could we add `form="with-article"` and `form="with-article-short"` to meet that? Data conventions in languages with gendered articles are to store the article in the data, whereas in English article aren't stored in the data (cf. the `presented at` term includes an article in the `en` locales, but not in the `fr` locales).
3. Do we want to remove the "category" terms now or in 1.1?
4. Should we fix the `sub verbo` term bug now or stick with the solution Rintze created?

Closes https://github.com/citation-style-language/schema/issues/103
Closes https://github.com/citation-style-language/csl-evolution/issues/34

## Type of change

- [ ] This change requires a documentation update


bwiernik:
@bdarcus I don't understand what this check means:
```
 Check schema formatting0s
  shell: /bin/bash -e {0}
Run for f in csl*.rnc; do diff -q $f schemas/styles/$f ; done
  for f in csl*.rnc; do diff -q $f schemas/styles/$f ; done
  shell: /bin/bash -e {0}
Files csl-terms.rnc and schemas/styles/csl-terms.rnc differ
```

denismaier:
> ## Description
> Add various new terms needed for APA, MLA, or other styles or to match new variables added to CSL 1.0.2.
> 
> The change in of `article` to `article-locator` is because we have the `article` type as well.
> 
> A few questions:
> 
> 1. Thinking about `"no-name"`, since is is used to mean "no-publisher", should we just call it that, even though the Latin literally means "no name"?


> 2. I'd like to be able to have a term for both "review of" and "review of the" to be able to accomodate both of these situations:
>    
>    * `Review of "Strangers on a Train"`
>    * `Review of the film "Strangers on a Train"`
>    
>    Could we add `form="with-article"` and `form="with-article-short"` to meet that? Data conventions in languages with gendered articles are to store the article in the data, whereas in English article aren't stored in the data (cf. the `presented at` term includes an article in the `en` locales, but not in the `fr` locales).

Something like this is definitively needed, yes- But this deserves perhaps its own issue.

> 3. Do we want to remove the "category" terms now or in 1.1?
It seems we are already on the way to 1.1. So it should be 1.1.

> 4. Should we fix the `sub verbo` term bug now or stick with the solution Rintze created?

I'd say this should be fixed. (I'll submit a PR)

> 
> Closes #103
> Closes [citation-style-language/csl-evolution#34](https://github.com/citation-style-language/csl-evolution/issues/34)
> 
> ## Type of change
> Please delete options that are not relevant.
> 
> * [x]  Variable addition (adds a new variable or type string)
> * [ ]  This change requires a documentation update

denismaier:
> * Thinking about `"no-name"`, since is is used to mean "no-publisher", should we just call it that, even though the Latin literally means "no name"?

So you're saying this is not indended for works without an author? (I now there's the term anonymous, but I'm not sure this is 100% the same thing as `no-author`.

bwiernik:
Yes, sine nomine (what `no-name` was added to cover) is used to mean "no publisher", not "no author": https://en.wikipedia.org/wiki/Sine_nomine

denismaier:
So, then let's just call it `no-publisher`.

bdarcus:
> So, then let's just call it `no-publisher`.

Yes, let's avoid latin unless necessary.

bdarcus:
I just tried to change the base branch to `v1.1`, but it resulted in the earlier problem of a ton of other commits. I don't really understand.

bwiernik:
Can we just merge this into 1.0.2 and cut a new 1.0.2pre2 tag?

bdarcus:
I think yes, for this one.

Do you why this happened?

Because you branches off master?

bwiernik:
Yes, when master is merged into v1.1 (or vice versa), there will be some conflicts that need to be resolved.

bdarcus:
Yes, I am aware.

One of the reasons I'll be happy to get that merged and hopefully just
focus on master going forward.

bwiernik:
I'm happy to do the resolution once we do the gold cut of 1.0.2 and merge 1.1 into master

Issue #278, opened 2020-06-26
CSL YAML


bdarcus:
[This](https://json-schema-everywhere.github.io/yaml) and [this](https://www.npmjs.com/package/pajv) suggests we can use our JSON schemas to validate a YAML alternarnative.

I suggest we do something with this, since it's zero work for us.

_Originally posted by @bdarcus in https://github.com/citation-style-language/schema/issues/277#issuecomment-650192158_

denismaier:
Is pandocs CSL YAML actually identical to CSL JSON? I'm not sure, but don't they differ at least in some aspects. 
E.g. regarding dates: CSL JSON has this construct:
```
"issued":{"date-parts":[[2015,4,2]]},
```

Whereas the YAML:
```yaml
  issued:
    - year: 2015
      month: 4
      day: 2
```

(I think I've tried to use the JSON schema for autocompletion and validation once, but I wasn't so lucky. Using Atom or VS Code as plain text reference managers could be very nice for some projects...)

bdarcus:
> Is pandocs CSL YAML actually identical to CSL JSON?

I hadn't checked, but wondering if devs like @jgm would find value in this.

Advantage is we have one schema, that is always in sync with the CSL spec.

bwiernik:
People definitely seem to like YAML in my research community. Easier to manually edit if needed, like BibTeX or RIS.

bdarcus:
> People definitely seem to like YAML in my research community. Easier to manually edit if needed, like BibTeX or RIS.

Exactly.

YAML is easily hand-editable. JSON isn't.

denismaier:
> YAML is easily hand-editable. JSON isn't.

And with auto-completion it would be even better!

denismaier:
@retorquere Do you have any input here?

bdarcus:
> > YAML is easily hand-editable. JSON isn't.
> 
> And with auto-completion it would be even better!

https://github.com/liuderchi/ide-yaml/issues/56

retorquere:
> Is pandocs CSL YAML actually identical to CSL JSON? I'm not sure, but don't they differ at least in some aspects.
> E.g. regarding dates: CSL JSON has this construct:

To the best of my knowledge, this is the only difference. `circa` and `season` are supported here, `circa` at the same level as `date-parts`, not per-date. But I do not know of a formal spec.

jgm:
> Is pandoc's CSL YAML actually identical to CSL JSON? 

[...]

Issue #251, opened 2020-06-16
add condition: year < 2000


denismaier:
As per @georgd [here](https://discourse.citationstyles.org/t/advice-on-legal-citation-style/1632/13?u=denis_maier):

> At somoe places, years should be formtatted with two digits if year < 2000 and with four digits otherwise. Am I right in assuming that that’s not possible? At least I haven’t found it in the documentation. It’s a requirement that I met in various (legal and non-legal) styles already, especially when dealing with EU sources. 

I think that sound like a reasonable suggestion. Other opinions about this? Anyone else has seen that requirement before?
If we add this: Perhaps it should be done in a not so specific way. Rather than `if after-2000` perhaps something like "`if date-range="1900-1999"` which would be adaptable to other requirements.

denismaier:
comment by @adam3smith 
I don't have a strong opinion on adding this, but I've seen requirements that make the inclusion of publisher place dependent on year (I think pre1900) so if we do this I think it should definitely be more flexible than just after-200

denismaier:
Yeah, MLA does this now that they omit publisher-place for current works. Don't know what APA does in these cases.

denismaier:
@bwiernik do you think we can combine this with #254 ?

bwiernik:
So, one option would be:
`before-1900="issued"` and `before-2000="issued"`

That's compact and consistent with existing attributes of `cs:if`, but not very flexible (it covers only the two cases listed here; but are there really more)? 

Another option would be:
`date-before="issued 1900" and `date-after="issued 1999"`

Here, `date-before` and `date-after` are space-delimited lists with the first element being a date variable and the second a numeric year. 

Another possibility would be similar to @denismaier proposal:
`date-range="issued <1900" or `date-range="issued >1999" or `date-range="issued 1900-1999" 

Here `date-range` again takes a space-delimited list with the first element being a date variable. The second element is either a date range or `<[year]` for before or ">[year]" for after.

The key thing about either the date-before or date-range syntaxes is that they have to specify what date variable is being tested.

> Don't know what APA does in these cases.

APA never gives publisher-place. For archival works, they might give archive-place.

bdarcus:
We need a solution that works in the context of xml processing tools. One question I was looking to find an answer for recently, but did not, was whether attribute value order is preserved in XML. Attributes order itself is NOT.

Anyway, more options:

```
date-range-key='issued' date-range-values='1900-2000'
```
[...]

PR #284, opened 2020-06-28
(input): Add EDTF as an option for date representation

bdarcus:
## Description

As part of #278, and to harmonize the JSON and YAML representations around a much more concise and expressive date format, this adds an option to use [EDTF](https://www.loc.gov/standards/datetime/).

While [EDTF](https://www.loc.gov/standards/datetime/) was originally an initiative of the US Library of Congress, it seems that in 2019, ISO adopted it as part of [8601-2](https://www.iso.org/obp/ui/#iso:std:iso:8601:-2:ed-1:v1:en).

Example of a date-range with a circa qualifier in the new alternative (using the YAML syntax):

``` yaml
issued: 2004-02-01/2005-02-08?
```
-----

## TODO

1. settle on the precise compliance levels and features we support (for documentation); see below
2. modify regex pattern for validation that reflects 1, or that tightens the validation beyond current
3. long-term status of `date-parts`; do we say deprecated now and remove later, for example, or recommend EDTF?

## Compliance Details

We need to say "Level 0 is supported, and in addition the following features of levels 1 and 2 are supported (list features)."

Note: biblatex formally supports levels 0 and 1.

### Level 0

- Date (`2010-03-10`)
- Time Interval (`2010-03-10/2010-03-11`)

Note: level 0 includes date-times, which I don't think we need.

### Level 1

- Seasons (`2010-21`)
- Negative calendar year (`-0500`)

I'm confident in the above; the rest below are uncertainties; I think I'd prefer to not support, on input at least.

#### Uncertain vs Approximate Dates

- Qualification of Date (`2010-03-04?` means "uncertain"; `~` indicates "approximate" and `%` both)

I remember this discussion during EDTF development, but never understood the practical distinction.

In our case, do we support one or the other, or both?

How?

 _I vote for only one_, which we treat as equivalent to our "circa." But no pun intended, I'm not 100% sure.

#### Other Issues

I'm not sure we need this:

- Unspecified digit(s) from the right (`2010-03-XX`)

We need these "Extended Interval " features on styles, but on input?

- Open end time interval (`1900/..`)
- Open start time interval (`../1900`)
- Time interval with unknown end  (`1900/`)
- Time interval with unknown start (`/1900`)

### Level 2

Do we need "all member" sets? (`{2010-03-04,2010-03-05}`)

### Extension

- season ranges (`2020-21/2020-22`).

----

## EDTF libraries

- https://github.com/inukshuk/edtf.js
- https://github.com/ixc/python-edtf

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -477,7 +477,7 @@
       "title": "EDTF datatype pattern",
       "description": "CSL input supports EDTF, validated against this regular expression.",
       "type": "string",
-      "pattern": "[0-9,\-,%,~,X,?,..,\/]"
+      "pattern": "[0-9,-,%,~,X,?,..,\/]"
     },
     "date-variable": {
       "title": "Date content model.",
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -476,7 +476,8 @@
     "edtf-datatype": {
       "title": "EDTF datatype pattern",
       "description": "CSL input supports EDTF, validated against this regular expression.",
-      "type": "string"
+      "type": "string",
+      "pattern": "[0-9,\-,%,~,X,?,..,\/]"
     },
     "date-variable": {
       "title": "Date content model.",
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -477,7 +477,7 @@
       "title": "EDTF datatype pattern",
       "description": "CSL input supports EDTF, validated against this regular expression.",
       "type": "string",
-      "pattern": "[0-9,-,%,~,X,?,..,\/]"
+      "pattern": "^[0-9-%~X?.\/]{4,}$"
     },
     "date-variable": {
       "title": "Date content model.",
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -481,7 +481,7 @@
     },
     "date-variable": {
       "title": "Date content model.",
-      "description": "The CSL input model supports two different date representations: a preferred EDTF string, and a more structured alternative.",
+      "description": "The CSL input model supports two different date representations: an EDTF string (preferred), and a more structured alternative.",
       "anyOf": [
         {
           "$ref": "#/definitions/edtf-datatype"
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -422,9 +422,14 @@
         "title": "Custom key-value pairs.",
         "type": "object",
         "description": "Used to store additional information that does not have a designated CSL JSON field. The note field can also store additional information, but custom is preferred for storing key-value pairs.",
-        "examples" : [
-          {"short_id": "xyz", "other-ids": ["alternative-id"]},
-          {"metadata-double-checked": true}
+        "examples": [
+          {
+            "short_id": "xyz",
+            "other-ids": ["alternative-id"]
+          },
+          {
+            "metadata-double-checked": true
+          }
         ]
       }
     },
@@ -468,8 +473,18 @@
         }
       ]
     },
+    "edtf-datatype": {
+      "title": "EDTF datatype pattern",
+      "description": "CSL input supports EDTF, validated against this regular expression.",
+      "type": "string"
+    },
     "date-variable": {
+      "title": "Date content model.",
+      "description": "The CSL input model supports two different date representations: an EDTF string, and a more structured equivalent.",
       "anyOf": [
+        {
+          "$ref": "#/definitions/edtf-datatype"
+        },
         {
           "properties": {
             "date-parts": {
@@ -496,6 +511,9 @@
             },
             "raw": {
               "type": "string"
+            },
+            "edtf": {
+              "$ref": "#/definitions/edtf-datatype"
             }
           },
           "additionalProperties": false
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -481,7 +481,7 @@
     },
     "date-variable": {
       "title": "Date content model.",
-      "description": "The CSL input model supports two different date representations: an EDTF string, and a more structured equivalent.",
+      "description": "The CSL input model supports two different date representations: a preferred EDTF string, and a more structured alternative.",
       "anyOf": [
         {
           "$ref": "#/definitions/edtf-datatype"


jgm:
Adopting EDTF sounds like a good plan to me -- much more compact, still readable, allows ranges (pandoc currently handles these by using an array of the year/month/day objects).
Also, it makes it impossible to specify, say, just a month but no year.

bdarcus:
So we could define a regex of at least some set of it we support?

I found this, BTW:

https://github.com/richardtallent/EdtfDotNet/blob/master/Edtf/EdtfRegexPattern.txt

bwiernik:
We should just adopt EDTF. It has defined levels of features, right? For citations, we need date ranges, uncertain dates, approximate dates, CE/BCE, seasons, and dates to specified precision (e.g., 194x). I think that covers it. What level of EDTF had that? 

@retorquere you’ve explored this in BBT right? 

@fbennett you’ve also explored accommodating Japanese imperial dates in CSLm. How does that work? Could we incorporate that and extend to Arabic calendar dates?

bdarcus:
I just pushed a commit that removes the date-part pattern and replaces it with a `edtf` key.

Note:

1. I only added that key because I wasn't sure if we could get rid of the `literal` and `raw` variables; would be simpler still if we could.
2. The regex pattern I put there is not right; it's just a placeholder.

FWIW, EDTF has been around for awhile, and I participated in the discussions largely to ensure it would work for our use case. It's been a few years since I have participated, but it seems a) they finished, and b) it does.

bwiernik:
Emiliano has done a lot of work on date parsing, so I think we can draw on that.

bdarcus:
> What level of EDTF had that?

I think/hope we get by with base + parts of level 1.

denismaier:
EDTF: yes please! Would be a massive simplification. I don't see what could speak against this...

denismaier:
This could also be useful for the proposed date range condition: https://github.com/citation-style-language/schema/issues/251

njbart:
I’m all for using the EDTF date format, but there’s one thing I am aware of that *can* be expressed in the current CSL JSON and CSL YAML date format, but, regrettably, has *not* been included in [EDTF](https://www.loc.gov/standards/datetime/edtf.html), and that’s season ranges, e.g., `2020-21/2020-22` for “Spring 2020 to Summer 2020”, something that appears not too infrequently in bibliographic data, and is recognised by style guides such as CMOS.

[edtf.js](https://github.com/inukshuk/edtf.js) does support these (“Seasons in intervals are supported at the experimental/non-standard level 3.“), and I feel CSL should declare support for EDTF dates *extended by season ranges* in similar terms. See also [this discussion](https://github.com/inukshuk/edtf.js/issues/12) on the edtf.js issue tracker.

biblatex, BTW, supports the `2020-21/2020-22` notation, too.

bdarcus:
> [edtf.js](https://github.com/inukshuk/edtf.js) does support these (“Seasons in intervals are supported at the experimental/non-standard level 3.“), and I feel CSL should declare support for EDTF dates _extended by season ranges_ in similar terms. See also [this discussion](https://github.com/inukshuk/edtf.js/issues/12) on the edtf.js issue tracker.

Thanks much @njbart; very helpful!

I agree, this is what we should do.

I revised the first post to include a TODO, of which this is is part of the first one.

schemas/input/csl-data.json
```diff
@@ -472,24 +477,8 @@
...
-                "maxItems": 3
-              },
-              "minItems": 1,
-              "maxItems": 2
-            },
-            "season": {
-              "type": ["string", "number"]
-            },
-            "circa": {
-              "type": ["string", "number", "boolean"]
```
dhimmel, line 492: So this PR will remove `date-parts` entirely? Since this is a backwards incompatible change, it should be part of a major not minor release.

Another option is to note that `date-parts` is deprecated and will be removed in the next major release. This would mean for the rest of the 1.x releases, `date-parts` is still an options as well as `edtf`.

bdarcus: Yeah, whatever you all think is fine. 

I'd particularly like to hear from people maintaining existing code: @fbennett, etc.

I partly removed it because I don't really understand the nuances of json schema like I do with rnc. I struggle enough with the basics without adding more complexity.

But we can revise it to allow that if there's consensus to that end.

retorquere:
> Emiliano has done a lot of work on date parsing, so I think we can draw on that.

My date parser works in stages (one of which is EDTF, done by edtf.js), but that's mostly because I try to make sense of dates that I should hope will never turn up in CSL:

1. I regex against a number of patterns to parse textual dates in various languages
2. I regex against a number of numerically formatted dates, taking into account the Zotero locale when there's ambiguity. These dates can have approx and seasons, but not both
3. when none of those match, I try EDTF, "upgrading" the input to the EDTF that edtf.js supports (replacing `unknown` by `*` etc)
4. when none of that matches, try to run it through `edtfy` and try `edtf.js` again
5. when none of those match, try to split on some common delimiters (`/`, `--`, etc) and try step 1-4 again to create a range; one of the sides may be empty to have an open-ended range.

You'll understand this is slow as molasses, but it makes sense of a lot of stuff I've been offered in test cases over the years.

schemas/input/csl-data.json
bwiernik: I think it is reasonable to maintain support for each of these formats, so long as we establish an order of priority:
1. edtf
2. date-parts
3. pandoc-style individual elements
4. raw
5. literal

bdarcus: The technical issue I was struggling with, here and on the discussion about adding relations, is how I can say that a property's value may either be a string, or an object/array. I wasn't even clear it was possible.

But this works as expected, so it it.

``` json
{
  "description": " test schema validating choice of string or object",
  "type": "object",
  "anyOf": [
    {
      "properties": {
        "date": {
          "type": "string"
        }
      },
      "required": [
        "date"
      ]
    },
    {
      "properties": {
        "date": {
          "year": {
            "type": "number"
          }
        }
      },
      "additionalProperties": false
    }
  ]
}
```

So the question is, would this approach be acceptable, @dhimmel?

We'd still need more feedback if it's wise, but it seems the cleanest solution longer term.

bdarcus: > I think it is reasonable to maintain support for each of these formats, so long as we establish an order of priority:
> 
> 1. edtf
> 2. date-parts
> 3. pandoc-style individual elements
> 4. raw
> 5. literal

We still need to clarify if we need the latter going forward. It's be cleaner to just say one representation or the other, but deprecate the older longer term.

bwiernik: We already support both strings (literal, raw) and arrays (date parts), so I think validation is not an issue?

bdarcus: It is if we want this to be valid:

``` yaml
issued: 2018-03-02
```

dhimmel: Having date-variables like `issued` support a string or an object is doable using [`anyOf`](https://json-schema.org/understanding-json-schema/reference/combining.html) as @bdarcus notes above.

There is the unresolved syntax question of whether we should support the following options:

```yaml
- id: my-reference
  type: report
  # option 1: flat
  issued: 2020-06-29
  # option 2: nested
  issued:
    edtf: 2020-06-29
    date-parts: [[2020, 6, 29]]
```

I think the best longterm option is 1, iff we are 100% sure EDTF is the best date-serialization format that will ever exist for bibliographic records and will suit this project's needs forever.

Supporting option 2 has the benefit that you can create CSL JSON that old and new citeprocs can handle. It allows for more backwards compatibility.

We could support option 1 and option 2. This is what I'm currently leaning towards

> We still need to clarify if we need the latter going forward. It'd be cleaner to just say one representation or the other, but deprecate the older longer term.

I think we should note that edtf format is "preferred" immediately. In the longterm, we should consider deprecating the older formats if we feel there is no longer any use for them. Removing the alternative formats helps simplify the burden on citeproc implementations, so its something to keep in mind for future MAJOR releases.

bwiernik: Maintaining backwards compatibility is important. There is a lot of embedded CSL-JSON out there that needs to remain valid. I think we should add edtf support now and specify the parsing order of priority as I gave above (e.g., in @dhimmel's second example, the date-parts will be ignored entirely in favor of the edtf).

bdarcus: > We could support option 1 and option 2. This is what I'm currently leaning towards.

Me too.

It could end up that 2 is more common in the json, and 1 more common in the yaml. And I think that's fine, given their different purposes.

bwiernik: So the processor instructions for (1) will be to treat this as a `raw` string that should be parsed first for edtf, then treated as literal if that fails, maybe following a process similar to @retorquere's below?

bdarcus: If this is the direction we want to go, I'll revise the PR to at least add back the earlier code, and do what I can on the json schema details.

bwiernik: So is the proposal then:
1. date-variables can be either a string or an object
1. if it's an object, it can have one or more of these named elements:
  1. elements
    1. edtf
    1. date-parts
    1. {pandoc-style individual elements}
       1. year
       1. month
       1. season
       1. day
    1. raw
    1. literal
    1. circa
  1. Elements are used in the listed order of priority (e.g., date-parts is ignored if edtf is present; literal is ignored if raw is present)
    1. circa is an exception. It is a supplement to the other non-edtf elements elements indicating an uncertain date)
  1. if `raw` is supplied, it is parsed as edft (and for other date formats?) following a procedure like Emiliano's
1. If it's a string, it is treated as `raw` and parsed the same way
1. Using EDTF-formatted dates entered in either string format or in the edtf element of an object is strongly preferred.

dhimmel: > It could end up that 2 is more common in the json, and 1 more common in the yaml. And I think that's fine, given their different priorities.

it seems like there is a large preference for avoiding nested values, so for consistency with this design principle, I like support option 1 (in addition to 2 for compat).

> So the processor instructions for (1) will be to treat this as a raw string that should be parsed first for edtf, then treated as literal if that fails, maybe following a process similar to @retorquere's below?

I like the idea of try edtf, and fallback to literal if parsing fails. I don't think it's a good idea to add all the additional fallbacks that @retorquere's implementation currently performs. Implementing all of these fallbacks in the same way across citeproc implementations in different languages would be a big burden. I think we want consistent behavior across citeproc implementations, which becomes unlikely if the spec gets too complex.

bdarcus:
Those details are TBD, but I'd rather not support the pandoc modeling here,
if we can avoid it.

I suggested EDTF as a better concise option for YAML.

It just adds more complexity all around to support three different
models/syntaxes.

But if @jgm objects, of course we can consider.

Part of the challenge is really just my limitations, and the limitations of
json schema.

With rnc, doing basically anything we want is easy and intuitive for me.

With json schema, it is not; at all.

So the more complex things get, the more I struggle to implement it.

But I guess that justifies narrative discussion before implementation.

We need input on this question from the people writing and maintaining the
CSL processors.

schemas/input/csl-data.json
bdarcus: > So is the proposal then ...

Those details are TBD, but I'd rather not support the pandoc modeling here,
if we can avoid it.

I suggested EDTF as a better concise option for YAML.

It just adds more complexity all around to support three different
models/syntaxes.

But if @jgm objects, of course we can consider.

Part of the challenge is really just my limitations, and the limitations of
json schema.

With rnc, doing basically anything we want is easy and intuitive for me.

With json schema, it is not; at all.

So the more complex things get, the more I struggle to implement it.

But I guess that justifies narrative discussion before implementation.

We need input on this question from the people writing and maintaining the
CSL processors.

bwiernik: re: pandoc parts, considering that only pandoc has currently adopted that syntax, so long as John is amenable to the edtf string structure, we could always not include it in the schema and pandoc could continue to support it as legacy off-spec behavior for compatibility.

bdarcus: Before I commit this @dhimmel, does this look right, aside from my notes in the comment?

https://gist.github.com/bdarcus/30807ba6466b3a089afc2b62be68bbc7#file-edtf-version-json-L403

bdarcus: Nevermind @dhimmel - just committed here.

bdarcus:
OK, all, after some back-and-forth with @bwiernik and @dhimmel, I revised the PR to basically give two choices:

1. current JSON representation
2. an EDTF string

I also added 2 to 1.

TBD still:

1. are we all OK with this in general?
2. language: right now, the language I put in the docstrings puts the two on equal footing; one might make the argument we should more strongly recommend EDTF over the current. if you have suggested tweaks to the language, please add as a line comment, with a "[suggestion](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request)."
2. what should the regular expression on the `edtf-datatype` definition I just added (currently just defined as a string)?

schemas/input/csl-data.json
```diff
@@ -468,8 +473,18 @@
         }
       ]
     },
+    "edtf-datatype": {
+      "title": "EDTF datatype pattern",
+      "description": "CSL input supports EDTF, validated against this regular expression.",
+      "type": "string"
```
bdarcus, line 479: The regular expression would get added as a `pattern` property here.

dhimmel: The regex to fully validate a EDTF might be pretty unwieldy. I think it's okay to not have a pattern or have one that checks only for major errors (like any invalid characters). But no strong opinion.

bdarcus:
> haven't tested the new JSON schema, but the changes look great @bdarcus (if they achieve the intended effect. Defining `edtf-datatype` separately was a smart move.

I did test it, and it works as expected.

schemas/input/csl-data.json
```diff
@@ -468,8 +473,19 @@
         }
       ]
     },
+    "edtf-datatype": {
+      "title": "EDTF datatype pattern",
+      "description": "CSL input supports EDTF, validated against this regular expression.",
+      "type": "string",
+      "pattern": "[0-9,\-,%,~,X,?,..,\/]"
```
bdarcus, line 480: This just validates the basic characters allowed. Would be nice to have some tighter.

schemas/input/csl-data.json
```diff
@@ -468,8 +473,19 @@
         }
       ]
     },
+    "edtf-datatype": {
+      "title": "EDTF datatype pattern",
+      "description": "CSL input supports EDTF, validated against this regular expression.",
+      "type": "string",
+      "pattern": "[0-9,-,%,~,X,?,..,\/]"
```
bdarcus, line 480: this only tests valid characters; would like tighter ideally

bdarcus:
I added details on the features we might support to the first post, with questions.

schemas/input/csl-data.json
```diff
@@ -468,8 +473,19 @@
...
+    },
     "date-variable": {
+      "title": "Date content model.",
+      "description": "The CSL input model supports two different date representations: an EDTF string, and a more structured equivalent.",
```
bdarcus, line 484:
```suggestion
      "description": "The CSL input model supports two different date representations: a preferred EDTF string, and a more structured equivalent.",
```

Possible modification, to prioritize EDTF.

schemas/input/csl-data.json
dhimmel: This regex doesn't look correct to me. `[0-9,-,%,~,X,?,..,\/]` only matches a single character (anything in the brackets).

You probably want something more like https://regex101.com/r/IZJFVM/1

If you describe what you want the regex to test I can help your with the syntax.

denismaier:
> 3\. long-term status of `date-parts`; do we say deprecated now and remove later, for example, or recommend EDTF?

I think we should remove `date-parts` eventually, unless it has features that are not covered by EDTF. Keeping both seems unnecessary burden to me.

schemas/input/csl-data.json
bdarcus: Thanks @dhimmel.

Your regex is good, but the final version will depend on answers to the compliance questions. For example, it doesn't currently support sets, but I'm not convinced it should.

I've updated the main post with examples, which I've used to test your mod. Here's [your's with the added test data](https://regex101.com/r/IZJFVM/2).

bdarcus:
```suggestion
      "pattern": "^[0-9\-~X?.\\\/]+$"
```

So, depending on any final tweaks, this?

njbart:
A few comments on the current version of the first post above:

- I’d ditch `date-parts` elements in the input file format in the long run; from a systematical point of view, they don’t offer any added value compared to EDTF.
- I’d also possibly ditch the `raw` element.
- The `literal` and `season` elements are useful for entering information that does not fit into the EDTF format, e.g. `"literal" : "13th century"` or `"season" : "Trinity"` (both examples from the old citeproc-js docs, [archived here](https://blog.eric.van-der-vlist.com/wp-content/var/owark/gsl-nagoya-u.net/http/pub/citeproc-doc.html/2011-01-21T07:20:02/citeproc-doc.html#id86); without `date-parts` elements, the latter will only make sense though if accompanied by a year in the edtf element). I feel it could be useful to keep the option of entering unstructured information – though neither CMOS nor APA seem to mention anything similar.
- If EDTF becomes the only structured input format, I guess we wouldn’t really need to keep the `circa` element either.
- EDTF details:
	- Level 0
		- Time stamps may be useful for blogs and the like (see CMOS 17e 14.209).
	- Level 1
		- CMOS 17e (14.145) and APA 7e (9.14) do not seem to distinguish between "uncertain" and "approximate". I’d still accept both in EDTF input.
	- Level 3
		- As discussed earlier, I’d add a statement that CSL is supporting one "nonstandard"/"experimental"/"Level 3" feature: season ranges such as `2020-21/2020-22`. (Just like edtf.js.)

bdarcus:
> * The `literal` and `season` elements are useful for entering information that does not fit into the EDTF format, e.g. `"literal" : "13th century"` or `"season" : "Trinity"` (both examples from the old citeproc-js docs, [archived here](https://blog.eric.van-der-vlist.com/wp-content/var/owark/gsl-nagoya-u.net/http/pub/citeproc-doc.html/2011-01-21T07:20:02/citeproc-doc.html#id86); without `date-parts` elements, the latter will only make sense though if accompanied by a year in the edtf element). I feel it could be useful to keep the option of entering unstructured information – though neither CMOS nor APA seem to mention anything similar.

OK. To be clear, the current code allows one representation, or the other.

But you could indeed do this:

``` yaml
issued:
   edtf: 1950
   season: Trinity
```

On the longer term question, beyond status of date-parts, would be whether the above would be allowed too; e.g. whether it'd just be the EDTF string.

PS - wouldn't this be an EDTF representation of 13th century: `12XX`?

njbart:
> We need these "Extended Interval " features on styles, but on input?

We do, e.g., for multivolume works where not all planned volumes have appeared yet (e.g., `1999/` or `1999/..`). (Strictly speaking, in this case I guess "unknown" would apply, since we expect some final volume, we just don’t know its publication date yet. For journals OTOH, "open" would seem more appropriate. For our purposes, I think the difference is moot.) Open/unknown start dates seem to be less common in bibliographic contexts, but, FWIW, biblatex supports all four forms, and does not distinguish open and unknown.

njbart:
> PS - wouldn't this be an EDTF representation of 13th century: 12XX?

Yes, and newer biblatex versions do support this as EDTF input – but I doubt whether even biblatex can actually render this as “13th century” (haven’t tested though).

bdarcus:
> > We need these "Extended Interval " features on styles, but on input?
> 
> We do, e.g., for multivolume works where not all planned volumes have appeared yet (e.g., `1999/` or `1999/..`). (Strictly speaking, in this case I guess "unknown" would apply, since we expect some final volume, we just don’t know its publication date yet. For journals OTOH, "open" would seem more appropriate. For our purposes, I think the difference is moot.) Open/unknown start dates seem to be less common in bibliographic contexts, but, FWIW, biblatex supports all four forms, and does not distinguish open and unknown.

I think at this point, I just want [the regex](https://github.com/citation-style-language/schema/pull/284#discussion_r447628624) to reject dates that have characters for features we don't want/need to support.

The remaining questions are sets (I agree, we can add), and approximate (`?`) vs uncertain (`~`).

bdarcus:
Maybe we should just do as biblatex does?

See it's [manual](http://mirrors.ibiblio.org/CTAN/macros/latex/contrib/biblatex/doc/biblatex.pdf): section: 2.3.8 Date and Time Specifications.

FWIW, it supports both approximate (circa) and uncertain dates

njbart:
> Maybe we should just do as biblatex does?

I'm all in favour. 

CSL could still support literal and season elements (biblatex recommends its `year` field if literal/unstructured date information is to be used, and suggests putting season(-ish) terms such as “Michaelmas term” in the `issue` field).

bdarcus:
> > We need these "Extended Interval " features on styles, but on input?
> 
> We do, e.g., for multivolume works where not all planned volumes have appeared yet (e.g., `1999/` or `1999/..`). 

What about this though: `{2010-03,2010-05,2010-03..2010-05}`?

njbart:
> What about this though: `{2010-03,2010-05,2010-03..2010-05}`?

Probably not. I can’t remember having encountered anything like this either in the wild, nor in CMOS or APA. Theoretically, such a feature might make sense if you wanted to highlight a particularly unusual publication history (“Doe (1901, 1902, 1910, 1969), My Autobiography in Four Volumes.“), but, again, I’ve never actually seen this anywhere.

bdarcus:
OK. I'm going to remove support for the brackets and comma then. We can revisit later.

fbennett:
Just one question, I guess. Will the `edtf.js` library be able to parse all valid CSL inputs on the `edtf` segment?

schemas/input/csl-data.json
bdarcus: > If you describe what you want the regex to test I can help your with the syntax.

@dhimmel - I think we want to exclude the set feature; here's an example: `{2010-03,2010-05,2010-03..2010-05}`.

So I guess that would mean not accepting commas or brackets. I'm unclear on why the above does accept them (at least in my testing).

njbart:
> FWIW, it supports both approximate (circa) and uncertain dates. Maybe we support both in this PR, and then finalize the decision on the documentation?

I’d support all three “flags” in EDTF input data, i.e., `~`, `?`, and `%`, regardless of whether CSL treats all as “circa” or starts making any distinctions.

bdarcus:
> Will the `edtf.js` library be able to parse all valid CSL inputs on the `edtf` segment?

`EDTF.js` claims to support the full EDTF spec.

Except, when I look at that page, there's this (what seems odd) note:

> Symbols for unknown and open dates in intervals have been switched: * makes more sense to represent an open date because it is often used as a wildcard to match "all" or "everything". 

Why the non-standard syntax there @inukshuk? Does this really mean, as it implies, that `../1900` is not valid for your parser?

To go back to your question, @fbennett, EDTF does not have support for literal or raw content in the current date representation. But it's a superset of date-parts + circa + season.

schemas/input/csl-data.json
bdarcus: Actually, I think the validator I'm testing with isn't using the regex. So I think what I have now should work.

schemas/input/csl-data.json
bwiernik: I would suggest:
`…an EDTF string (preferred)…`

schemas/input/csl-data.json
```diff
@@ -468,8 +473,19 @@
...
     },
+    "edtf-datatype": {
+      "title": "EDTF datatype pattern",
+      "description": "CSL input supports EDTF, validated against this regular expression.",
+      "type": "string",
+      "pattern": "^[0-9-%~X?.\/]{4,}$"
+    },
     "date-variable": {
+      "title": "Date content model.",
+      "description": "The CSL input model supports two different date representations: a preferred EDTF string, and a more structured alternative.",
```
bdarcus, line 484:
```suggestion
      "description": "The CSL input model supports two different date representations: an EDTF string (preferred), and a more structured alternative.",
```

bwiernik:
From the above, I’m not sure where we stand on open ended date ranges like `2006/..`. I think we should support them—e.g., with the new `periodical` type, such a format might be used to refer to the journal under its current editor tenure.

denismaier:
> From the above, I’m not sure where we stand on open ended date ranges like `2006/..`. I think we should support them—e.g., with the new `periodical` type, such a format might be used to refer to the journal under its current editor tenure.

Yes, open ended ranges are needed for journals or multi-volume works that are not complete yet.

bdarcus:
> From the above, I’m not sure where we stand on open ended date ranges like `2006/..`. I think we should support them—e.g., with the new `periodical` type, such a format might be used to refer to the journal under its current editor tenure.

Nothing here would preclude that. Just a topic for documentation.

I think this PR is RTM.

Would it be possible for @bwiernik, @denismaier, and @njbart to collaborate on a documentation PR sometime in the near future? As I suggested, I think biblatex would be the place to start, given a) it's well-designed in general, and b) they already integrated and implemented this support.

bdarcus:
> I think biblatex (which claims support for level 0 and 1) would be the place to start, given a) it's well-designed in general, and b) they already integrated and implemented this support.

PS - the other advantage of following biblatex as closely as possible is compatibility.

larsgw:
I missed this discussion, but I am personally not a fan of serializing information that could be represented differently. Converting CSL-JSON *with* EDTF to BibTeX or RIS now requires parsing EDTF, which is potentially pretty slow, versus an array access. I'd rather improve the current format, I think something like below could encode at least the same as EDTF.

```js
{
  accessed: [{ date: [2020, 7, 1] }],
  issued: [{ date: [1960], season: 'Spring' }],
  'original-date': [{ literal: '13th century' }],
  'event-date': [
    { date: [1960, 3, 2] },
    { date: [1960, 3, 5] }
  ]
}
```

(You could even remove the outer array for non-range dates)

However, I do see the use with YAML and I get it if this is considered a very minor concern.

bdarcus:
Except biblatex does support EDTF, so no parsing needed there.

It could be we enhance the structured option so it supports the exact same
features that we support in EDTF, for cases where that's a better option.

larsgw:
I meant specifically BibTeX, not BibLaTeX, which has `year` and `month` fields instead. I think it would be good to enhance the structured option.

bwiernik:
That’s basically adding the pandoc fields as an option, but providing for all of those options would mean requiring processors to support every format. 

Personally I think BibLaTeX is a more important thing to be compatible with; supporting both requires EDTF anyway.

Issue #285, opened 2020-06-29
New condition: Test for equality


denismaier:
**Is your feature request related to a problem? Please describe.**
Some styles require special formatting if certain variables are equal. E.g., APA style wants `publisher` to be suppressed when `publisher` and `author` are the same. Other styles suppress `container-author` when `container-author` and `author` are the same. Currently, there's no way to get this result as it's not possible to test if two variables are equal. 

**Describe the solution you'd like**
We should implement a new condition that tests true if two or variables are equal. Something like:

```
<if equal="author container-author">
</if>
```

Or maybe:

```
<if variable="container-author" equals="author">
</if>
```

The second example would require a more serious revision of the conditions syntax: `@variable` would specify the variable to be tested, the second attribute would specify the test. A single `@variable` would simply return true if a variable is present (the current behaviour). I'll open a seperate issue for such a revision.

**Additional context**
supersedes https://github.com/citation-style-language/csl-evolution/issues/27

** Additional requirements as per @bwiernik 

One wrinkle is how to handle equality for name variables, of which there may be many. Let's say that we have following data:

```
author: American Psychological Association
author: National Center for Measurement in Education
author: Society for Industrial-Organizational Psychology
publisher: American Psychological Association
```

Should 
```
<if variable="author" equals="publisher">
</if>
```

return TRUE or FALSE here? I'm not sure what, for example, what APA would prefer here. In the case of both a persona and corporate author, I suspect they would say TRUE:

```
author: Smith || John
author: American Psychological Association
publisher: American Psychological Association
```

If we decide that it should return FALSE, then that makes this a problem:
```
author: American Psychological Association
author: National Center for Measurement in Education
author: Society for Industrial-Organizational Psychology
publisher: American Psychological Association; National Center for Measurement in Education; Society for Industrial-Organizational Psychology
```

So, we would need `publisher` to behave like a `names` variable and accept multiple values. That is probably a good idea anyway.

_Originally posted by @bwiernik in https://github.com/citation-style-language/csl-evolution/issues/27#issuecomment-617259512_

denismaier:
Concerning @bwiernik's question above: Perhaps this warrants for an additional test like: 

```
variable="publisher" contained-in="author"
variable="publisher" part-of="author"
```

Or: 
```
variable="author" contains="publisher"
```

That would be less strict than a test for equality: These would return true if the publisher element is also part of the `author`. But testing for strict equality would return false.

bwiernik:
[...]

Issue #254, opened 2020-06-16
Test for date parts


bwiernik:
**Is your feature request related to a problem? Please describe.**
Several styles call for different date formatting depending on whether a date is year-only or has various more specific parts.

**Describe the solution you'd like**
CSLm has a solution by adding three condition attributes for `if` that test how specific a date variable is (see [here](https://citeproc-js.readthedocs.io/en/latest/csl-m/index.html#has-day-extension)):
  - `has-year-only`
  - `has-to-month-or-season` (has month or season, but no day)
  - `has-day`

I think we should adopt these attributes.

**Additional context**
Replaces https://github.com/citation-style-language/schema/issues/32#issuecomment-644727755

bdarcus:
I agree.

denismaier:
I agree as well.
But I don't understand the name of  `has-to-month-or-season` (why has-to?)

bdarcus:
Makes no sense, so change it?

bwiernik:
It makes sense because `has-month` would logically be `TRUE` for the date "March 20, 2005". The meaning of this attribute is "has month, but no day". Perhaps `has-month-and-year-only` or `has-month-year-only`?

bwiernik:
(So, the "has-to" is "has information up to [until] month or season")

bdarcus:
What about running it around as "missing-"?

denismaier:
Or we check for `if date-parts=year month day`?

bdarcus:
> Or we check for `if date-parts=year month day`?

That seems ideal; no? Just have the content be a space-separated list?

Anything we're missing with that?

bwiernik:
Okay, let's go with:
  - `has-day`
  - `missing-day`
  - `year-only`

> space-separated list: Anything we're missing with that?

Two issues. First, how do you specify which date variable is being tested? Second, it makes for awkward tests I think because the test is usually for what's missing, so it would have be to have `match="none"`, which would make it hard to combine with other conditions.

denismaier:
Ok with me.

Regarding:

> so it would have be to have `match="none"`, which would make it hard to combine with other conditions.

I propose to add add a easier condition syntax anyway.

denismaier:
Just a question after reading your suggestions in the linked [issue](https://github.com/citation-style-language/schema/issues/251): Could we find a solution that aligns more with the syntax there, namely:

```
date-range-variable='issued' date-range-start='1900' date-range-end='2000'
[...]

PR #297, opened 2020-07-01
Add new conditions

### LINKED ISSUE #251: add condition: year < 2000

denismaier:
## Description

WIP
This will add new test condîtions that require two attributes:

- `@tested`: specifies the variables to be tested
- `@matches`: specifies the variable to be tested against
- `@date-range`: tests if a variable is in a given date range
- `@date-precision`: tests whether certain date parts are present.

Fixes #251 #254  #285 

## Questions

The patterns lead to duplicate suggestions when using autocompletion. Anyone knows what to do here?

## Todo
- [X] Currently, the date test requires `date-precision` **and** `date-range`. How can one of them be made optional?
- [X] How can we combine the other tests with the new complex tests? Currently they can't be used together.

## Type of change

## Checklist

### DIFF
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -11,7 +11,7 @@ div {
   
   ## When specifying conditions, we allow either the attribute syntax or the element-based one.
   choose.condition =
-    (condition.atts+, match?)
+    (condition.atts, match?)
     | element cs:conditions { match, condition.elem+ }
   choose.if = element cs:if { choose.condition, rendering-element* }
   choose.else-if =
@@ -21,7 +21,8 @@ div {
   ## This element allows for more complex conditional logic; for 
   ## example, if you need to specify mixed "match" rules.
   condition.elem = element cs:condition { match?, condition.atts }
-  condition.atts =
+  condition.atts = ((condition.atts.simple+, condition.atts.complex) | condition.atts.simple+ | condition.atts.complex)
+  condition.atts.simple =
     
     ## If used, the element content is only rendered if it disambiguates two
     ## otherwise identical citations. This attempt at disambiguation is only
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -12,7 +12,7 @@ div {
   ## When specifying conditions, we allow either the attribute syntax or the element-based one.
   choose.condition =
     (condition.atts, match?)
-    | element cs:conditions { match, condition.elem+ }
+    | element cs:conditions { match?, condition.elem+ }
   choose.if = element cs:if { choose.condition, rendering-element* }
   choose.else-if =
     element cs:else-if { choose.condition, rendering-element* }
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -64,6 +64,17 @@ div {
       attribute variable {
         list { variables+ }
       }
+      |
+      tests-dates
+  
+  # Tests for date variables that require two attributes:
+  tests-dates =
+    attribute tested { 
+      list { variables.dates+ }
+     },
+    attribute date-precision { text },
+    attribute date-range { text }
+
   match =
     
     ## Set the testing logic.
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -64,13 +64,13 @@ div {
       attribute variable {
         list { variables+ }
       }
-      |
-      condition.atts.complex.dates
-      |
-      condition.atts.complex.general
-
   
-  # Tests that require two attributes 
+  # Tests that require two attributes
+  condition.atts.complex.dates = 
+    condition.atts.complex.dates
+    |
+    condition.atts.complex.general
+
   # Dates
   condition.atts.complex.dates =
     ## specifies the variable to be tested
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -81,8 +81,11 @@ div {
      (date-precision | date-range)+
     
   ## Tests the precision of a date variable
-  date-precision = attribute date-precision { 
-    list { ("year" | "month" | "day") }
+  date-precision = attribute date-precision {
+    ## specifies the highlest level of precision of a date variable
+    ## e.g., date-precision="month" means date-parts "year" and "month"
+    ## are present, but "day" is missing.
+    "year" | "month" | "day"
    }
   
   ## Tests whether a date variable is within a given range
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -11,7 +11,7 @@ div {
   
   ## When specifying conditions, we allow either the attribute syntax or the element-based one.
   choose.condition =
-    (condition.atts, match?)
+    (condition.atts+, match?)
     | element cs:conditions { match, condition.elem+ }
   choose.if = element cs:if { choose.condition, rendering-element* }
   choose.else-if =
@@ -66,7 +66,7 @@ div {
       }
   
   # Tests that require two attributes
-  condition.atts.complex.dates = 
+  condition.atts.complex = 
     condition.atts.complex.dates
     |
     condition.atts.complex.general
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -21,7 +21,7 @@ div {
   ## This element allows for more complex conditional logic; for 
   ## example, if you need to specify mixed "match" rules.
   condition.elem = element cs:condition { match?, condition.atts }
-  condition.atts = ((condition.atts.simple+, condition.atts.complex) | condition.atts.simple+ | condition.atts.complex)
+  condition.atts = ((condition.atts.simple+, condition.atts.complex?) | condition.atts.complex)
   condition.atts.simple =
     
     ## If used, the element content is only rendered if it disambiguates two
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -70,23 +70,26 @@ div {
       condition.atts.complex.general
 
   
-  # Tests that require two attributes
+  # Tests that require two attributes 
   # Dates
   condition.atts.complex.dates =
     ## specifies the variable to be tested
     attribute tested { 
       list { variables.dates+ }
      },
-    ## Tests the precision of a date variable
-    attribute date-precision { text },
-    ## Tests whether a date variable is within a given range
-    attribute date-range { text }
+     (date-precision | date-range)+
+    
+  ## Tests the precision of a date variable
+  date-precision = attribute date-precision { text }
+  
+  ## Tests whether a date variable is within a given range
+  date-range = attribute date-range { text }
   
   # General
   condition.atts.complex.general =
     ## specifies the variable to be tested
     attribute tested { 
-      list {variables+ }
+      list { variables+ }
      },
     ## Tests whether the tested variable matches another variable
     attribute matches { 
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -65,15 +65,33 @@ div {
         list { variables+ }
       }
       |
-      tests-dates
+      condition.atts.complex.dates
+      |
+      condition.atts.complex.general
+
   
-  # Tests for date variables that require two attributes:
-  tests-dates =
+  # Tests that require two attributes
+  # Dates
+  condition.atts.complex.dates =
+    ## specifies the variable to be tested
     attribute tested { 
       list { variables.dates+ }
      },
+    ## Tests the precision of a date variable
     attribute date-precision { text },
+    ## Tests whether a date variable is within a given range
     attribute date-range { text }
+  
+  # General
+  condition.atts.complex.general =
+    ## specifies the variable to be tested
+    attribute tested { 
+      list {variables+ }
+     },
+    ## Tests whether the tested variable matches another variable
+    attribute matches { 
+      list {variables+ }
+     }
 
   match =
     
--- a/schemas/styles/csl-choose.rnc
+++ b/schemas/styles/csl-choose.rnc
@@ -80,7 +80,9 @@ div {
      (date-precision | date-range)+
     
   ## Tests the precision of a date variable
-  date-precision = attribute date-precision { text }
+  date-precision = attribute date-precision { 
+    list { ("year" | "month" | "day") }
+   }
   
   ## Tests whether a date variable is within a given range
   date-range = attribute date-range { text }


denismaier:
Feedback welcome...

denismaier:
Is it possible that this currently does not allow multiple conditions?

bwiernik:
DOI isn’t a number variable, I think that would be the issue?

denismaier:
> DOI isn’t a number variable, I think that would be the issue?

No, I don't think that's the issue here. I'm getting this message: 
```
attribute "is-numeric" not allowed here; expected attribute "match"
```

I think, when working on the new condition elements @bdarcus has changed the pattern so it allows only one testing attribute and `@match`:

```
  choose.condition =
    (condition.atts, match?)
    | element cs:conditions { match, condition.elem+ }
```

On master we have currently this:

```
 choose.if = element cs:if { condition+, match, rendering-element* }
```

I don't know if that was on purpose.

Shall I change the pattern, so multiple tests will be possible again? =>

```
  choose.condition =
    (condition.atts+, match?)
    | element cs:conditions { match, condition.elem+ }
```

bdarcus:
> Shall I change the pattern, so multiple tests will be possible again?

We never really did fully test this feature before I merged it.

Could you please do that, with Atom, and do what seems sensible? 

I have been worried about allowing match in too many places, but I don't remember if I consciously did that or not ATM; it's all a blur!

bdarcus:
But I do think the answer to your question is yes; was probably just a mistake.

denismaier:
Question: the current solution allows specific tests for `@tested` with number variables, and other tests for `@tested` with other variables. It works, but it's not particularly pretty. In autocompletion it shows "tested" twice. And variable suggestions differn depending on which "tested" I choose. Is there a better way?

bdarcus:
> In autocompletion it shows "tested" twice. And variable suggestions differ depending on which "tested" I choose. Is there a better way?

Not following here. Could you post a screenshot, and explain what you'd prefer?

denismaier:
https://imgur.com/a/6MwVuTl 
this shows "tested" appears twice"
Forget about the second part. I can't reproduce this.

schemas/styles/csl-choose.rnc
```diff
@@ -64,6 +64,38 @@ div {
...
+  date-precision = attribute date-precision { text }
+  
+  ## Tests whether a date variable is within a given range
+  date-range = attribute date-range { text }
+  
+  # General
+  condition.atts.complex.general =
+    ## specifies the variable to be tested
+    attribute tested { 
+      list { variables+ }
```
bdarcus, line 92: This includes all variables, including dates. Is that what you want?

denismaier:
Another problem:  this change does not work:


 ```
choose.condition =
   (condition.atts+, match?)
 | element cs:conditions { match, condition.elem+ }
```

The + is not allowed in the attributes here. I don't really see if that can be solved. Suggestions?

**Comment by bdarcus** [comment]
> The + is not allowed in the attributes here.

Maybe because you can't have more than one of the same attribute on an element?

So control those details on the `condition.atts` pattern?

**Comment by denismaier** [comment]
> > The + is not allowed in the attributes here.
> 
> Maybe because you can't have more than one of the same attribute on an element?
> 
> So control those details on the `condition.atts` pattern?

This here worked:

```
choose.if = element cs:if { condition+, match, rendering-element* }
```

I don't really see where the change happened that causes this.

**Comment by bdarcus** [comment]
Maybe the parenthesis you have? That's a difference between the two versions.

**Comment by denismaier** [comment]
> Maybe the parenthesis you have? That's a difference between the two versions.

Yes, it must have to do with the grouping. I'm getting this error message:

> Could not process schema or catalog: a group of attributes must not be repeatable (section 7.1 of the RELAX NG specification requires that the simplified XML form of the schema not contain any elements matching the path oneOrMore//group//attribute)

After a couple of attempts it's always this again.

File: schemas/styles/csl-choose.rnc [hunk 2]
```diff
@@ -64,6 +64,38 @@ div {
...
+
+  
+  # Tests that require two attributes 
+  # Dates
+  condition.atts.complex.dates =
+    ## specifies the variable to be tested
+    attribute tested { 
+      list { variables.dates+ }
+     },
+     (date-precision | date-range)+
```
bdarcus, line 80: > Currently, the date test requires date-precision and date-range. How can one of them be made optional?

```suggestion
     (date-precision, date-range?)
```

... requires precision, leaves range optional.

denismaier: Oh I forgot to check the box. I think `(date-precision | date-range)+` already means that at least one of them is needed. The other is optional. I don't think we should prescribe which one is required

schemas/styles/csl-choose.rnc
```diff
@@ -64,6 +64,38 @@ div {
...
+    
+  ## Tests the precision of a date variable
+  date-precision = attribute date-precision { text }
```
bdarcus, line 83: why is the value text?

denismaier: because I still have to add the list... ;-) Getting multiple conditions to work got in the way.

denismaier: Should be ok now.

bdarcus:
> this shows "tested" appears twice"

is this now resolved?

denismaier:
> > this shows "tested" appears twice"
> 
> is this now resolved?

No, it's even worse now. I have just pushed a commit that should now allow multiple tests and complex tests together. But this populates the autocompletion with tons of duplicate entries.

denismaier:
> > Maybe the parenthesis you have? That's a difference between the two versions.
> 
> Yes, it must have to do with the grouping. I'm getting this error message:
> 
> > Could not process schema or catalog: a group of attributes must not be repeatable (section 7.1 of the RELAX NG specification requires that the simplified XML form of the schema not contain any elements matching the path oneOrMore//group//attribute)
> 
> After a couple of attempts it's always this again.

I have now found a solution that works. The errors were due to the newly introduced complex tests. It's not pretty and can perhaps be further simplified.

denismaier:
Should I remove `cs:conditions` now that I'm at it?

bdarcus:
> Should I remove `cs:conditions` now that I'm at it?

WDYT, now that you've tested it?

I'd say, if you have a strong preference, yes; if in doubt, keep it?

bdarcus:
I'm turning off draft to see if that allows me to pull the branch.

bdarcus:
Not sure why (might be my fault?), but currently `@form` is required on `cs:text`; need to make that optional of course.

bdarcus:
Can you please make this change? See position of the `?`?

``` diff
diff --git a/schemas/styles/csl.rnc b/schemas/styles/csl.rnc
index 6577d5c..581f268 100644
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -650,7 +650,7 @@ div {
         [ a:defaultValue = "long" ] attribute form { "short" | "long" })
        | (attribute variable { variables.titles },
           [ a:defaultValue = "long" ]
-          attribute form { "short" | "long" | "sub" | "main" })?)
+          attribute form { "short" | "long" | "sub" | "main" }?))
```

denismaier:
> Can you please make this change? See position of the `?`?
> 
> Actually, not sure this is right. Checking.
> 
> ```diff
> diff --git a/schemas/styles/csl.rnc b/schemas/styles/csl.rnc
> index 6577d5c..581f268 100644
> --- a/schemas/styles/csl.rnc
> +++ b/schemas/styles/csl.rnc
> @@ -650,7 +650,7 @@ div {
>          [ a:defaultValue = "long" ] attribute form { "short" | "long" })
>         | (attribute variable { variables.titles },
>            [ a:defaultValue = "long" ]
> -          attribute form { "short" | "long" | "sub" | "main" })?)
> +          attribute form { "short" | "long" | "sub" | "main" }?))
> ```

Either way, shouldn't we make this change on the 1.1 branch? What about form for other variables, shouldn't that be optional as well?

denismaier:
> Can you please make this change? See position of the `?`?
> 
> Actually, not sure this is right. Checking.
> 
> ```diff
> diff --git a/schemas/styles/csl.rnc b/schemas/styles/csl.rnc
> index 6577d5c..581f268 100644
> --- a/schemas/styles/csl.rnc
> +++ b/schemas/styles/csl.rnc
> @@ -650,7 +650,7 @@ div {
>          [ a:defaultValue = "long" ] attribute form { "short" | "long" })
>         | (attribute variable { variables.titles },
>            [ a:defaultValue = "long" ]
> -          attribute form { "short" | "long" | "sub" | "main" })?)
> +          attribute form { "short" | "long" | "sub" | "main" }?))
> ```

By the way, Also with this feature: a lot of duplicates in autocompletion.

bdarcus:
I'm looking at to this if there's a fix, of if maybe just something weird with the atom extension.

bdarcus:
But let's set aside two distinct things:

1. (and more important): does it behave as expected from a validation perspective; can you do things you want, and not do things you shouldn't?
2. the auto-complete presentation in Atom

denismaier:
I'll do some testing tomorrow.

Concerning the auto-completion: I guess this might come from those alternate patterns that partly consist of identical elements or attributes. (I'll check if oxygen can remove duplicates from the auto-completion suggestions.) 
Or, perhaps we could simplify the grammar and express the constraint schematron rules t

denismaier:
Hmm, first result with Oxygen:
Autocompletion works much better. No duplicate suggestions.

But:

Validation gives different errors than with Atom. How can that be? 
E.g. it does not accept the new `cs:conditions`. (Actually, it suggests the element in autocompletion, but throws an error once the element is there. 
Then, a very simple `<text variable="title" />` leads to this error message in Oxygen:
> attribute "plural" not allowed here; expected attribute "display", "font-style", "font-variant", "font-weight", "prefix", "quotes", "strip-periods", "suffix", "text-case", "text-decoration" or "vertical-align"

Atom complains:

> element "text" missing required attribute "form" [RNC]

The strange thing is both use Jing so how come there's such a difference?

bdarcus:
You might report that as a bug to oxygen?

Because it's non-sensical.

bdarcus:
And actually, one of us should report to the atom jing package, to see if
it's there bug, or our's.

I suspect it's there's.

bdarcus:
I just reported this to the the Atom jing package.

https://github.com/aerhard/linter-autocomplete-jing/issues/63

denismaier:
I've prepared a few tests:
https://gist.github.com/denismaier/a42d5f8bf65b92cfd289c9e4ca940ad3

Currently, things work as expected. (Autocompletion aside.)

denismaier:
> 
> 
> I've prepared a few tests:
> https://gist.github.com/denismaier/a42d5f8bf65b92cfd289c9e4ca940ad3
> 
> Currently, things work as expected. (Autocompletion aside.)

I may have missed some combinations. So please test as well.

bdarcus:
That's very useful @denismaier.

We could use those as the basis for adding to this branch of the test-suite.

https://github.com/citation-style-language/test-suite/tree/v1.1-experiments/processor-tests/humans/experiments

denismaier:
> We could use those as the basis for adding to this branch of the test-suite.

The valid tests, yes. (The others are only there to test if the schema is correct. That shouldn't make it into the processor test-suite, or should it?

bdarcus:
Correct.

schemas/styles/csl-choose.rnc
bdarcus: Bumping this. Looks like you should accept this change regardless of whether it fixes the auto-completion issue.

denismaier: Are you sure we should do this? Currenty the idea is to have tests that work for all variables, and some that only work for dates.

bwiernik:
@denismaier Is this ready for final review?

denismaier:
@bwiernik Des.

denismaier:
@cormacrelf Thoughts about this?

schemas/styles/csl-choose.rnc
bwiernik: Two options are to either (1) list the attributes under both general and date or (2) allow general to accept any variable type (including dates). Don't see much of a reason to prefer one over the other, so your call @bdarcus.

denismaier: Yeah currently it's version 2. But the other version is ok as well...

denismaier: So, do you have strong preference for one of the options @bdarcus ?

bdarcus: I guess I don't understand the issue. How does this statement square with the implementation?

> Currenty the idea is to have tests that work for all variables, and some that only work for dates.

denismaier: At the moment, we have one complex test that works for all variables: `if tested="X" matches="Y"`.
All the other complex tests are specific for date variables. If I accept your suggestion, I'll have to add `@matches` to the "complex tests for dates" section. Do you prefer that?

schemas/styles/csl-choose.rnc
bdarcus: Status of this comment?

denismaier: Should be ok now. (Just one question: Does it need to be the a list if you can only select one of the values?)

bdarcus: > Does it need to be the a list if you can only select one of the values?)

No; list refers to the value.

schemas/styles/csl-choose.rnc
```diff
@@ -64,6 +65,40 @@ div {
...
+  condition.atts.complex.dates =
+    ## specifies the variable to be tested
+    attribute tested { 
+      list { variables.dates+ }
+     },
+     (date-precision | date-range)+
+    
+  ## Tests the precision of a date variable
+  date-precision = attribute date-precision { 
+    list { ("year" | "month" | "day") }
```
bdarcus, line 85: This says this is valid:

``` xml
<if date-precision="year month"/>
```

denismaier: Thanks. I have changed this according to your suggestion, and I've also added a docstring.

bdarcus:
What about your "todo" items?

denismaier:
> 
> 
> What about your "todo" items?
> 
> I can take a look again if you want.

They should be ok by now (that's why they're checked off.) I did some testing with a minimal style and everything should work. But doing another test won't hurt either.

bwiernik:
Just to be clear, the resolutions are:

>  Currently, the date test requires date-precision and date-range. How can one of them be made optional?

One or the other is optional.

And 

> How can we combine the other tests with the new complex tests? Currently they can't be used together.

Simple and complex tests can be combined?

denismaier:
@bwiernik Exactly.

bwiernik:
Looks good to me.

bdarcus:
Just noting here that the jing autocomplete project fixed the issue of duplicate attributes. So the Atom support should handle this correctly now.

PR #299, opened 2020-07-02
adds locale option initial-family-name-delimiter

### LINKED ISSUE #236: Make space between initial and last name customizable
Specifically, French typography typically requires a non breaking space here, whereas US typography very much doesn't. 
I.e. `A.&nbsp;Smith` vs. `A. Smith`
The former is currently impossible in Zotero.
This should likely be a locale spec, along the lines of 
initial-last-name-delimiter=" "

denismaier:
## Description

per https://github.com/citation-style-language/schema/issues/236#issue-635743993:
> Specifically, French typography typically requires a non breaking space here, whereas US typography very much doesn't.
> I.e. `A.&nbsp;Smith` vs. `A. Smith`
> The former is currently impossible in Zotero.
> This should likely be a locale spec

This adds a new locale option `initial-family-name-delimiter`.

Closes #236 

## Type of change

## Checklist

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -231,6 +231,11 @@ div {
   locale.style-options =
     element cs:style-options {
       
+      ## Specify the space character to be used between initialized first names and family names, 
+      ## e.g. a non-breaking space.
+      [ a:defaultValue = " " ]
+      attribute initial-family-name-delimiter { text }?,
+      
       ## Limit the "ordinal" form to the first day of the month.
       [ a:defaultValue = "false" ]
       attribute limit-day-ordinals-to-day-1 { xsd:boolean }?,


denismaier:
> Why "initial"?

Because it deals with initials? (Or what do you mean?)

bdarcus:
No, I'm just wondering: why is this only specific to initials? Is it feasible some styles would require this on non-initialized names?

denismaier:
It's possible, yes. But I think it would be rather uncommon. Having a non-breaking space after an abbreviation and the following word (if they form a unit) is a rather common typographic rule on the other hand.

denismaier:
Hmm, even Chicago recommends non-breaking spaces between initials. (6.121)
This leads to the question: How should we deal with delimiters between multiple initials?

bwiernik:
That is handled by `initialize-with`

denismaier:
Why is this not linked to #236?

bwiernik:
No idea why the closing keyword aren't working.

bwiernik:
I don't really see styles specifying that names in general cannot break across lines, but styles/locales do have varying rules regarding line breaks after initials or abbreviations, so I think limiting the behavior to initialized names is good.

bdarcus:
Should the value really be "text"? Or should we itemize, with just the single item for now?

denismaier:
You mean
```
attribute initial-family-name-delimiter { "non-breaking-space" | "space" }?,
```

bdarcus:
Yes, with space default.

denismaier:
What do you think @bwiernik ? Should we make this change?

denismaier:
Just for the record: `initialize-with` is currently also defined as `text`

```
## Activate initializing of given names. The attribute value is appended
## to each initial (e.g., with ". ", "Orson Welles" becomes "O. Welles").
attribute initialize-with { text }?,
```

bdarcus:
> Just for the record: `initialize-with` is currently also defined as `text`
> 
> ```
> ## Activate initializing of given names. The attribute value is appended
> ## to each initial (e.g., with ". ", "Orson Welles" becomes "O. Welles").
> attribute initialize-with { text }?,
> ```

The problem I see with using text is twofold:

1. UX - how is it a style author to know how to enter a non-breaking space?
2. What is the content supposed to actually be? a unicode code point? If yes, see above.

bwiernik:
It should just be text. 

They should enter an HTML entity. This is why is done in hundreds of styles in the repository. It might be good to add a comment somewhere that non-ASCII characters should be entered as HTML entities.

bwiernik:
LGTM

bdarcus:
> It should just be text.
> 
> They should enter an HTML entity. This is why is done in hundreds of styles in the repository. It might be good to add a comment somewhere that non-ASCII characters should be entered as HTML entities.

Ugh.

> LGTM

Hang on.

We should not be using HTML entities for this in my view.

bwiernik:
Why not? CSL styles/locales are XML. It's natural for them to be HTML entities. 

Beyond that:
1. It's already done that way in hundreds of styles.
1. As you note, it's often easier for style writers to type HTML entities than extended unicode characters.
1. HTML entities don't have potential locale/encoding corruption issues that just typing Unicode characters might have.

bwiernik:
As to why it should just be text, why make a more complex system with an enumerated list that (1) processors would have to build a handler for and that (2) works differently from every other "delimiter" attribute in CSL, rather than just letting style and locale writers enter the text that should appear. 

This really is an already solved problem.

bwiernik:
See for example: https://github.com/citation-style-language/locales/blob/master/locales-fr-FR.xml

bdarcus:
> Why not? CSL styles/locales are XML. It's natural for them to be HTML entities.

Not really. HTML entities are a funky legacy of a time before XML and unicode.

> Beyond that:
> 
> 1. It's already done that way in hundreds of styles.
> 2. As you note, it's often easier for style writers to type HTML entities than extended unicode characters.
> 3. HTML entities don't have potential locale/encoding corruption issues that just typing Unicode characters might have.

1. we can correct this, so this is a non-issue
2. both suck though from a style authoring perspective
3. but the below doesn't either, and doesn't suck (though does have other limitations, but not clear to me that matters here)

``` 
initial-family-name-delimiter='non-breaking-space'
```

> As to why it should just be text, why make a more complex system with an enumerated list that (1) processors would have to build a handler for and that (2) works differently from every other "delimiter" attribute in CSL, rather than just letting style and locale writers enter the text that should appear.

You raise some good points here, but in the end, because CSL is an output-format agnostic format; HTML entities aren't helpful for formats that don't support them.

If we go this way, we should specify unicode, and can use a script to automate the translation on the styles repo.

bwiernik:
I absolutely oppose anything that isn't just typing text. Delimiters in CSL are entered as simple text--we shouldn't make that inconsistent. 

Every processor needs to be able to parse XML already, so they can decode entities if the output format requires it. 

There is also the problem that in some cases, using Unicode characters directly is _less clear_. For example, is a space a regular space or a non-breaking space? `&nbsp;` is unambiguous when editing. 

I think switching from HTML entities to Unicode would be more of problem than the existing status quo that is _already implemented by styles, locales, and every CSL processor_.

I spent months a few years ago trying to get locale-agnostic rendering of Unicode characters to work nicely with R. Unicode is very poorly supported on Windows even to this day. I fear that entering Unicode characters is going to lead to lots of confusion from infrequent editors who try to edit styles in programs like Notepad that don't have good encoding recognition behavior. 

Before we arbitrarily switch conventions, I'd want to hear from @adam3smith and @rmzelle since they are the ones who would likely have to deal with this.

bdarcus:
Let's also tag @PaulStanley here (sorry Paul) again, since I believe he's working on a new processor that targets an output format that does not support HTML entities.

bdarcus:
> I absolutely oppose anything that isn't just typing text. Delimiters in CSL are entered as simple text--we shouldn't make that inconsistent.

I'll just mention here, in retrospect, this was probably a mistake, per the earlier discussion about whitespace handling in XML.

bdarcus:
This what xmllint, arguably the most widely used xml parser, does this with an entity.

``` console
❯ cat t.xml 
<test text="&nbsp;"/>
                                                                                                                                 
❯ xmllint -format t.xml
t.xml:1: parser error : Entity 'nbsp' not defined
<test text="&nbsp;"/>
 ```

adam3smith:
Just reading quickly:
I dislike the lack of flexibility introduce by something like:
`initial-family-name-delimiter='non-breaking-space'`
What, for example, if someone wants a narrow non-breaking space here (we currently discourage that on the repository due to poor support in fonts, but that might change and it can already be used for custom styles). So I'd lean towards text.

I don't see why we'd support HTML escaping (i.e. `&nbsp;`) over XML escaping `&#160;`, which we currently use, which is legal XML and which works with all the linters (try e.g. with Bruce's example above). I'm guessing @bwiernik is actually referring to the latter because introducing random HTML into XML does indeed sound like a very bad idea.

On the repository, we escape a number of common unicode characters to escaped XML because, as Brenton says, they're easier to distinguish. That's definitely em-dash and non-breaking space, but there are likely others (the script running formatter.citationstyles.org/ does this automatically as does Rintze's python reindent/reorder script (afaik they're identical).

bdarcus:
> I don't see why we'd support HTML escaping (i.e. `&nbsp;`) over XML escaping `&#160;`

So if there are examples of `&nbsp;` in the styles repo, that's an oversight we should fix?

bdarcus:
I'll merge this then with `text`, but urge we explicitly say people should use unicode everywhere in CSL files in the spec, and NOT use HTML entities.

adam3smith:
> So if there are examples of &nbsp; in the styles repo, that's an oversight we should fix?

There isn't. Validation  fails with a &nsbp; ("undeclared entity"). Given that I think this is generally true for XML validation and not just an artifact of our validator, I'm not sure we need to specifically exclude HTML entities any more than we need to exclude unescaped ampersands -- this has never been a source of confusion. (I also don't have any strong objections to *including* it if there's any indication of confusion I'm not aware of.)

bwiernik:
I misspoke when I was remembering the entity name before.

bdarcus:
It is indeed a general rule; XML documents with undeclared entities are not
valid.

PR #301, opened 2020-07-06
input: Align date-parts and edtf string models

### LINKED ISSUE #300: input: harmonize input date model with EDTF
As part of recent addition of EDTF as an option on input date rendering, I'd like to propose we also update the object representation to be fully compatible with it.

See linked PR.

cc @dhimmel @njbart @jgm

bdarcus:
This modifies the structured date object to align with the EDTF model.

It moves the qualifiers to the date object, and then defines a date
range as an array of these date objects.

The result is that dates can either be represented as an EDTF string, 
or as a structured object date, or a date range array.

The intention is this structured variant will go away in time, so that 
in the future the only option will be the EDTF string.

Closes #300 

-----

In CSL JSON 1.0, dates are either a two-level nested array ...

``` js
  "issued": {
    "date-parts": [
      [
        2000,
        3,
        15
      ],
      [
        2000,
        3,
        17
      ]
    ],
    "circa": "true"
  }
```

... or a `raw` string, with is a standard EDTF-like date (note, however, that the example isn't compliant because months are not two-digits, as they are in ISO 8601).

``` js
"accessed": {
     "raw": "2005-4-12"
},
"issued": {
     "raw": "2000-3-15/2000-3-17"
}
```

In CSL JSON 1.1, we move `raw` to an explicit EDTF string option on the property itself,

So in this option, the above becomes:

``` js
"accessed": "2005-4-12",
"issued":  "2000-3-15/2000-3-17"
```

This PR defines:

1. a date as an object, and moves the qualifiers onto the object, to make it consistent with EDTF (where you can have different qualifiers on the begin and end of a range)
2. a date range as a a two-item array of date objects

A single date would be:

``` js
  "issued": {
      "date-parts": [
        2000,
        3,
        15
      ],
      "approximate": "true"
    }
```

... and the first range example becomes this ("approximate" is the EDTF language what we have called "circa"):

``` js
  "issued": [
    {
      "date-parts": [
        2000,
        3,
        15
      ]
    },
    {
      "date-parts": [
        2000,
        3,
        17
      ]
    }
  ]
```
Note: open question is how to handle the mismatch between circa on 1.0 and this.

And we can then do things like season ranges (not formally supported by EDTF, but easy for us to add):

``` js
  "issued": [
    {
      "date-parts": [
        2000,
        21
      ]
    },
    {
      "date-parts": [
        2000,
        22
      ]
    }
  ]
```

This would seem a better long-term that would allow us to painlessly drop the object/array representation at some point in the future.

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -445,7 +445,18 @@
     },
     "date-object": {
       "title": "Date Object",
-      "description": "An EDTF date represented as an object.",
+      "description": "An EDTF date represented as an object. Seasons can be represented using the 21-24 month values of EDTF, decades and centuries using the X notation, and the qualifiers with the included boolean properties",
+      "examples": [
+        {
+          "id": "one",
+          "type": "book",
+          "title": "The Title",
+          "issued": {
+            "date-parts": [2000, 3, 10]
+          },
+          "approximate": "true"
+        }
+      ],
       "type": "object",
       "properties": {
         "date-parts": {
@@ -478,6 +489,20 @@
       "title": "Date Range",
       "description": "An EDTF range is a two-item array. An open end or begin of a range can be represented with an empty (date) object.",
       "type": "array",
+      "examples": [
+        {
+          "id": "range-1",
+          "type": "book",
+          "title": "The Title",
+          "issued": [{ "date-parts": [2000] }, { "date-parts": [2001] }]
+        },
+        {
+          "id": "range-2",
+          "type": "book",
+          "title": "The Title",
+          "issued": [{ "date-parts": [2000] }, {}]
+        }
+      ],
       "items": {
         "$ref": "#/definitions/date-object"
       },
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -452,7 +452,9 @@
           "type": "book",
           "title": "The Title",
           "issued": {
-            "date-parts": [2000, 3, 10]
+            "year": 2000,
+            "month": 3,
+            "day": 10
           },
           "approximate": "true"
         }
@@ -462,13 +464,13 @@
         "year": {
           "title": "year",
           "description": "A year; can be negative.",
-          "examples": ["2012","-32"],
+          "examples": ["2012", "-32"],
           "type": "integer"
         },
         "month": {
           "title": "month",
           "description": "A month; can also represent seasons using the EDTF 21-24 values.",
-          "examples": ["2","22"],
+          "examples": ["2", "22"],
           "type": "integer"
         },
         "day": {
@@ -490,8 +492,21 @@
           "title": "Uncertain",
           "description": "Indicates uncertainty about a date representation, which may be presented as `May 23, 1972?`",
           "type": "boolean"
+        },
+        "undefined": {
+          "title": "Undefined",
+          "description": "Use to represent an open end of a range.",
+          "type": "boolean"
         }
-      }
+      },
+      "oneof": [
+        {
+          "required": ["year"]
+        },
+        {
+          "required": ["undefined"]
+        }
+      ]
     },
     "date-range": {
       "title": "Date Range",
@@ -502,13 +517,13 @@
           "id": "range-1",
           "type": "book",
           "title": "The Title",
-          "issued": [{ "date-parts": [2000] }, { "date-parts": [2001] }]
+          "issued": [{ "year": 2000 }, { "year": 2001 }]
         },
         {
           "id": "range-2",
           "type": "book",
           "title": "The Title",
-          "issued": [{ "date-parts": [2000] }, {}]
+          "issued": [{ "year": 2000 }, { "undefined": true }]
         }
       ],
       "items": {
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -499,7 +499,7 @@
           "type": "boolean"
         }
       },
-      "oneof": [
+      "oneOf": [
         {
           "required": ["year"]
         },
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -437,6 +437,78 @@
     "additionalProperties": false
   },
   "definitions": {
+    "edtf-string": {
+      "title": "EDTF date string",
+      "description": "CSL input supports EDTF, levels 0 and 1.",
+      "type": "string",
+      "pattern": "^[0-9-%~X?./]{4,}$"
+    },
+    "date-object": {
+      "title": "Date Object",
+      "description": "An EDTF date represented as an object.",
+      "type": "object",
+      "properties": {
+        "date-parts": {
+          "type": "array",
+          "items": {
+            "type": "integer"
+          },
+          "minItems": 1,
+          "maxItems": 3
+        },
+        "literal": {
+          "title": "Literal date",
+          "description": "A date that should be passed through to the processor as is; for dates that cannot be represented in EDTF.",
+          "examples": ["Han Dynasty"],
+          "type": "string"
+        },
+        "approximate": {
+          "title": "Approximate",
+          "description": "Indicates an approximate or circa date, which may be presented as `circa May 23, 1972`.",
+          "type": "boolean"
+        },
+        "uncertain": {
+          "title": "Uncertain",
+          "description": "Indicates uncertainty about a date representation, which may be presented as `May 23, 1972?`",
+          "type": "boolean"
+        }
+      },
+      "required": ["date-parts"]
+    },
+    "date-range": {
+      "title": "Date Range",
+      "description": "An EDTF range is a two-item array. An open end or begin of a range can be represented with a null.",
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/date-object"
+      },
+      "minItems": 2,
+      "maxItems": 2
+    },
+    "date-structured": {
+      "title": "Structured Date",
+      "description": "Can either be an object, or a two-item array.",
+      "oneOf": [
+        {
+          "$ref": "#/definitions/date-object"
+        },
+        {
+          "$ref": "#/definitions/date-range"
+        }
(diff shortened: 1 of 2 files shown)


schemas/input/csl-edtf-date.json
```diff
@@ -0,0 +1,69 @@
...
+      "title": "Date Object",
+      "description": "An EDTF date represented as an object.",
+      "type": "object",
+      "date-parts": {
+        "type": "array",
+        "items": [
+          {
+            "type": "integer",
+            "description": "year; a minus sign at the beginning denotes a negative year",
+            "pattern": "[0-9,X]{4}"
```
bdarcus, line 17:
```suggestion
            "pattern": "-?[0-9,X]{4}"
```

bdarcus:
> @bdarcus Should `pattern` include the minus sign? And for precision, it should probably say "a hyphen-minus…"

Yes, I meant to do that earlier.

dhimmel:
So the idea is to modify date-parts to make it better at handling a parsed EDTF string?

> Aside from need for review of the specifics of the schema, also leaves open whether and how to use in the csl-data.json schema.

As per https://json-schema.org/understanding-json-schema/structuring.html, I think you can use `$ref` to point to load this schema via its URL.

bdarcus:
> So the idea is to modify date-parts to make it better at handling a parsed EDTF string?

Yes; to make them consistent.

It's based on the [EDTF.js](https://github.com/inukshuk/edtf.js) library's modeling, though there what we have as `date-parts` is simply `values`.
 
> > Aside from need for review of the specifics of the schema, also leaves open whether and how to use in the csl-data.json schema.
> 
> As per https://json-schema.org/understanding-json-schema/structuring.html, I think you can use `$ref` to point to load this schema via its URL.

Thanks, Do you have a view on whether we should?

dhimmel:
> Thanks, Do you have a view on whether we should?

Is there any reason this whole file shouldn't go in the CSL Data schema as a variable?

bdarcus:
No; more just that I wanted to see what it would look like without touching
the main schema file.

But if there's support for this, easy enough to add it directly.

bdarcus:
> Is there any reason this whole file shouldn't go in the CSL Data schema as a variable?

I just pushed a commit that moves it back into csl-data.json.

bdarcus:
Though I may have a little bug I need to fix!

schemas/input/csl-data.json
```diff
@@ -437,6 +437,78 @@
...
+          "title": "Uncertain",
+          "description": "Indicates uncertainty about a date representation, which may be presented as `May 23, 1972?`",
+          "type": "boolean"
+        }
+      },
+      "required": ["date-parts"]
+    },
+    "date-range": {
+      "title": "Date Range",
+      "description": "An EDTF range is a two-item array. An open end or begin of a range can be represented with a null.",
```
bdarcus, line 480: The null option for open currently isn't implemented. Not sure how to do this.

bwiernik: Would this or permitting an empty `date-parts` array be better?

schemas/input/csl-data.json
```diff
@@ -437,6 +437,102 @@
...
+        "uncertain": {
+          "title": "Uncertain",
+          "description": "Indicates uncertainty about a date representation, which may be presented as `May 23, 1972?`",
+          "type": "boolean"
+        }
+      }
+    },
+    "date-range": {
+      "title": "Date Range",
+      "description": "An EDTF range is a two-item array. An open end or begin of a range can be represented with an empty (date) object.",
```
bwiernik, line 490:
```suggestion
      "description": "An EDTF range is a two-item array. An open end or beginning of a range can be represented with an empty (date) object.",
```

bwiernik:
This LGTM. We may wish to include some guidance in the processor addendum to the spec about how to process the old date model in CSL 1.1+ (e.g., what to do if `circa` is encountered).

schemas/input/csl-data.json
bdarcus: IDK; if I don't hear from any developers, am just going to go with this and figure we can correct it before merging to master.

But I was looking for the most minimal representation of the open item. I also considered this:

``` json
 {
    "id": "four",
    "type": "book",
    "title": "The Title",
    "issued": [
      { "date-parts": [2000] },
      { "open": "true" }
    ]
  }
```

bdarcus:
> This LGTM. We may wish to include some guidance in the processor addendum to the spec about how to process the old date model in CSL 1.1+ (e.g., what to do if `circa` is encountered).

We'll in general need to figure out that last issue.

In a style, for example, what to do with these?

``` yaml
issued: 1945-04-05~/1945-04-07
```

``` yaml
issued: 1945-04-05/1945-04-07~
```

``` yaml
issued: 1945-04-05~/1945-04-07~
```

bdarcus:
I think I should mention this, since I mentioned [here](https://github.com/citation-style-language/schema/issues/278#issuecomment-656159699):

I think date-parts is better represented as an object, as it is in pandoc-citeproc. I can see no advantage at all to the array, given what we're modeling here, other than that it fits with how `EDTF.js` models it.

But unless developers want to argue the case, it seems we should defer to status quo on this.

In any case, just to illustrate what I mean, `date-parts` would no longer be needed, so result would be:

``` json
{
 "id": "four",
 "type": "book",
 "title": "The Title",
 "issued": {
   "year": "1932",
   "month": "10",
   "day": "22",
   "approximate": "true"
  }
}
```

But that's an incompatible change of course, and it may not be worth is given the option to deprecate this entirely longer term in favor of the EDTF string.

bwiernik:
Compatibility with edtf.js is a major advantage I think. It's not clear to me how the object format you describe handles date ranges, which I think is the major motivation behind the array structure?

bdarcus:
> Compatibility with edtf.js is a major advantage I think. It's not clear to me how the object format you describe handles date ranges, which I think is the major motivation behind the array structure?

Easy; the range is still an array.

``` json
{
   "id":"four",
   "type":"book",
   "title":"The Title",
   "issued":[
      {
         "year":"2012",
         "month":"21"
      },
      {
         "year":"2012",
         "month":"22"
      }
   ]
}
```

bwiernik:
Oh, I see. You mean that the date itself is an object, whereas the date range is an array.

bdarcus:
> Oh, I see. You mean that the date itself is an object, whereas the date range is an array.

Not exactly.

Date is an object in this PR, and date-range is an array.

Currently, the date-parts property within the date object is also an array.

This is what I was focusing on in these comments; that this array was not necessary, and in hindsight, not the best decision.

inukshuk:
@bdarcus the CSL `date-parts` may have influenced the `values` array initially but mainly it can be used very efficiently internally: on the one hand, the array's length corresponds to the date's 'precision' and it's easy to feed the array values to JavaScript's `Date` constructor; approximate, undefined and uncertain values are stored as bitmasks so representing dates as an array of numbers makes computing and applying those masks relatively easy, regardless of the date's precision level.

Overall, I'd say the `values` array owes much to edtf.js implementation details and JavaScript in general. I think that using single object, without a `date-parts` array is probably much easier to read and write.

bdarcus:
Thanks much @inukshuk.

FYI , I just pushed a change to remove date-parts and add year, month, day,  so people can see the concrete change. 

I can remove it if objections.

schemas/input/csl-data.json
```diff
@@ -437,6 +437,110 @@
...
+    "date-object": {
+      "title": "Date Object",
+      "description": "An EDTF date represented as an object. Seasons can be represented using the 21-24 month values of EDTF, decades and centuries using the X notation, and the qualifiers with the included boolean properties",
+      "examples": [
+        {
+          "id": "one",
+          "type": "book",
+          "title": "The Title",
+          "issued": {
+            "date-parts": [2000, 3, 10]
```
bdarcus, line 455:
```suggestion
            "year": "2000",
            "months": "3",
            "day": "10"
```

schemas/input/csl-data.json
```diff
@@ -437,6 +437,110 @@
...
+    "date-range": {
+      "title": "Date Range",
+      "description": "An EDTF range is a two-item array. An open end or beginning of a range can be represented with an empty (date) object.",
+      "type": "array",
+      "examples": [
+        {
+          "id": "range-1",
+          "type": "book",
+          "title": "The Title",
+          "issued": [{ "date-parts": [2000] }, { "date-parts": [2001] }]
```
bdarcus, line 505: If we ditch date-parts:

```suggestion
          "issued": [{ "year": "2000" }, { "year": "2001" }]
```

schemas/input/csl-data.json
```diff
@@ -437,6 +437,110 @@
...
+        },
+        {
+          "id": "range-2",
+          "type": "book",
+          "title": "The Title",
+          "issued": [{ "date-parts": [2000] }, {}]
```
bdarcus, line 511: If we ditch date-parts:

```suggestion
          "issued": [{ "year": "2000" }, { }]
```

bdarcus:
So should we just GTM?

bwiernik:
So long as give processor instructions for handling the old model, LGTM.

bdarcus:
This is basically ready, but the CI script is killing me. The test on push passes, but the test on pull doesn't.

dhimmel:
The error at https://github.com/citation-style-language/schema/pull/301/checks?check_run_id=859242336#step:5:38 makes me think the CI run is for the old commit where the code has `true` and not `'true'` (although you may want `True`).

Perhaps try rerunning CI... maybe some github actions problem where it ran on an old commit

bdarcus:
> although you may want `True`

This turned out to be the issue.

I had reran the job, and it had still failed.

In any case, thank you!

bdarcus:
Aside: the json schema landscape is pretty frustrating. Things that pass as
valid with some tools, don't with others. So it's hard to know what's
correct.

PR #304, opened 2020-07-08
Reconcile v1.1 with v1.0.2

bwiernik:
## Description

Merge v1.0.2 into v1.1 and resolve conflicts.


bdarcus:
Thanks. After you run trang on csl-choose, I presume we should merge this now?

bdarcus:
Tip: if you use the script, or run trang on the main csl.rnc file, it will also format the included ones.

bwiernik:
I haven't been able to get the pre-commit hook to work on my system. It never completes. I ran trang on all of the rnc files. Thanks for the tip on the main csl.rnc.

bdarcus:
> I haven't been able to get the pre-commit hook to work on my system.

That's good to know; I was wondering.

Did you get any meaningful error messages?

On git, am I doing a standard merge commit; not a squash commit?

If the former, do you want to rebase the last two formatting fixes for a cleaner history?

Or shall I just merge as is?

bwiernik:
One thing I wonder is why trang wasn't concerned about some of these line breaks before when they were committed on the master branch?

bwiernik:
No error messages, just never hanged.

bdarcus:
I wonder if it's possible there's a version mismatch between what we use
and what's on Ubuntu (which is what the CI uses)?

Granted, trang is pretty stable, so I doubt it.

bdarcus:
The current script isn't supposed to fail; it's supposed to silently fix
the formatting.

bwiernik:
I'll rebase against master and trang everything instead.

PR #306, opened 2020-07-09
Add conditional logic to terms

### LINKED ISSUE #282: Allow conditional logic in terms definitions

Some languages have multiple forms for a term depending on the item type (usually books versus journals). For example, [Turkish](https://forums.zotero.org/discussion/81900/is-it-possible-different-ibid-term-for-books-article-and-web-page#latest) uses different phrases for "ibid" for books vs. journals. [German](https://github.com/citation-style-language/csl-evolution/issues/34) uses different words for "volume" for books vs. journals. 

@denismaier suggested we allow `choose` within terms: https://github.com/citation-style-language/csl-evolution/issues/34#issuecomment-624647159. I think that will work to accommodate this.

denismaier:
## Description

As per https://github.com/citation-style-language/schema/issues/282#issue-646718876:
Some languages have multiple forms for a term depending on the item type (usually books versus journals). For example, [Turkish](https://forums.zotero.org/discussion/81900/is-it-possible-different-ibid-term-for-books-article-and-web-page#latest) uses different phrases for "ibid" for books vs. journals. [German](https://github.com/citation-style-language/csl-evolution/issues/34) uses different words for "volume" for books vs. journals.

This allows `cs:choose` as a child element of `cs:term`.

Fixes #282 

## Question?

Should we allow other tests as well? Some of them/all of them?

I admit the current definition is overly complex as it follow the new pattern for conditions. This is unnecessary given that this currently only allows testing for item type. Even with more tests the pattern is ridiculously complex. We could probably simplify to

```xml
<term name="volume>
  <if type="article>asdf</if>
  <else>jklö</else>
<term>
```

But that would only increase complexity on another level, as the processor would now have to process two different conditional structures.

## Type of change

## Checklist

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -293,7 +293,7 @@ div {
   ## When specifying conditions, we allow either the attribute syntax or the element-based one.
   terms.term.choose.condition =
     (terms.term.condition.atts, match?)
-    | element cs:conditions { match,  terms.term.choose.condition.elem }
+    | element cs:conditions { match?,  terms.term.choose.condition.elem }
   terms.term.choose.if = element cs:if { terms.term.choose.condition,  terms.term.elem }
   terms.term.choose.else-if =
     element cs:else-if { terms.term.choose.condition,  terms.term.elem }
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -276,11 +276,37 @@ div {
   
   ## The "cs:term" element can either hold a basic string, or "cs:single" and
   ## "cs:multiple" child elements to give singular and plural forms of the term.
+  ## To conditionally define terms use "cs:choose" as child element of "cs:term".
+
   terms.term =
     element cs:term {
-      term.attributes,
-      (text | (term.single, term.multiple))
+      term.attributes, (terms.term.elem | terms.term.choose)
     }
+
+  terms.term.elem = text | (term.single, term.multiple)
+
+  terms.term.choose =
+
+    ## Used to conditionally define terms.
+    element cs:choose { terms.term.choose.if, terms.term.choose.else-if*, terms.term.choose.else? }
+  
+  ## When specifying conditions, we allow either the attribute syntax or the element-based one.
+  terms.term.choose.condition =
+    (terms.term.condition.atts, match?)
+    | element cs:conditions { match,  terms.term.choose.condition.elem }
+  terms.term.choose.if = element cs:if { terms.term.choose.condition,  terms.term.elem }
+  terms.term.choose.else-if =
+    element cs:else-if { terms.term.choose.condition,  terms.term.elem }
+  terms.term.choose.else = element cs:else { terms.term.elem }
+
+  terms.term.choose.condition.elem = element cs:condition { match?, terms.term.condition.atts }
+
+  terms.term.condition.atts = 
+    ## Tests whether the item matches the given types.
+    attribute type {
+      list { item-types+ }
+    }
+
   term.attributes =
     (attribute name { terms },
      [ a:defaultValue = "long" ] attribute form { term.form }?)


denismaier:
Concerning the question "should we adopt the regular choose/if/conditions/condition-syntax or should we simplify", we'd probably need feedback from @fbennett, @jgm, @cormacrelf, @PaulStanley

PaulStanley:
I don't massively love this because it means terms have quite different types: some of them are just strings (which are NOT a type within a CSL file) and others are fragments of CSL, which would have to be parsed as such, and  that will get a bit messy. And what about fallback? If I get a term, but none of the conditions match, do I fallback? What to?

Would an alternative, if it really is just "type" to allow terms to be defined with a "for-type" attribute (like "gender-form") which would then allow fallback to a term which did not have any "for-type" attribute ...

denismaier:
> Would an alternative, if it really is just "type" to allow terms to be defined with a "for-type" attribute (like "gender-form") which would then allow fallback to a term which did not have any "for-type" attribute ...

So, essentially that would be similar to my alternative proposals [here](https://github.com/citation-style-language/schema/issues/282#issuecomment-654799767) and [here](https://github.com/citation-style-language/schema/issues/282#issuecomment-654944265), right? 

I think that would be easier overall, at least from a style coding perspective, so yes, sounds good to me.
What do you @PaulStanley think about @bwiernik's objections/doubts to such an approach?

bwiernik:
I don't have a problem with `type="book"` type specification in terms, so long as something like `type="book chapter"` is allowed to permit a term to apply to multiple types. Note that is is not the same behavior as `type=*` generally.

denismaier:
Hmm, perhaps we should call it `types` then?

bwiernik:
Yeah, a new `types` attribute for terms would work.

denismaier:
I actually imagine you will only define terms for specific types in styles, the general definition being in the locale files.

PaulStanley:
I don't see any technical objection to specifying multiple types. Although in a locale file multiple types might be specified on a single term, and that usage is familiar from the conditional test too. In practice, I suspect, one would treat a single

```
<term ... type="book chapter">volume</term>
```

as merely syntactic sugar for
```
<term ... type="book">volume</term>
<term ... type="chapter">volume</term>
```

It might be useful to have allow a special syntax in the CSL file when looking for a term `type="matched-to-type"` which would instruct the processor to use *whatever type the source has", so one could have

```
<text term="volume" type="matched-to-type" form="short"/>
```

Or one might even want to specify that this is the default, though that would probably be overkill. Anyway, that would avoid wasting time with explicit type tests in the CSL. I don't know if that works from a CSL developer perspective ...

bwiernik:
@denismaier No, I think that would be really rare. I imagine the primary place where these would be defined would be for the `volume` term in languages like German and Turkish that have different terms for periodicals versus books and similar. Turkish makes the same distinction for `ibid`.

denismaier:
I thought the rendered term should always match the definition for the current type.

bwiernik:
> It might be useful to have allow a special syntax in the CSL file when looking for a term type="matched-to-type" which would instruct the processor to use *whatever type the source has", so one could have

I think that this should just be the behavior all the time. I don't see a need for a style to not have that behavior.

bdarcus:
Maybe a dumb/lazy question, but are all types now tokenized?

PaulStanley:
That's fine. I just wondered whether it was overkill. Most terms (months, ordinals, pages etc) are going to be the same most of the time, so most lookups for specific types will fail, in which case one wastes a few cycles ... But not technically troublesome.

bwiernik:
@bdarcus No types were ever not tokenized, and we will definitely be sure none are ever added that aren't.

fbennett:
> @bdarcus No types were ever not tokenized, and we will definitely be sure none are ever added that aren't.

Zotero has a ``document`` type that once shipped to the processor with an undefined item type, which (after it was discovered) was handled in styles by amending them where necessary to include a `match="none"` condition for every type known to CSL. It's no longer an issue, since Zotero now maps its ``document`` type to ``article``.

denismaier:
If I remember correctly, we introduce a generic `document` item type in 1.0.2. So Zotero can map its own `document` type to CSL ` document`

denismaier:
@PaulStanley I understand. But I think that's preferable to having to know which terms might need an item specific rendering.

bdarcus:
I closed this in favor of #308, since the description here focuses on terms. We can always add to the other solution as needed.

PR #308, opened 2020-07-11
add @types to terms

### LINKED ISSUE #282: Allow conditional logic in terms definitions

denismaier:
## Description

Alternative PR to #306

Some languages have multiple forms for a term depending on the item type (usually books versus journals). For example, Turkish uses different phrases for "ibid" for books vs. journals. German uses different words for "volume" for books vs. journals.

Instead of using the standard `cs:choose` syntax this PR just adds a new attribute `types` to be used with term definitions.

In styles or locale files this will be used like this:

```xml
<terms>
  <term name="volume" types="article article-journal article-magazine article-newspaper">a term</term>
  <term name="volume">another term</term>
</terms>
```

Fixes #282 (issue)

## Type of change

## Checklist

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -283,24 +283,31 @@ div {
     }
   term.attributes =
     (attribute name { terms },
-     [ a:defaultValue = "long" ] attribute form { term.form }?)
+     [ a:defaultValue = "long" ] attribute form { term.form }?,
+     term.types)
     | (attribute name { terms.ordinals },
        attribute form { "long" }?,
        attribute gender-form { "masculine" | "feminine" }?,
        attribute match {
          "last-digit" | "last-two-digits" | "whole-number"
-       }?)
+       }?,
+       term.types)
     | (attribute name { terms.long-ordinals },
        attribute form { "long" }?,
        attribute gender-form { "masculine" | "feminine" })
     | (attribute name { terms.gender-assignable },
        attribute form { "long" }?,
-       attribute gender { "masculine" | "feminine" })
+       attribute gender { "masculine" | "feminine" },
+       term.types)
   
   ## "verb-short" reverts to "verb" if the "verb-short" form is not available.
   ## "symbol" reverts to "short" if the "symbol" form is not available.
   ## "verb" and "short" revert to "long" if the specified form is not available.
   term.form = "long" | "verb" | "short" | "verb-short" | "symbol"
+  term.types = 
+    attribute types {
+        list { item-types+ }
+      }
   term.single =
     
     ## Singular version of the term.
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -304,10 +304,11 @@ div {
   ## "symbol" reverts to "short" if the "symbol" form is not available.
   ## "verb" and "short" revert to "long" if the specified form is not available.
   term.form = "long" | "verb" | "short" | "verb-short" | "symbol"
-  term.types = 
+  # Used to specify terms per item type
+  term.types =
     attribute for-types {
-        list { item-types+ }
-      }
+      list { item-types+ }
+    }
   term.single =
     
     ## Singular version of the term.
@@ -767,7 +768,12 @@ div {
     
     ## Reformat page ranges in the "page" variable.
     attribute page-range-format {
-      "expanded" | "minimal" | "minimal-two" | "minimal-oup" | "chicago" | "mhra" 
+      "expanded"
+      | "minimal"
+      | "minimal-two"
+      | "minimal-oup"
+      | "chicago"
+      | "mhra"
     }?
   citation.cite-group-delimiter =
     
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -305,7 +305,7 @@ div {
   ## "verb" and "short" revert to "long" if the specified form is not available.
   term.form = "long" | "verb" | "short" | "verb-short" | "symbol"
   term.types = 
-    attribute types {
+    attribute for-types {
         list { item-types+ }
       }
   term.single =


denismaier:
@PaulStanley WDYT?

PaulStanley:
I don't have any difficulty with the structure. I'm still a bit worried that (here as elsewhere) additional tests are being introduced to cope with rare situations which in practice add slightly to the processing time in common cases to no particular purpose. Whether that is *actually* a problem, only time will tell. I don't know how heavy the hit will be, or how critical processing times are. With anything that is processing in a "batch" method (e.g. I guess Pandoc), it's probably at worst a slight nuisance. With something which expects to update citations "live" it could be more troublesome. But I think the best thing is probably to write it first how you think will be best from the user perspective, and then deal with performance issues if/when they arise, on the usual principle that premature optimisation etc etc.

denismaier:
> I'm still a bit worried that (here as elsewhere) additional tests are being introduced to cope with rare situations which in practice add slightly to the processing time in common cases to no particular purpose. 

Do you have a suggestion how this could be done better?

I actually thought you wouldn't have to do the tests when the items are rendered, but you'd just create different arrays of terms when parsing the style. So the processor would then call the term from the appropriate array. (Most of the arrays would be identical of course, and I'm not sure this is better than testing when terms are in fact rendered...)

PaulStanley:
I think you are basically right, which is why I say the "hit" is tiny.

Let's suppose this.

I parse a style file, and I load the relevant locale terms from one or more locales. Let's suppose (quite realistically) that I end up with four possible "sources" for a term defined in (a) the style locale, (b) a default locale (say "fr-CA"), (c) a fallback locale (say "fr-FR") and (d) the ultimate fallback "en-US".

Now I hit a request for a term (let's say "month-01") which is not specialised by type. My search process runs as follows:

* Is `month-01 with type="book"` in the style locale (No)
* Is it in the default locale (No)
* Is it in the fallback locale (No)
* Is it in en-US (No)
* Is `month-01 with type="any"` in the style locale (Yes)

So that is four failed queries for one successful one. And that happens for *every* request for *every* term. A failed query is still a "thing I had to do", whereas if I *know* in advance that type is irrelevant, I need make only one query (plus one test of course to see if I am dealing with a term for which it might be relevant).

Now, to be sure, this is no big deal. Four queries to a structure in which lookup should be O(1) are not a big deal, and if it turned out they *were* a big deal there are some ways I can speed things up (e.g. I could actually store every term for every type, which would slow down locale creation and increase the amount of space I need but guarantee a hit every time, but these are not big files mostly; I can cache individual citations in many cases). So I'm not actually objecting to the proposal. I'm just saying that one needs to be a tiny bit careful because sometimes performance mismanagement can be death by a thousand cuts, and if you *know* this will only be needed for a few terms, it wouldn't be wrong limit it to those terms.

denismaier:
Not sure, but couldn't you construct an array for month-01 at the beginning? And then you'll get the term from this array? So you'll have the four times lookup only once, but of course for each term and for each item type, even if it's not relevant. Don't know which is better...
  
  
   
    PaulStanley <notifications@github.com> hat am 14.07.2020 12:00 geschrieben:
   
   
    
   
   
    
   
   
   I think you are basically right, which is why I say the "hit" is tiny.
   Let's suppose this.
   I parse a style file, and I load the relevant locale terms from one or more locales. Let's suppose (quite realistically) that I end up with four possible "sources" for a term defined in (a) the style locale, (b) a default locale (say "fr-CA"), (c) a fallback locale (say "fr-FR") and (d) the ultimate fallback "en-US".
   Now I hit a request for a term (let's say "month-01") which is not specialised by type. My search process runs as follows:
   
    Is month-01 with type="book" in the style locale (No)
    Is it in the default locale (No)
    Is it in the fallback locale (No)
    Is it in en-US (No)
    Is month-01 with type="any" in the style locale (Yes)
   
   So that is four failed queries for one successful one. And that happens for every request for every term. A failed query is still a "thing I had to do", whereas if I know in advance that type is irrelevant, I need make only one query (plus one test of course to see if I am dealing with a term for which it might be relevant).
   Now, to be sure, this is no big deal. Four queries to a structure in which lookup should be O(1) are not a big deal, and if it turned out they were a big deal there are some ways I can speed things up (e.g. I could actually store every term for every type, which would slow down locale creation and increase the amount of space I need but guarantee a hit every time, but these are not big files mostly; I can cache individual citations in many cases). So I'm not actually objecting to the proposal. I'm just saying that one needs to be a tiny bit careful because sometimes performance mismanagement can be death by a thousand cuts, and if you know this will only be needed for a few terms, it wouldn't be wrong limit it to those terms.

PaulStanley:
I don't think so. Because you don't know what term you need until you have a *cite* with a type for it, which by definition you don't know as you parse the style file. You could, as I say, construct multiple terms for "month-01" (form "book"), "month-01" (form "article") ... and that would give you a 1 hit lookup, but at the price of spending time initially constructing ALL those sub-terms for EVERY term, which in many cases will be a big waste. You could cache month-01-book etc as used, I suppose, but you are still then adding 1 cache seek, and my hunch is that the more sensible cacheing behaviour (I haven't tried it) is to cache whole cites because it's *relatively* easy to determine whether a cite is cacheable (loosely if it has no conditional check which depends on position, no locator, and no per-cite disambiguation). In fact you *should* be able to cache cites that meet those criteria on the first disambiguation attempt ...

But I don't want to get bogged down in the detail. You are *absolutely right* to say that this is by no means a deal-breaker in this case. My only point, which I don't want to press too hard, is that every time you add a special form which is intended to deal with just a few individual cases but in a way that requires a generalised fallback structure you may be adding to the processing burden because we will have to tolerate, in the general case, a number of misses for a single hit. Bearing in mind that a citation processor will quite likely "prepare" citations more often than it actually uses them (as it goes through disambiguation and so on) and bearing in mind that responsiveness in this context is a virtue, you want to keep half an eye on those things, and make sure you are satisfied the game is worth the candle. In the end, however, be guided by what is right for users and leave it to those who design processors to make it work. There's nothing complicated about a processor: it's just a tangle of conditional checks!

bdarcus:
So I'm unclear where we're at now. We have two PRs to address the same issue. Which do you prefer @PaulStanley?

We can separately assess the clarity for style authors, so this is just from the processing standpoint.

PaulStanley:
I prefer this proposal: I don't like having "dynamic" CSL in locales, except in very well-defined cases (like localised date parts) which in effect define macros via locale files, and think this proposal is clearer.

bdarcus:
Thanks.

Not having followed this closely, question:

First, what does "@types" mean on a term? Is that not effectively syntactic sugar for a conditional (choose/if)? So what's the practical difference in the end?

PaulStanley:
In general (at present) terms defined in locale files are not processor instructions or fragments of them, but simply strings. (The exception is localized date parts).

So when a processor needs a term, it simply carries out a lookup into a data structure holding the terms, with fallback if it can't find a particular term or specialised form/gender/number etc.

The `types` syntax preserves this behaviour. There is essentially a (possible) specialised term "volume-singular-long-book-masculine-singular", and one simply looks it up and falls back if it's not there (e.g. to "volume-singular-long") etc. The `choose` syntax would introduce a completely new behaviour where after finding a "term" one then has to process it with processor instructions. That in effect changes the *type* of a "term": it's now not always a string, but sometimes a special form of processor instruction. So now you have to handle two types: "strings" and "processor instructions to create strings". That's messy, if avoidable.

So it's not just syntactic *sugar*: "type" won't translate to "choose" under the hood. Type will simply produce a specialised term which may or may not be present, with fallback rules to select alternatives where it isn't. Terms remain (always) strings which are looked up, and never become functions that produce strings.

bdarcus:
Ok, thank you!

>

denismaier:
`for-types` is good for me. 
If we are sure (at least for the moment) that we don't  need other tests, I also think we should close the other PR.

schemas/styles/csl.rnc
```diff
@@ -283,24 +283,31 @@ div {
...
-       attribute gender { "masculine" | "feminine" })
+       attribute gender { "masculine" | "feminine" },
+       term.types)
   
   ## "verb-short" reverts to "verb" if the "verb-short" form is not available.
   ## "symbol" reverts to "short" if the "symbol" form is not available.
   ## "verb" and "short" revert to "long" if the specified form is not available.
   term.form = "long" | "verb" | "short" | "verb-short" | "symbol"
+  term.types = 
+    attribute types {
```
bdarcus, line 308:
```suggestion
    attribute for-types {
```

bdarcus: Also, don't you need to add a docstring here?

bwiernik: I like this. LGTM to merge after that.

bwiernik:
I am fairly confident this is the only test (beyond gender, plural that already exist) we will need. I think this is the right direction.

schemas/styles/csl.rnc
denismaier: docstring is there now

bdarcus:
Thanks.

PR #314, opened 2020-07-12
Fix typo

zuphilip:


bdarcus:
Can you add the link to the discourse forum here too, so we can bundle the changes into one commit?

zuphilip:
Done. Please have a look at the new version.

PR #317, opened 2020-07-13
Add YAML schema, workflow to publish json schema, docs

### LINKED ISSUE #316: Use YAML to maintain the input schema
I find working with JSON awkward, and would like to suggest we move to a YAML representation instead.

Digging around a bit, it seems people are using YAML to maintain JSON schemas.

- https://www.tutorialspoint.com/yaml/yaml_json_schema.htm
- https://stenci.la/blog/2019-02-19-json-schema/

Here's our's in YAML:

https://gist.github.com/bdarcus/4afa2ba2a0a57d8f9af7ccf8af7c5320

We could have a Github Action that would keep a JSON variant in sync, but do new development on the YAML.

Thoughts?

cc @dhimmel @bwiernik @denismaier

bdarcus:
Add new YAML version of the data schema (citation schema to come), and beginning of workflow 
to convert YAML schema to JSON and publish on GitHub pages.

Goal is for the documentation to be published in the same workflow.

Still to settle:

1. folder structure vis-a-vis versions? currently it's `input/1.0/csl-data.json`, for example
2. update id? do we need the redirect? if the schema is yaml and json, should the id have an extension?
3. the action is failing with a command not found error, after success on installing the command
4. python setup
5. do I `git rm` for the json files, or `git mv`?

Closes #316.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -13,7 +13,7 @@ jobs:
         run: |
           npm install js-yaml
           # pip install json-schema-for-humans
-          # js-yaml schemas/input/csl-data.yaml > schemas/input/latest/csl-data.json # currently leave off, as no yaml file
+          js-yaml schemas/input/csl-data.yaml > schemas/input/latest/csl-data.json # currently leave off, as no yaml file
           # TODO generate-schema-doc -–deprecated-from-description schemas/input/csl-data.yaml build/csl-data/index.html
 
       - name: Deploy 🚀
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,24 @@
+name: Publish
+on: [push]
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v2 # If you're using actions/checkout@v2 you must set persist-credentials to false in most cases for the deployment to work correctly.
+        with:
+          persist-credentials: false
+
+      - name: Install and Build 🔧 # convert the yaml schema to json version and docs, and publish on gh-pages
+        run: |
+          npm install js-yaml
+          # pip install json-schema-for-humans
+          # js-yaml schemas/input/csl-data.yaml > schemas/input/latest/csl-data.json # currently leave off, as no yaml file
+          # TODO generate-schema-doc -–deprecated-from-description schemas/input/csl-data.yaml build/csl-data/index.html
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@releases/v3
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH: gh-pages # The branch the action should deploy to.
+          FOLDER: schemas/input/latest # The folder the action should deploy.
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -13,6 +13,7 @@ jobs:
         run: |
           npm install js-yaml
           # pip install json-schema-for-humans
+          mkdir schemas/input/latest
           js-yaml schemas/input/csl-data.yaml > schemas/input/latest/csl-data.json # currently leave off, as no yaml file
           # TODO generate-schema-doc -–deprecated-from-description schemas/input/csl-data.yaml build/csl-data/index.html
 
--- a/schemas/input/csl-data.yaml
+++ b/schemas/input/csl-data.yaml
@@ -1,0 +1,476 @@
+description: JSON schema for CSL input data
+$schema: http://json-schema.org/draft-07/schema#
+$id: https://resource.citationstyles.org/schema/latest/input/json/csl-data.json
+type: array
+items:
+  type: object
+  properties:
+    type:
+      type: string
+      enum:
+      - article
+      - article-journal
+      - article-magazine
+      - article-newspaper
+      - bill
+      - book
+      - broadcast
+      - chapter
+      - classic
+      - collection
+      - dataset
+      - document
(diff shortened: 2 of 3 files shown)


schemas/input/csl-data.yaml
```diff
@@ -0,0 +1,476 @@
+description: JSON schema for CSL input data
+$schema: http://json-schema.org/draft-07/schema#
+$id: https://resource.citationstyles.org/schema/latest/input/json/csl-data.json
```
bdarcus, line 3: should we leave the extension off the id?

bdarcus:
@dhimmel - if you have any time over the next few days, can you please help me figure this out?

I tried the most minimal command of simply converting yaml to json and moving to gh-pages, but even that fails with a "command not found" error, even though the npm module I use there gets correctly installed.

Am wanting also to create the docs too (commented out now), but that's less important as a first step. I just want to move the yaml version for development.

I did manually add the 1.0 and latest version schemas, so we at least have something there:

- http://citation-style-language.github.io/schema/input/1.0/csl-data.json
- http://citation-style-language.github.io/schema/input/latest/csl-data.json

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v2 # If you're using actions/checkout@v2 you must set persist-credentials to false in most cases for the deployment to work correctly.
+        with:
+          persist-credentials: false
+
+      - name: Install and Build 🔧 # convert the yaml schema to json version and docs, and publish on gh-pages
+        run: |
+          npm install js-yaml
```
dhimmel, line 14: > but even that fails with a "command not found" error, even though the npm module I use there gets correctly installed.

does `--global` fix the issue. otherwise have no idea... don't have much expertise with javascript packaging.

bdarcus, line 14:
```suggestion
          npm install --global js-yaml
```

bdarcus:
This was helpful @dhimmel; thank you.

I think I'm going to do this differently then:

Use the action you suggest, integrate it in the main workflow file (probably named to be more generic), and just do all python.

Currently the 1.1 branch is screwed up, and missing a lot of changes in master, including the recent updates to CI.

So I think I'll wait on this until we can tag 1.0.2 on master, and @bwiernik can then merge 1.1 and get our git history cleaned up.

tools/yaml2json.py
```diff
@@ -0,0 +1,20 @@
...
+import json
+import glob
+import sys
+import os
+
+# grab the yaml files
+files = glob.glob('schemas/input/csl-*.yaml')
+
+for file in files:
+    ofn = os.path.join('build', os.path.splitext(file)[0] + '.json')
```
bdarcus, line 13: this is going to need to be placed in the right directory for the gh-pages action to correctly work

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,25 @@
...
+          # pip install json-schema-for-humans
+          mkdir schemas/input/latest
+          js-yaml schemas/input/csl-data.yaml > schemas/input/latest/csl-data.json # currently leave off, as no yaml file
+          # TODO generate-schema-doc -–deprecated-from-description schemas/input/csl-data.yaml build/csl-data/index.html
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@releases/v3
```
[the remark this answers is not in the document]
bdarcus: I will indeed do this.

tools/yaml2json.py
dhimmel: the python [pathlib](https://docs.python.org/3/library/pathlib.html) builtin library provides slightly more elegant ways to do these path operation. No need to switch, but just wanted to let you know if you're interested.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,26 @@
...
+          node-version: ${{ matrix.node-version }}
+      - name: Install and Build 🔧 # convert the yaml schema to json version and docs, and publish on gh-pages
+        run: |
+          npm install js-yaml
+          # pip install json-schema-for-humans
+          mkdir schemas/input/latest
+          js-yaml schemas/input/csl-data.yaml > schemas/input/latest/csl-data.json # currently leave off, as no yaml file
+          # TODO generate-schema-doc -–deprecated-from-description schemas/input/csl-data.yaml build/csl-data/index.html
+
+      - name: Deploy 🚀
```
dhimmel, line 21: Just realized you probably want to make this step conditional... you don't want to deploy pull requests or forks or perhaps non-master branches.

Example at https://github.com/manubot/rootstock/blob/ce827c7789163dca33dfa3bd2df7aa828e57485d/.github/workflows/manubot.yaml#L57

bdarcus:
I just switched to pathlib. Thanks!

How best to write the file? To a "build" directory?

And how best to do that with pathlib?

>

tools/yaml2json.py
```diff
@@ -1,17 +1,17 @@
...
 # -*- coding: utf-8 -*-
 import ruamel.yaml
+from pathlib import Path
 import json
 import glob
 import sys
-import os
 
 # grab the yaml files
 files = glob.glob('schemas/input/csl-*.yaml')
```
dhimmel, line 10: pathlib.Path objects have a .glob method if you want to get rid of another import.

dhimmel:
> How best to write the file? To a "build" directory?

yeah usually there should be a directory that you put the outputs in. You could name it `output`, `build`, etc. Make sure to add it to .gitignore so you don't track it in other branches. Then you use the `actions-gh-pages` to copy the directory contents to the branch of your choosing.

> And how best to do that with pathlib?

path.write_text() is nice.

PR #318, opened 2020-07-13
Feedback 1.0.2

bwiernik:
Feedback from @zuphilip and @adam3smith from https://discourse.citationstyles.org/t/feedback-csl-1-0-2-publication-type-event/1668/11 and https://discourse.citationstyles.org/t/feedback-csl-1-0-2-media-roles/1669/8

- Renamed `creator` to `series-creator`
- Renamed `writer` to `script-writer`
- Added `event-title` and deprecated `event`


schemas/input/csl-data.json
```diff
@@ -310,6 +310,9 @@
       "event": {
         "type": "string"
```
zuphilip, line 311: Can we use here a `description` field to indicated that this is deprecated?

zuphilip:
Can you also add publication type `event` in the JSON schema where it is still missing? I.e.  insert it to these lines
https://github.com/citation-style-language/schema/blob/357e7df15c1198165923d091a11c8acaaa932036/schemas/input/csl-data.json#L22-L28

schemas/input/csl-data.json
bdarcus: We should be adding `title` and `description` properties as we go, per #190.

bwiernik: How do we want to keep the schema description and rst documentation in sync? Or do we leave the schema description as a shorter brief description?

bdarcus:
Good question.

I want to generate the JSON documentation from the schema, with a GitHub
action.

Maybe we have that primary and point to it, so we only maintain in one
place?

In any case, details still TBD.

schemas/input/csl-data.json
```diff
@@ -308,6 +309,10 @@
         "type": ["string", "number"]
       },
       "event": {
+        "description": "Alias for 'event-title'. Deprecated. Will be removed in CSL 1.1.",
```
bdarcus, line 312:
```suggestion
        "description": "[Deprecated - use 'event-title' instead. Will be removed in 1.1]",
```

The above is a convention used in the documentation tool I was planing to use.

bdarcus:
Is this RTM @bwiernik?

bwiernik:
Sure. This is all the current feedback.

Issue #291, opened 2020-06-30
Revise name data model


denismaier:
See https://github.com/citation-style-language/csl-evolution/issues/29

I'm just adding this here, so we can add it to the 1.1 milestone.

@bwiernik Do you still think this should be in 1.1?

Closes https://github.com/citation-style-language/csl-evolution/issues/3#issuecomment-651939406

bwiernik:
Yes.

bwiernik:
A specific use case for the `label` specification required in a style manual: https://forums.zotero.org/discussion/84042/tromso-recommendations-for-citation-of-research-data-in-linguistics#latest

denismaier:
This can be closed, right?

PR #322, opened 2020-07-18
Expand support for institutional names, other name features

bwiernik:
## Description

Data model:
1. Add literal-short name element to hold an institutional abbreviation
1. Add is-uncertain name element 
1. Add alternate name element (e.g., for screen names, real names for pseudonymous authors, transliterated names)
1. Document that personal mononyms should be stored in "family", not "literal"

Styles:
1. Defines how the form attribute should work for non-personal names (cf. [the current spec](https://docs.citationstyles.org/en/1.0.1/specification.html#name))
1. Add is-uncertain-name condition for choose
1. Add `cs:name-type` element to provide separate names formatting for personal and non-personal names
   - "Personal" is defined as the variable having one or more names with "family" parts
   - "Non-personal" is defined as the variable having only names with "literal"/"literal-short" parts (no "family" or "given").
1. Add literal and alternate as options for `name-part` formatting

Closes https://github.com/citation-style-language/schema/issues/291
Closes https://github.com/citation-style-language/csl-evolution/issues/7

## Type of change

## Other details

One other thing we might consider is to add behavior like citeproc-js's handling of personal Eastern names that should not be abbreviated (e.g., 毛泽东 or Mao Zedong) to the forthcoming processor addendum.

See that behavior here:
https://github.com/Juris-M/citeproc-js/blob/0ef50770a7322d1c2237232ee12481226bc0284b/src/util_names_render.js#L1039
https://github.com/Juris-M/citeproc-js/blob/0ef50770a7322d1c2237232ee12481226bc0284b/src/util_names_render.js#L468
https://github.com/Juris-M/citeproc-js/blob/50570b1321a890ffc03d1367c0c9ce63a458dbfe/src/load.js#L623

@fbennett


schemas/input/csl-data.json
```diff
@@ -463,10 +465,27 @@
...
+            },
+            "alternate": {
+              "title": "Alternative name, such as screen name for online item or real name of pseudonymous author.",
+              "description": "E.g. rendered as 'Smith, J. [@JSmith]'",
               "type": "string"
             },
             "parse-names": {
               "type": ["string", "number", "boolean"]
+            },
+            "is-uncertain": {
```
bdarcus, line 485:
```suggestion
            "uncertain": {
```

To align with new date model.

schemas/input/csl-data.json
```diff
@@ -463,10 +465,27 @@
...
+              "title": "Indicates uncertain authorship",
+              "description": "E.g. rendered as '[Smith, J.?]'",
+              "type": ["string", "number", "boolean"]
```
bdarcus, line 488:
```suggestion
              "type": "boolean"
```

To align with new date model. There's no reason to leave this so open.

schemas/input/csl-data.json
```diff
@@ -463,10 +465,27 @@
               "type": ["string", "number", "boolean"]
             },
             "literal": {
```
bdarcus, line 467: Why not simply `name` and `name-short`?

schemas/input/csl-data.json
```diff
@@ -463,10 +465,27 @@
...
+              "description": "Use for short institutional creator names; e.g. 'NIH'",
+              "type": "string"
+            },
+            "alternate": {
+              "title": "Alternative name, such as screen name for online item or real name of pseudonymous author.",
+              "description": "E.g. rendered as 'Smith, J. [@JSmith]'",
               "type": "string"
             },
             "parse-names": {
               "type": ["string", "number", "boolean"]
```
bdarcus, line 483: I know you're not changing this, but per below, why is this so open?

schemas/input/csl-data.json
bwiernik: Okay. I was wondering the same thing.

schemas/input/csl-data.json
bwiernik: It seems that Booleans are all quite open for some reason.

bdarcus:
We should probably fix that.

schemas/input/csl-data.json
bdarcus, line 483:
```suggestion
              "type": "boolean"
```

Let's fix this now. A boolean that isn't a boolean seems pretty useless.

Also, we need a description here.

schemas/input/csl-data.json
```diff
@@ -442,6 +442,8 @@
         {
           "properties": {
             "family": {
+              "title": "Family name of creator",
```
bdarcus, line 445:
```suggestion
              "title": "Family name of contributor",
```

schemas/input/csl-data.json
bdarcus, line 467: I had wondered why we don't call this `name`, as we do in the CSL style schema?

Another decision you didn't make, but thought worth asking here.

schemas/styles/csl.rnc
```diff
@@ -507,11 +531,28 @@ div {
...
       font-formatting,
       text-case
     }
+  name.name-type =
+    
+    ## Use to apply separate formatting when a variable contains only personal
+    ## names or a combination of personal and non-personal names ("names") versus
+    ## contains only non-personal names.
+    element cs:name-type {
+      attribute name { "personal" | "non-personal" },
```
bdarcus, line 545: I don't really have a super-strong opinion here, but more common names for this are "corporate" (this is language from the library world) or "organizational"?

bwiernik: Don't have a strong opinion either. My thought was "Is something like 'Committee on Personality Disorders' or 'The Reproducibility Project' really 'organizational' or 'institutional'?" I think it would probably be fine to go with "organizational". I think "corporate" is a little too jargon-y.

schemas/input/csl-data.json
bwiernik: `name ` as opposed to `literal`? I think because the whole structure with `family`, `given`, `prefix`, etc. is a `name`. I had thought about whether we should rename this `organization` or `institution`.

bwiernik: > Another decision you didn't make, but thought worth asking here.

Yeah, feel free to raise these. Now's a good time to review the data model.

bdarcus:
I don't have a strong opinion. Just seemed an obvious question.

>

bwiernik:
@fbennett I based some parts of this on the CSLm [`cs:institution` element](https://citeproc-js.readthedocs.io/en/latest/csl-m/#cs-institution-extension)

Here are a few thoughts on why I wrote the PR the way I did:

1. I did not attempt to address author affiliations in any way
  - It seems like the combined treatment of affiliations and institutional authors is an artifact of the constraints of the Zotero interface? An author's affiliation seems like it should be entered as a name-part, not a separate name?
1. The subunit use-first/use-last and rearrangement attributes seemed like overkill to me. 

With those two considerations in mind (especially the first), it then seemed like it would make cleaner conceptual sense and style logic to make the element that controlled institutional versus personal name formatting (here `cs:name-type`) a child of `cs:name` rather than a sibling to it. The only things that would need to be controlled separately for the two name types would be the form, affixes, and text formatting, rather than all of the other name attributes.

@bdarcus 
One thing I'm not sure how to handle `form="long-short"` for an institution in combination with `cs:intext`. For example, APA style wants this for standard and narrative citations:
1. (American Psychological Association [APA], 2019)
1. American Psychological Association (APA, 2019)

This probably isn't that big a deal (though I can imagine getting some complaints; APA users are weird):
1. American Psychological Association (APA) (2019)

I'm thinking not worry about this? Or do you see any elegant solution?

bdarcus:
> One thing I'm not sure how to handle `form="long-short"` for an institution in combination with `cs:intext`. For example, APA style wants this for standard and narrative citations:

I'm not following here. What is `long-short` and why do we need it, as opposed, here, to simply `long` and `short`?

Because the current rules are names won't print twice?

If yes, could we make an exception for institutional names, and/or alternate forms?

bwiernik:
For institutional names, long-short renders the long form followed by the short form like this:

National Institutes of Health (NIH)

That is a common form, and it isn't well-covered by long (which presumably just renders long) or short (just renders short). It needs to be its own form because the long and short forms for each name need to be paired together. A new form is the easiest approach because it doesn't require a major change in the children of `cs:names`.

Styles vary in whether institution abbreviations are used and when long/short forms are rendered (e.g., both forms the first citation, then just short form for subsequent citations, just long form in bibliography). 

For personal names, long-short just renders the same as long. 

The limitation of the current PR is if long-short is rendered as a narrative citation. APA style, for example, wants this format for regular citations:
(National Institutes of Health [NIH], 2010)

And this format for narrative citations:
National Institutes of Health (NIH, 2010)

The current PR would render:
National Institutes of Health (NIH) (2010)

I don't see a way around that without making `cs:intext` more complicated. But I'm not too concerned.

bdarcus:
> For institutional names, long-short renders the long form followed by the short form like this:
> 
> National Institutes of Health (NIH)
> 
> That is a common form, and it isn't well-covered by long (which presumably just renders long) or short (just renders short). It needs to be its own form because the long and short forms for each name need to be paired together. A new form is the easiest approach because it doesn't require a major change in the children of `cs:names`.

To be clear, I was meaning some logic by which you could test if institutional, and then do the following, instead of using `long-short`:

``` xml
<text variable="author" form="long"/>
<text variable="author" form="short" prefix="[" suffix="]"/>
```
If not that, then don't we need to have special delimiters for these to support this?

bwiernik:
Yes, I understand. That doesn't work. For example:

(American Educational Research Association [AERA], American Psychological Association [APA], & National Council on Measurement in Education [NCME], 2014)

Each individual institution name should be immediately followed by by its short form. 

As an alternative to long-short, we could add a new element `cs:abbreviation` that is sibling to `cs:label`. For an institutional name, if `form="long"`, `cs:abbreviation` renders `form="short"`. Otherwise, it's empty. Then, specify the style like this:
```
<names variable="author">
  <name>
    <name-type name="personal" form="short"/>
    <name-type name="institutional" form="long" form-subsequent="short"/>
  </name>
  <abbreviation prefix=" [" suffix="]"/>
</names>
```

Thinking about it, this actually does work better for several cases than `form="long-short"`.

bdarcus:
> As an alternative to long-short, we could add a new element `cs:abbreviation` that is sibling to `cs:label`.

@denismaier - could you please review this idea that @bwiernik came up with?

bwiernik:
@bdarcus How should I interpret the eyes reaction?

bwiernik:
@bdarcus How should I interpret the eyes reaction?

bdarcus:
> @bdarcus How should I interpret the eyes reaction?

It's meant to signal reviewing/thinking about.

schemas/styles/csl.rnc
```diff
@@ -387,21 +390,29 @@ div {
...
+      name.forms,
       affixes,
       
       ## Set the delimiter for names in a name variable (e.g., ", " in
       ## "Doe, Smith")
       [ a:defaultValue = ", " ] delimiter,
       font-formatting,
-      name.name-part*
+      name.name-part*,
+      name.name-type*
```
denismaier, line 401: Is it intended that `name-part` and `name-type` are siblings? I'm not sure, but couldn't it make sense to define `name-part` as a child element of `name-type`? Or to allow both variants.

```xml
<names variable="author">
  <name>
    <name-part name="family" text-case="uppercase"/>
  </name>
</names>
```

And:

```xml
<names variable="author">
  <name>
    <name-type name="personal" form="short">
      <name-part name="family" text-case="uppercase"/>
    </name-type>
  </name>
</names>
```

bwiernik:
By adding cs:abbreviation, that order can be handled by placing abbreviation first, then using the `if-institution-short="true"` attribute on name parts:

```
<abbreviation/>
<name>
  <name-type="institution">
    <name-parts name="institution" if-institution-short="true" prefix=" (" suffix=")"/>
  </name-type>
</name>
```

bwiernik:
As for “henceforth”, either that’s hardcoded in prefix or https://github.com/citation-style-language/schema/issues/240

bdarcus:
So is this RTM?

schemas/styles/csl.rnc
denismaier: Any comments about this question?

denismaier:
> So is this RTM?

Wait a moment. I was wondering about `name-part` and `name-type` being siblings (see above)?

bdarcus:
Probably the best thing to do when you want to hold a PR is to click "request changes" on the review. That will clearly flag it to the maintainer that makes merge decisions.

Just make sure to approve it when done, so you clear the hold.

bdarcus:
Also, there is a merge conflict that needs fixing.

bwiernik:
The idea is that name-type is optional—styles without it (including all existing) just treat both types of names the same. Attributes and elements that aren’t specified in name-type are inherited from the parent name element. This is similar to how name-part already works—it inherits form and text styling attributes from name.

bdarcus:
Also, these are kinds of changes that we need tests for, so if you could find the time for a PR on the test-suite, that would be awesome.

bwiernik:
@bdarcus RTM?

Regarding tests, before we open 1.1 up for broad comment, I think we really should have spec language/examples and/or tests illustrating behavior available. If we solicit comments with just the schema, I fear that's going to lead to a lot of confusion and negative reactions.

denismaier:
> Regarding tests, before we open 1.1 up for broad comment, I think we really should have spec language/examples and/or tests illustrating behavior available. If we solicit comments with just the schema, I fear that's going to lead to a lot of confusion and negative reactions.

Absolutely. And a provisional changelog would also be a good idea.

bdarcus:
Agreed on the test and changelog points.

There's still a merge conflict on the PR though.

PR #325, opened 2020-07-20
input: Convert locator to an object array

bdarcus:
## Description

CSL input needs to accommodate possibility of multiple locators. Rather than requiring processors to parse a single string, this adds the structures to properly represent these data.

Example, in YAML

``` yaml
locators: 
  - 
    page: 23
  - 
    line: 12
```

This was easy enough to throw together, so I thought I'd post it for comment.

I also added a version property while I was at it.

Do you see any problem with doing this @fbennett?

## Type of change

### DIFF
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -1,110 +1,152 @@
 {
-    "description": "JSON schema for CSL citation objects",
-    "$schema" : "http://json-schema.org/draft-07/schema#",
-    "$id": "https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json",
-    "type": "object",
-    "properties": {
-        "schema": {
-            "type": "string",
-            "enum" : [
-                "https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json" 
-            ] 
-        },
-        "citationID": {
-            "type": [
-                "string",
-                "number" 
-            ]
-        },
-        "citationItems": {
+  "description": "JSON schema for CSL citation objects",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json",
+  "type": "object",
+  "properties": {
+    "schema": {
+      "type": "string",
+      "enum": [
+        "https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json"
+      ]
+    },
+    "citationID": {
+      "type": ["string", "number"]
+    },
+    "citationItems": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "id": {
+            "type": ["string", "number"]
+          },
+          "itemData": {
+            "$ref": "csl-data.json#/items"
+          },
+          "prefix": {
+            "type": "string"
+          },
+          "suffix": {
+            "type": "string"
+          },
+          "locators": {
             "type": "array",
             "items": {
-                "type": "object",
-                "properties": {
-                    "id": {
-                        "type": [
-                            "string",
-                            "number" 
-                        ]
-                    },
-                    "itemData": {
-                        "$ref": "csl-data.json#/items" 
-                    },
-                    "prefix": {
-                        "type": "string" 
-                    },
-                    "suffix": {
-                        "type": "string" 
-                    },
-                    "locator": {
-                        "type": "string" 
-                    },
-                    "label": {
-                        "type": "string",
-                        "enum": [
-                            "appendix",
-                            "article",
-                            "book",
-                            "canon",
-                            "chapter",
-                            "column",
-                            "elocation",
-                            "equation",
-                            "figure",
-                            "folio",
-                            "issue",
-                            "line",
-                            "note",
-                            "opus",
-                            "page",
-                            "paragraph",
-                            "part",
-                            "rule",
-                            "section",
-                            "sub verbo",
-                            "supplement",
-                            "table",
-                            "timestamp",
-                            "title",
-                            "verse",
-                            "volume" 
-                        ] 
-                    },
-                    "suppress-author": {
-                        "type": [
-                            "string",
-                            "number",
-                            "boolean" 
-                        ] 
-                    },
-                    "author-only": {
-                        "type": [
-                            "string",
-                            "number",
-                            "boolean" 
-                        ] 
-                    },
-                    "uris": {
-                        "type": "array",
-                        "items": {
-                            "type": "string" 
-                        } 
-                    } 
-                },
-                "required": ["id"],
-                "additionalProperties" : false 
-            } 
+              "$ref": "#/definitions/locators"
+            }
+          },
+          "suppress-author": {
+            "type": ["string", "number", "boolean"]
+          },
+          "author-only": {
+            "type": ["string", "number", "boolean"]
+          },
+          "uris": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          }
         },
-        "properties": {
-            "type": "object",
-            "properties": {
-                "noteIndex": {
-                    "type": "number" 
-                } 
-            },
-            "additionalProperties" : false 
-        } 
+        "required": ["id"],
+        "additionalProperties": false
+      }
     },
-    "required": ["schema", "citationID"],
-    "additionalProperties" : false 
+    "properties": {
+      "type": "object",
+      "properties": {
+        "noteIndex": {
+          "type": "number"
+        }
+      },
+      "additionalProperties": false
+    }
+  },
+  "required": ["schema", "citationID"],
+  "additionalProperties": false,
+  "definitions": {
+    "locators": {
+      "properties": {
+        "appendix": {
+          "type": "string"
+        },
+        "article": {
+          "type": "string"
+        },
+        "book": {
+          "type": "string"
+        },
+        "canon": {
+          "type": "string"
+        },
+        "chapter": {
+          "type": "string"
+        },
+        "column": {
+          "type": "string"
+        },
+        "elocation": {
+          "type": "string"
+        },
+        "equation": {
+          "type": "string"
+        },
+        "figure": {
+          "type": "string"
+        },
+        "folio": {
+          "type": "string"
+        },
+        "issue": {
+          "type": "string"
+        },
+        "line": {
+          "type": "string"
+        },
+        "note": {
+          "type": "string"
+        },
+        "opus": {
+          "type": "string"
+        },
+        "page": {
+          "type": "string"
+        },
+        "paragraph": {
+          "type": "string"
+        },
+        "part": {
+          "type": "string"
+        },
+        "rule": {
+          "type": "string"
+        },
+        "section": {
+          "type": "string"
+        },
+        "sub-verbo": {
+          "type": "string"
+        },
+        "supplement": {
+          "type": "string"
+        },
+        "table": {
+          "type": "string"
+        },
+        "timestamp": {
+          "type": "string"
+        },
+        "title": {
+          "type": "string"
+        },
+        "verse": {
+          "type": "string"
+        },
+        "volume": {
+          "type": "string"
+        }
+      }
+    }
+  }
 }
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -10,6 +10,10 @@
         "https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json"
       ]
     },
+    "version": {
+      "type": "string",
+      "pattern": "1.1"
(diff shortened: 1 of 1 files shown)


bwiernik:
Haven't looked at the exact code, but I support converting locator and page to an array. Been on my list.

denismaier:
Yes, better support for multiple locators is needed so this is a good change. How will this support mixing locators with narrative text in the suffix?

bwiernik:
> How will this support mixing locators with narrative text in the suffix?

No change. The full locator renders before the suffix.

bdarcus:
OK, I'll test it a bit more and then merge.

bdarcus:
Just tested, and this is working correctly.

``` js
{
  "version": "1.1",
  "citationID": "1",
  "schema": "https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json",
  "citationItems": [
    {
      "id": "ITEM-1",
      "locators": [
        {
          "page": "23"
        },
        {
          "line": "4"
        }
      ]
    }
  ],
  "properties": {
    "noteIndex": 1
  }
}
```

bdarcus:
Aside: the schema id is required. That could be a way to version the files, but for now I think the redundancy is valuable, in part because here I'm using "latest."

Not sure though; worth thinking about as we head towards 1.1 release.

Any thoughts on this specific point @dhimmel: versioning the json files? Use the id, or a new version property?

bdarcus:
One other little thing: I changed the name from `locator` to `locators`. Is this OK?

PaulStanley:
Fine, but we need to understand the semantics of this. What does a list of multiple locators *mean*, and how are the to be processed? Do multiple locators represent

(a) a flat list (so (page 1), (page 23) become "pages 1, 23".
(b) a hierarchical structure (page 1), (line 3)? And in this case how does one represent page 1 line 23 -- page 3 line 5?
(c) something else ...

bdarcus:
Yeah, good question @PaulStanley.

Right now, the assumption is A, in large part because when there are multiple locators, they refer to the same thing, so that there's no need for hierarchy beyond this.

Except ... you're right; it wouldn't work for your last example on B. It would not support multiple locations within a cited item (which maybe speakers in favor of calling it 'locator' if we keep this as is).

Do any of you expect that should be possible?

If yes, we'd want to modify this.

bdarcus:
I was basing this on how I understand [pandoc implemented locators](https://pandoc.org/MANUAL.html#citations), though I need to look at the code.

bwiernik:
(citeproc-js has essentially the same locator parsing behavior as pandoc-citeproc and citeproc-rs.)

> in large part because when there are multiple locators, they refer to the same thing

Even if they are referring to multiple things (e.g., "section 2, line 2, section 3, line 18"), we might not need for processors to do any sort of complex parsing of hierarchical arrangements by processors I don't think. The array is ordered, so we can rely on the user to structure the locators correctly. Just display the label and value in the order given. Collapse the labels for locators of the same type that are contiguous. 

### Multiple locators of the same type
Handling of multiple locators of the same type in one element is a bit of a tricky issue. I think that for practical usability, allowing comma-delimited multiple locators of the same type (especially for pages, but we don't need special treatment) is a good idea. 

The main place where this is important for the processor is pluralization of the labels. For pages, comma-delimited numbers generally works fine. For something like volume or supplement though, they might be identified by a title (e.g., vol. "Social, Personality, and Developmental Psychology"). citeproc-js's current rule it applies whenever it renders a label in a style (in locator, but also when, say `page` or `volume` is rendered) is to treat comma-delimited values as plural if they are numeric, but not if they are non-numeric. I think that's a reasonable rule to implement across all style processing. If "Social, Personality, and Developmental Psychology" are three volumes, this can be made clear easily by providing three `volume` elements in the array.

### Hierarchical locator formatting
One other complication is that some styles want specific formatting for "sets" of locators. E.g., [apparently Chicago style](https://forums.zotero.org/discussion/comment/360314/#Comment_360314) wants volume-page combinations to be formatted as "vol. 2:323" rather than "vol. 2, p. 323". A legal style might want section-paragraph-line combinations to be formatted in `§2, ¶4: 10–15; §4, ¶1: 15`

If we wanted to support that sort of formatting (not that we necessarily have to), we would need to provide for a way to styles to specify such formatting and for processors to understand how locators are arranged. For the latter, I see a few options:
  1. Assume each set of locators refers to only one place. Allow `locators` to be nested with multiple arrays, each array referring to one place.
  1. Assume each set of locators refers to only one place. Allow multiple `locators` elements on a citation, each array referring to one place.
  1. Allow multiple places to be referred to in one array. Specify a set of hierarchical relationships. For example:
        - `volume < chapter < page < line`
        - `supplement < section < paragraph < line`
      If an element is lower on the hierarchy, it refers to the same place. If it's higher on the hierarchy or not part of the hierarchy, it refers to a new place.

Of those, I think multiple `locators` elements is probably the easiest to create a user interface for, either in a text environment like pandoc or in a GUI like Zotero. If we decide to support this, we can work out style syntax.

bdarcus:
So what's your bottom-line @bwiernik - we should be fine with this for 1.1?

denismaier:
I'm contemplating whether we should now also consider adding support for mixing locators with narrative text. I'm thinking of something like this:

```
{ "suffix": 
  [
    ["string", "e.g."],
    ["chapter", "19"],
    ["string", "especially"],
    ["page", "581"]
  ]
}.
```

cormacrelf:
There's something to that structured approach, but I think you need to design it in concert with a system for formatting. I didn't progress the multi locators thing at the time because I wanted to make a `<locators>` element to replace `<text/number variable="locator">` and see how it worked. Thinking about it now, this is roughly what I was thinking:

```xml
<locators delimiter="; " and="text">
  <!-- each locator is rendered with one of these -->
  <locator match="volume page">
    <!-- custom; introduce loc element to render one specific locator's content (still have label available in here) -->
    <loc type="volume" suffix=":" />
    <loc type="page" />
  </locator>
  <locator delimiter=", ">
    <!-- default; here's what you put in each comma-separated locator atom, whose content goes in the loc element -->
    <label variable="locator" suffix=" " strip-periods="true" />
    <loc/>
  <locator>
</locators>
```

It's purely a game of labels and delimiters, and to do it, you need the groups pre-formed. The styles themselves know what the hierarchy is, so just give the processor the groupings.

If you wanted to express that two specific bits of a piece of legislation are relevant, you could do this:

```json
{ "locators": [[["part", "2"], ["division", "7, 9"]], ["schedule", "4"]] }
```

```txt
pt 2, divs 7, 9; sch 4
```

Notice the extra square braces grouping the first two together. Grouping the pieces together is really a user interface problem after all, which is best presented to the user as they are writing the locators in.

----

Side note, I'm back. Hello all.

bdarcus:
Welcome back @cormacrelf!

> { "locators": [[["part", "2"], ["division", "7, 9"]], ["schedule", "4"]] }

So this PR had locators as a single-level array of locator objects.

Your's above is an arbitrarily nested array of arrays?

Why is the fundamental unit an array `["part", "2"]` rather than an object `{"part": "2"}`, which seems more intuitive, easier to validate, etc?

cormacrelf:
I don't think it can or should be arbitrarily nested. If you need arbitrary nesting to express the pinpoint of your citation, then you should probably draw a diagram instead... or use a freeform suffix to describe it. I don't think any style guides would be contemplating arbitrary trees. So it's a two-level array.

And there's no logic behind the fundamental unit as I recall. The objects look better in YAML and are probably easier to read when debugging, other than that nothing much different.

```yaml
locators-object:
  - chapter: 17
  - page: 291
locators-array:
  - 
    - chapter
    - 17
  -
    - page
    - 291
```

Side note for me [using serde](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=23d362f900c66d931bf3d23dc05e2d3c)

denismaier:
> ```yaml
> locators-object:
>   - chapter: 17
>   - page: 291
> locators-array:
>   - 
>     - chapter
>     - 17
>   -
>     - page
>     - 291
> ```

@cormacrelf So this should actually quite easily allow what I was suggesting above, i.e. interspersing narrative text with locators. Just like so:

```yaml
suffix-object:
  - string: "e.g."
  - chapter: 17
  - page: 291
suffix-array:
  - 
    - string
    - "e.g."
  - 
    - chapter
    - 17
  -
    - page
    - 291
```

What do you think?

bdarcus:
> And there's no logic behind the fundamental unit as I recall. The objects look better in YAML and are probably easier to read when debugging, other than that nothing much different.

So functionality this is no different than this merge PR.

The advantage, though, is we can actually validate it. E.g. if someone did, a validator could flag an error.

``` yaml
locators-object:
  - chapt: 17
```

Oh, to be clear, because it's array of objects, Paul's example is actually valid.

``` json
[{
	"page": "1"
}, {
	"page": "2"
}]
```

Not sure we really want that, but it's not wrong from the JSON perspective.

denismaier:
> Thinking about it now, this is roughly what I was thinking:
> 
> ```
> <locators delimiter="; " and="text">
>   
>   <locator match="volume page">
>    ...
>   </locator>
>   <locator delimiter=", ">
>    ...
>   <locator>
> </locators>
> ```

Could that be an element `cs:locators` sibling to `cs:citation`, `cs:bibliography`? Or would you call this under `cs:bibliography`, `cs:citation`?

bdarcus:
I mean, at this point, this is a merged PR. So the question to resolve is, do we need one or more additional PRs, to amend this?

Issues which I see that could be addressed separately:

1. Add more hierarchy. But from what I can gather, nobody is arguing we should do this. If we did, that change would be an additional array within the top array.
2. Add (rich text) string as an object option

Changing object to array is another, but I don't think @cormacrelf is suggesting we do that.

@denismaier - I'm agnostic on your suggestion on 2. I get a little nervous about the UX implications for GUI apps, but as I consistently say, that's not our concern. And adding a string object is straightforward from a technical perspective.

denismaier:
> @denismaier - I'm agnostic on your suggestion on 2. I get a little nervous about the UX implications for GUI apps, but as I consistently say, that's not our concern. And adding a string object is straightforward from a technical perspective.

Well, this change here will already have some profound UX implications for GUI apps. So adding strings to the object should be actually rather trivial, I hope...

denismaier:
> * Add additional validation logic to only allow a single given locator object within the array. I _think_ this could be done. Should we?

Is that really useful? What about this?

```yaml
locators-object:
  - chapter: 17
  - page: 291
  - chapter: 18
  - page: 205
```

fbennett:
If you're validating these, numeric values might be a series or a range, and the values themselves might have an alpha prefix, or be expressed as roman numerals.

bdarcus:
> If you're validating these, numeric values might be a series or a range, and the values themselves might have an alpha prefix, or be expressed as roman numerals.

Good point @fbennett.

Currently they're just defined as strings.

Do we want to add something tighter? If yes, what?

We can use the `pattern` property with a regular expression if we like, for example.

fbennett:
Range collapsing applies, so you need to know whether things are numeric and collapse-compatible on both sides of a range delimiter. You could express that with an array of objects for each element in the array, with fields on each object for a character prefix, if any, and for joining punctuation on the second and subsequent elements, and wrapping the array in an object that supplies the label. A UI with all of those selections would be cumbersome to use, so the end result would likely be to push the task of parsing the string input entered by users upstream to the client. It seems likely that CSL implementations that aim to serve a wider range of projects would implement string parsing themselves, and flatten any structured input they receive to their own syntax. Defining the structures expressible in locators is a good thing; it's just that (I think) few clients will have the spare dev capacity to generate input that validates directly against the schema. It's quite a high hurdle.

bdarcus:
So you're saying keep it a string?

fbennett:
Oh, no. It's good to have a schema that defines exactly what structures are possible. That thought cuts the other way, I think: if this isn't treated as a mandatory input format, but as an abstract expression of structures that a processor should recognize and handle (whatever the form of the input), it adds a lot to stability. Not sure if that's helpful, but I didn't mean to suggest that casting a schema for locators is futile.

bdarcus:
> if this isn't treated as a mandatory input format, but as an abstract expression of structures that a processor should recognize and handle

I'm thinking of it more as the second, but that ideally people (developers building apps and tools users maintaining data themselves, say in YAML, etc.) could use it for the first if they choose.

I'm terrible with regular expressions. Anyone want to suggest one to include?

bdarcus:
> That could possibly speak in favor of an additional array level?

This is what I was meaning as the alternative. Yes.

Alternately, we'd want to amend the description on the schema to make clear that each object should describe one location, so that we didn't have this:

``` yaml
locators:
  - chapter: 17
  - page: 291
  - chapter: 17
  - page: 292
```

Issue #94, opened 2012-05-03
Expand number of locator terms


rmzelle:
See http://forums.zotero.org/discussion/14717/universal-locator-type/

In addition to just expanding the number of locator terms, we could also add an empty-string locator type, so that the user can store the label in the locator variable field.

Regardless of the solution chosen, we probably should make an inventory of missing locator types.

Identified so far:
Plays: Act I, scene i, lines 12-23 (or act, scene, verse)
Ancient sources: Book 1, lines 1-8; Book 6, chapters 57‐58
Bibles: Psalm 86:5; 1 Cor. 13:1 New International Version (Book Chapter:Verse, Translation)
Legal: article

http://www.longwoodshakespeare.org/quoteverse.html
http://blog.apastyle.org/apastyle/2009/12/happy-holiday-citing-citation-of-classical-works.html
http://www.skidmore.edu/classics/courses/2006fall/ssp100/citations.pdf

---

See also the discussion at http://xbiblio-devel.2463403.n2.nabble.com/Improving-support-for-locators-td7578007.html

denismaier:
Anyone knows where we are here? @bdarcus @bwiernik ? We have added some additional locator variables in https://github.com/citation-style-language/schema/pull/201, but not those mentioned here. Also: We haven't considered multiple locators yet, right?

bwiernik:
Multiple locators we considered—they can be entered as a comma delimited list with the supplying labels or their abbreviations indicating a change in type (same as we discussed with page). I haven’t had a chance to prepare a PR for that.

Given that we are doing that, I agree with making the label key optional (if there, it applies to the first locator in the list if it doesn’t have a label). 

act, scene, line, and verse sound good for adding. Didn’t think of those. 

For biblical verses and the like, that could be handled with a label-less locator or a citation suffix. “Psalms” isn’t a locator label, but part of the value.

bdarcus:
If I haven't mentioned it already, do look at how pandoc does this.

denismaier:
> act, scene, line, and verse sound good for adding. Didn’t think of those.

So it's act and scene. Line and verse are already there.

bwiernik:
> If I haven't mentioned it already, do look at how pandoc does this.

https://github.com/jgm/pandoc-citeproc/blob/b05d286b15b3545faf477301267e1a12be192e87/src/Text/CSL/Pandoc.hs#L472

bdarcus:
Also:

[...]

PR #326, opened 2020-07-22
input: Clarify modelling of citation locators

bdarcus:
This expands and clarifies the modeling of locator data.

A "locators" array consists of one-or-more locator objects.

Each locator object consists of one-or-more properties (page, line, 
etc), the value of each of which is a number or (for cases like 
"VII") a string.

In addition, adds a simple begin/end range object.

Finally, it adds a sub-array, for a secondary locator group.

-----

## Background

The initial change to locators in 1.1 left the individual locator object values as strings:

``` js
"locators: [
  {"page":  "1, 2, 3-5"}
]
```
But, this can also be represented as:

``` js
"locators: [
  {"page":  "1"},
  {"page":  "2"},
  {"page":  "3-5"}
]
```

Either way, this is by far the most common scenario, with page locators.

The values are easy enough to parse (exceptions being pages like "A-23"), but aside from the fact one can represent the same thing in multiple ways, more complex cases (per below, compound ranges, non-integer numbers, etc.) present additional challenges.

## Requirements

We _must_ support the following features, in rough order of priority and commonality:

1. **simple locators**  (1, 3, 5)
2. **simple ranges** (23-35)
3. **compound locators** (p2, line 4)
4. **compound ranges** (p2, line 12-14/p4, line 2)
5.  **grouping** (Doe, 2018, p45; see also p54)

## Not Supported

6. ability to modify representation of lists on an item-by-item basis (in some places, in the same document, "p1, 2", and in other places "p1, p2")

## Proposal

Locator objects represent a specific simple or compound location within the cited item. 

Each locator can have one-or-more locator properties (page, line, etc), the value of each of which is one of these:
  
- a number (integer or float)
- a string, for numbers that are not integers or floats

In addition, we add a simple begin/end range object, and a "sub-group" array (for 5).

The above covers all requirements.

Note: earlier iterations of this represented the range with a two-item array, which would have a problem noted below.

## Examples

**1; The simplest, most common, locator**

``` js
"locators": [{"page": 2 }]
```

**1/2**

``` js
 "locators": [
  { "page": 23 },
  { "begin": { "page": 25}, "end": { "page": 28}}
]
```

**1/2 with strings**

``` js
 "locators": [
  { "page": 23 },
  { "begin": { "page": "a-10"}, "end": { "page": "a-15"}}
]
```
**3**

``` js
 "locators": [
  { "page": 259, "line": 12}
]
```
 **4**

``` js
 "locators": [
  {"begin": { "page": 259, "line": 12}}, 
  {"end": { "page": 273, "line": 3}}
]
```

 **5**

``` js
 "locators": [
  { "page": 259, "line": 12}, 
  [
    "see also",
    {"page": 273, "line": 3}
  ]
]
```

### Parsing Strings

Possible string parsing rules; ideally to add [here](https://github.com/citation-style-language/data-parsing):

1. **simple locators** are delimited by commas (1, 3, 5)
2. **simple ranges** are delimited by hyphens (23-35); note this can be brittle in some circumstances, but could be addressed by quoting such sub-strings
3. **compound locators** (p2, line 4) are delimited by semi-colons (p2, line 12; p4, line 2)
4. **compound ranges** are, as with EDTF, delimited with a forward-slash (p2, line 12-14/p4, line 2)
5. **grouping** not sure, as we already use semi-colons for 3; maybe use colons for range, and slash here?

Also, labels precede the value for locators, but pages are default.

## Final Point

At least aspects of whatever we decide here should probably apply also to other numbers, particularly pages for article and chapters and such, in the main data schema.

The range-object approach above, for example, suggests modifying the current 1.1 data representation; so maybe we just have a general "range" property definition, that gets used with both numbers and dates.


bwiernik:
Representing this in the data input is trivial in my opinion. The question is how common is this need to actual citation styles and what would an appropriate style syntax be. If we can’t nail that down, i don’t think we support this. 

Cc @adam3smtih

bwiernik:
Representing this in the data input is trivial in my opinion. The question is how common is this need to actual citation styles and what would an appropriate style syntax be. If we can’t nail that down, i don’t think we support this. 

Cc @adam3smtih

bdarcus:
> The question is how common is this need to actual citation styles and what would an appropriate style syntax be. If we can’t nail that down, i don’t think we support this.

I agree, but if we don't, we should use the opportunity to clarify the modeling, in both input and styling schema.

E.g. either we support a single locator (location within some item), or a list of locators. 

This goes to Paul's question about the semantics.

I wonder where we should look for styles examples?

bdarcus:
Separate question: @denismaier's suggestion about allowing a string?

``` yaml
    - foo
    - page: 291
```

denismaier:
Obviously, I'd be in favour.

bdarcus:
> Obviously, I'd be in favour.

Could you post a couple of examples here of where this would be used, so people here can see, and if agreed, we can include it in the embedded documentation?

I used "foo" in the example, because I couldn't think of one.

denismaier:
What about this:

```yaml
- string: "e.g."
- page: 291
- string: "among others"
- page: 291
- string: "especially"
- page: 292
```

Or this?

```yaml
- page: 291
- string: "and throughout"
- chapter: 5
```

bdarcus:
The "e.g." would be a suffix; right?

denismaier:
> The "e.g." would be a suffix; right? Or no?

Don't think so. In Zotero a suffix appears after the complete citation: (Doe 2018: p. 291. SUFFIX).
In pandoc everthing after "Doe 2018" is the suffix, with the locators being parsed out. My idea was that we could switch to a more pandoc-style (or biblatex-style) model: prefix, citation, suffix. The suffix then getting parsed in a sort of AST.

PaulStanley:
There is a problem with the Bibllatex suffix model. Some citations (especially legal) need the locators embedded in the citation

>  Fairclaim v Shamtitle, 231 F.3d 232, *299* (2013)

When I wrote a legal citation package for Biblatex, I had to go to rather tedious lengths to get that to work. The current model where the locator becomes an integral part of the citation data is actually sounder.

(OTOH the Biblatex idea that the default *label* for a locator can be set in the cite data is better than the insistence on having it set explicitly. But that's a different story ...)

georgd:
The locator needs to be independently movable when it comes to citation
formatting. All three positions can be found: at the beginning,  in the
middle and at the end of the citation. (Juridic citations might even
require a second locator for loose leaf publications, CSLm has an extension
for this.)

But I don't think that this proposal wants to bind the locator to the
suffix in the formatted citation. The reference to pandoc rather points to
input formatting?

PaulStanley <notifications@github.com> schrieb am Fr., 24. Juli 2020, 23:36:

bdarcus:
> The current model where the locator becomes an integral part of the citation data is actually sounder.

Which "current model" are you referring to @PaulStanley; biblatex, or CSL?

And do you have a view on the locator vs locators question?

> But I don't think that this proposal wants to bind the locator to the suffix in the formatted citation. The reference to pandoc rather points to input formatting?

This PR is only focused on input formatting. But @bwiernik rightly points out that it has implications for the styling end too.

PaulStanley:
I was only responding to what seemed to be an idea that one could more or less dispense with locators and just deal with it in the cite suffix. I may have got the wrong end of the stick.

As far as the proposal goes, I'm neutral really. I'd be inclined to keep it as simple as possible however because (a) locators are by definition "one cite" objects, so elaborate data structures may be counterproductive and (b) they can get very complex and hard to define, so one is always going to have to allow for the possibility that the user is just going to have to enter them manually as a string. But that's not a reason to oppose any reasonable proposal to structure them. A flat list sounds like the sweet spot, but it would need to allow for ranges.

denismaier:
My proposal to include strings and to parse that from a single suffix field wasn't intended to mean we should restrict everything to a single suffix field. The idea was that a suffix like "especially p. 39, and many other locations in ch. 5" will be parsed into something like this:

```yaml
suffix:
  string: "especially"
  page: 39
  string: "and many other locations in"
  chapter: 5 
```

Now, the locators should be fully accessible to style. But maybe this is more complicated than I initially thought?

denismaier:
On the inital PR here: we should indeed keep this as simple as possible. If an ordinary flat list will work, I'm absolutely ok with that. No need to make things more complicated than necessary. Just wondering if we can get something like "ch. 5, p. 290; ch. 6; p. 300" from a flat list.

bdarcus:
My assumption was no.

>

bdarcus:
But maybe if it's just a flat list, all that a processor is doing is
handling the labels?

In which case it could be fine?

PaulStanley:
I may be making some mistake. But I think it's more or less possible:

I'm going to use four examples:

A. pp 12, 13-14, 23 (multiple locators with a range)
B. p 12, col 4; p 13, col 5 (multiple locators with doubled key)
C. p 12, cols 3-7; p 13 (multiple locators with a range)
D. p 12, col 3-p 13, col 8; p 20 col 10 (ranged double key)

Now suppose one represents A with a list, using a nested list to represent ranges

```
locators:
  page: 12
  page: [13, 14]
  page: 23
```

That looks reasonable, and deals with the most common case.

How do we do B, then?

```
locators:
  page: 12
  col: 4
  page: 13
  col: 5
```

The trouble is that the boundary between (page/col) pairs and the boundary within page/col pairs shouldn't have the same "weight". Still we can suggest a reasonable rule. Give it the same weight iff the labels are the same (Example A), but not if they are different (Example B), i.e. treat the first locator label given as marking the main separation. That works.

Combining the two rules does indeed let us deal with example C.

```
locators:
  page: 12
  col: [3, 7]
  page: 13
```

The tricky one is Example D. In theory I think it's possible to do it if one allowed not only lists of references (as in Example A and C) but lists of dictionaries too. One could then have (no idea if this is legal YAML, but it's perfectly reasonable so I guess it should be) ...

```
locators:
  [page: 12, col: 3, page: 13, col 8]
  page 20
  col 10
```

(One can regard C as a sugared form of D
```
page: 12
[col: 3, col 7]
page: 13
```

Our rules are then these:

* A locator consists of a flat list.
* Each member of that list takes one of three forms:
  + a dictionary pair consisting of a label and a string -- `page: 12`
  + a dictionary pair consisting of a label and a list of two strings -- `page: [12, 13]`
  + a list of two pairs of labels and strings
* The outer list is interpreted as a list of locators. The first locator is treated as the primary locator. When a list is output:
  + if it consists only of primary locators then the `basic-separator` is put between them
  + if it consists of a mixture of primary and secondary locators then a `primary-separator` is put between primaries, and all other locators are separated by a `secondary-separator`. (Usually the `secondary-separator` is the same as the `basic-separator` and the `primary-separator` is something "bigger".) So, e.g. with `basic-separator` = ", ", `primary-separator` = "; " and `secondary-separator` = ", "

> [page: 12, page: 13] => pp. 12, 13 (basic-separator)
> [page: 12, col: 13] => p. 12, col. 13 (secondary-separator)
> [page: 12, col: 13, page: 15] => p. 12, col. 13; p. 15 (secondary-separator, primary-separator)
* Inner lists are treated as ranges. A label followed by a list is treated as if it were a list containing identical labels, i.e.
  `col: [12, 13]` is equivalent to `[col: 12, col:13]`. In either case, what is produced is a range, separated by a `range-delimiter`.
* Labels are "collapsed" when the next label would be identical, and pluralised if necessary

```
page: 12
page: 13 
``` 
> pp 12, 13 (not p 12, p 13).

But
```
page: 12
col: 13
page: 20
col: 10 
```
> p 12, col 13; p 20, col 10

* A sub-list is taken as a range, with the range marker placed immediately before the second appearance of whatever locator-type first appears in the list.

What you give up with this structure is the ability just to put a list of numbers after a single locator (because that gets treated as a ranged sublist). But I'd pay that price for the ability to specify ranges with precision.

bwiernik:
One thing that is hard for me to follow in this thread is that we aren’t talking about actual rendered citations, so it’s hard for me to be sure what format is trying to be accommodated. 

I’m working my way through the thread, but @denismaier it seems like most of those eg/and others/etc examples should be part of one page element. 

If you are trying to get “(Jones, 2010, p. 10 and throughout)”, that would be accommodated by entering page as “10 and throughout”. I don’t know that I’ve ever seen “e.g.,” prepended before a locator label. Is that common enough to bother with the big increase in complexity? My inclination would be that “eg, p. 10 and throughout” should just be entered entirely in the suffix field.

bwiernik:
> In pandoc everthing after "Doe 2018" is the suffix, with the locators being parsed out. 

I don’t think this is a good way to think of it. The structure of citations in both Zotero and pandoc is:

Prefix | Cite | Locator | Suffix

The parsing pandoc is doing is determining whether there is a locator or not. Locators follow strict text patterns—if the text doesn’t match one of those, it’s treated as a suffix, not a locator. 

I don’t see how a text-based system like pandoc would be able to distinguish a locator with arbitrary text strings from a simple suffix. Beyond that, I’m not really sure why that would be needed. As mentioned by Paul above, the major purpose of the locator field is to control style-specific formatting of the label and potentially placement. If the user is entering arbitrary strings, that’s not going to be really transportable across styles anyway. At that point, the user should just enter the text as a suffix.

bwiernik:
I don’t think we can give up entering a list or range of one type of locator in one field. That is by far the most common case in citations. 

Explicit specifications of a complex range could instead be had by having a `range` attribute that takes an array of two locators as values. 

I suggest these rules:
1. `locators` can contain one or more `locator`
1. Each `locator` list is understood as one point reference, with the elements being rendered in order. For example, “vol. 2, p. 26”. 
1. Elements in a `locator` list can contain multiple values that are parsed using the existing rules for numeric lists and ranges. These are all rendered in the same place—not parsed and broken into bits. For example, “vol. 2, pp. 26–34” or “p. 15, col. 2–4”. 
1. Simple ranges can be expressed as above. Complex ranges like Paul’s example D where labels need to be repeated can be expressed as by using a `range` attribute whose value is a length-2 array of `locator` elements:

```
locators:
  - range:
      - page: 5
         column: 2
      - page: 7
         column: 3
```    


To handle multiple locators (even just a flat list), we will need a locator-delimiter attribute (probably on the style). If we have that, then

PaulStanley:
That works. I marginally prefer it to my own suggestion because it makes the range *explicit* which I think is desirable (you are not "inferring the type" merely from the list. I had assumed that ranges were at least as common as lists of pinpoints, but that's really a pragmatic question. The short point is that you *either* need an explicit range type OR you can't have a pure list of the 

```
page: [1, 3, 10]
```

variety.

It's going to be sort of tempting to allow lists like
```
page: [1, 3, 4-7]
```

but I would resist the temptation, which takes us back to parsing, whereas i think this is a structure which should reflect the *result* of parsing at an earlier stage, if it needs to.

bwiernik:
Those don't need to be lists. They can just be:
`page: 1, 3, 10`
`page: 1, 3, 4-7`

PaulStanley:
Well they don't *need* to be, but it would really be better if they *were*. Otherwise we have to pre-parse every string in order to transform it into a different structure.

In other words, given

```
page: 1, 2, 3-4
```

We are going to have to parse this into a structure

```
page: 1
page: 2
range: 
   [page 3, page 4]
```
Anyway, and it would be kinda nice not to have to do that. Given that this is not likely to be a structure "written by hand" in any event ... But it's not such a big deal, so long as the requirements are clear.

bwiernik:
That is the existing behavior (so existing citeprocs already do it), and that sort of parsing is necessary anyway for rendering any numeric variable elsewhere in CSL. This is just applying the same number-parsing rules to a locator element as to any other variable in CSL. It’s consistent.

PaulStanley:
Yes ... but. If you are going to have a format which is (a) unlikely ever to be entered by an actual human being because it's already way complicated and (b) intended to be ambiguous, I'd far sooner avoid these sorts of fudges. The desire to present the processor with something that is neither raw material nor fully structured but in some indeterminate half-masticated state is, on balance, to be avoided.

bwiernik:
To be clear, from the perspective of the processor, `page: 1,3-4` is not a list at all. It is a single numeric variable which is formatted like all other numeric variables in CSL. There is no reason to disallow data entry that would be valid elsewhere on locator. That creates a weird exception that has to be accommodated.

PaulStanley:
I think I'd put it the other way round.

If this is a *good* structure for specifying *locations*, then it would be a good structure for specifying locations *wherever* they are needed, e.g. on `pages`, `volume` etc. And then there is an excellent reason to prefer it over the really rather half-baked and sometimes impossibly ambiguous way that we try to chop muddled text into number ranges, which leave us puzzling over "1-110-1-145" and the like, when

```
pages:
   range:
       page: 1-110
       page: 1-145
```

or

```
pages:
   range:
      page: 1-110-1
      page: 1-110-145
```

Would give us precision, if we needed it.

bdarcus:
Just wanted chime that this is a good discussion!

bwiernik:
That is a really big change from existing citeproc behavior. I don't see the benefit of placing a novel restriction on data when a well-functioning system is already in place. This is especially the case when most numeric ranges exist in wild data as flat variables (most notably page ranges). I think it's a much more accessible and workable change to expand the options into (1) allowing for an explicit range specification and (2) providing for a list of locators.

PaulStanley:
Number parsing and reformatting is actually one of the areas where the spec and expected behaviour are particularly difficult to fathom.

For instance consider this from the test suite [https://github.com/citation-style-language/test-suite/blob/master/processor-tests/humans/page_Expand.txt](url). 

> A. Input `110-5`. Output: 110–115 (treated as range)
> B. Input: `N110 - 5`. Output: N110-5 (treated as single page)
> C. Input: `N110 - N5`. Output: N110–N115 (treated as range)
> D. Input: `110 - N6`. Output 110-N6 (treated as single page)
> E. Input: ` N110 - P5`. Output: N110-P5 (treated as single page)
> F. Input: `123N110 - N5`. Output: 123N110-N5 (treated as single page)
> G. Input: `123N110 - N5, 456K200 - 99`. Output: 123N110-N5, 456K200-99 (treated as single pages)
> H. Input: `123N110 - N5, 000c23 - 22`. Output: 123N110-N5, 000c23-22

Some of these (A, E) seem obviously right. Some seem questionable. It seems to me that one might make the argument that B. should be "N110-N115" on the (reasonable) assumption that the prefix has simply been omitted from the second part of what is in fact a range, whereas the test assumes this is a single number (the same goes for the second example in G, which could perfecly reasonably be intended as 456K200–456K299). Some depend on making reasonable assumptions (e.g. that ranges must go up rather than down) . But it's not always clear what is being assumed. For instance, does F come out as it is only because 5 is less than 110 (in which case H simply repeats this test)? Or something else. And how, if it comes do that, do I know that 5 *is* less than 110, if I am expanding an *minimal* range? Couldn't the user have entered "110-5" meaning "110-115". To put the icing on the cake, test C assumes *exactly that*, and as it stands I just cannot see how Test C and Test F hang sensibly together.

So there is ambiguity aplenty, to put it nicely.

But you will struggle to find *any* of this in the spec, which simply says:

> If a number variable is rendered with cs:number and only contains numeric content (as determined by the rules for is-numeric, see Choose), the number(s) are extracted. Variable content is rendered “as is” when the variable contains any non-numeric content (e.g. “Special edition”).
> During the extraction, numbers separated by a hyphen are stripped of intervening spaces (“2 - 4” becomes “2-4”). Numbers separated by a comma receive one space after the comma (“2,3” and “2 , 3” become “2, 3”), while numbers separated by an ampersand receive one space before and one after the ampersand (“2&3” becomes “2 & 3”). [**This incidentally isn't actually what the tests assume will happen, because the general practice seems to be not simply to strip hyphens of their spaces but generally to change hyphens to en-dashes, I think, though this is a point of detail.]
> ...
> [`page-range-format`] Activates expansion or collapsing of page ranges: “chicago” (“321–28”), “expanded” (e.g. “321–328”), “minimal” (“321–8”), or “minimal-two” (“321–28”) (see also Appendix V - Page Range Formats). Delimits page ranges with the “page-range-delimiter” term (introduced with CSL 1.0.1, and defaults to an en-dash). If the attribute is not set, page ranges are rendered without reformatting.
> “expanded” Abbreviated page ranges are expanded to their non-abbreviated form: 42–45, 321–328, 2787–2816

So this is actually a classic example of a case where (1) the tests assume behaviour that the spec does not specify and (2) some of that behaviour is not *obviously* right.

The only really reliable way to deal with these cases is to start with pages in a canonical form (not made minimal!) with ranges clearly specified. Which is why one needs an input format which supports that. As usual, it may be reasonable to specify "simpler" formats, but really one should expect them to deal only with the simple cases.

Critically there needs to be a reliable way for the user to say "Please don't guess what I mean, *this* is what I mean. If the processor is going to parse

```
page: 10-13
```

as a range, the user loses the ability to specify unambiguously that it is **NOT** a range dammit, but a single location `10-13`.

bdarcus:
> So this is actually a classic example of a case where (1) the tests assume behaviour that the spec does not specify and (2) some of that behaviour is not _obviously_ right.

One way to think of the input data model PRs of late is I am attempting to make (more) explicit what in 1.0 is too often implicit.

There are good historical reasons for this, but with the experience we've since garnered, it seems we should be able tighten this up.

So let us summarize the issues here, with examples, in order of priority and commonality (using pages only, since most common):

1. a) single page location: (Doe, 2018:23), b) page location that is a range: (Doe, 2018:23-24).
2. page location that is a non-contiguous list: (Doe, 2018:23, 25-26).
3. page location with subpage details: (Doe, 2018, p23, l4).
4. 3 + 1/2? Is this the right way to express it? I guess this was my question for this: _should_ we support this?

Leaving aside whether we should do this, the fully structured expression of these should probably be:

1a

``` yaml
page: 23
```

1b

``` yaml
page:
  range:
    - page: 23
    - page: 24
```

2

``` yaml
page: 23
  range:
    - page: 25
    - page: 26
```

3

``` yaml
page: 23
line: 4
```

4

``` yaml
-
  page: 23
  line: 4
-
  page: 25
  line: 12
```

bwiernik:
I agree that the numeric spec could use clarifications or corrections. But I don't think it's reasonable to expect processors to not have to do any parsing of numbers.

A number like `1-3` is definitely the rare exception, and we shouldn't design the entire data and style system around that. It would be a huge burden on users and applications to prohibit entry of numbers as a flat field. 

If we want to include a way to explicitly state that `1-3` is not a range, we could provide an escape, such as a `{literal: "1-3"}` rich text markup field.

PaulStanley:
Yes, it's *reasonable* to expect some parsing (though I'd be inclined to say "pretty minimal parsing" because it only needs to deal with the typical/common case since there's a mechanism to hand for more complex cases). I'd need to think more about the relative merits of making the *parsing* explicit (`page: {raw: 1, 3-6, 7}`) or countermanding it. We'd also need to be clear, if we allow parsing, about what parsing is required and how deep it goes. If I get `page: 11, col 1 - 12, col 3` (i.e. something that is really nonsense, given the schema) what do I do?

bwiernik:
Here is my proposal:

1. `locator` and other numeric variables need to be parsed for the on-spec 1.0 behavior--applying range delimiters and pluralizing labels. So "p. 2, 4, 6-8".
2. Separating multiple `locator` elements is out of scope for a processor--that's in the realm of pre-processing as either a separate stage by the processor or by the calling application.
3. We should have a "literal" markup tag that prevents any parsing to accommodate `page: 2-7` with the hyphen being literal not a range. We can provide another recommended pre-processing syntax, such as `\-` for a literal hyphen or BibTeX-style `{literal text here}` to permit easy entry, but a processor should expect the explicit literal tag.

bdarcus:
> My proposal to include strings and to parse that from a single suffix field wasn't intended to mean we should restrict everything to a single suffix field. The idea was that a suffix like "especially p. 39, and many other locations in ch. 5" will be parsed into something like this:
> 
> ```yaml
> suffix:
>   string: "especially"
>   page: 39
>   string: "and many other locations in"
>   chapter: 5 
> ```
> 
> Now, the locators should be fully accessible to style. But maybe this is more complicated than I initially thought?
> 
> Edit: But essentially it looks like these strings I've had in mind are just affixes to otherwise ordinary locators, right?

I'd like more discussion of this, as it seems to have gotten lost.

It seems to me adding strings is merging`suffix` and `locator`.

Is this wise?

bdarcus:
> * Separating multiple `locator` elements is out of scope for a processor--that's in the realm of pre-processing as either a separate stage by the processor or by the calling application.

I don't understand. Can you clarify?

> * We should have a "literal" markup tag that prevents any parsing to accommodate `page: 2-7` with the hyphen being literal not a range. We can provide another recommended pre-processing syntax, such as `\-` for a literal hyphen or BibTeX-style `{literal text here}` to permit easy entry, but a processor should expect the explicit literal tag.

It strikes me we should add the bibtex convention to the [parsing suggestions](https://github.com/citation-style-language/data-parsing/), to be usable anywhere.

In the experimental [rich text schema](https://github.com/citation-style-language/schema/blob/v1.1/schemas/input/csl-rich-text.yaml), we have a `preserve` property that is designed for case currently, but maybe should, per bibtex, apply this case too?

bdarcus:
Another idea (just throwing it out there):

JSON does have string and int datatypes.

Combined with a structured representation, wouldn't this be perfectly clear to a processor what to do; where it views a string as a literal, and int as a number?

Maybe not, but I thought it worth asking ...

``` js
[
  {
    "page": [
      23,
      25
    ]
  },
  {
    "page": {
      "begin": "10-13",
      "end": "10-15"
    }
  }
]
```

denismaier:
> If the user is entering arbitrary strings, that’s not going to be localizable or transportable across styles anyway. At that point, the user should just enter the text as a suffix.

I'm not sure I agree here. Let's say you have this in a footnote:

> Smith, especially p. 12.

Now, you switch to a different footnote style that omits the label with `locator="page"`. If "especially p. 12" is a suffix, you cannot automatically do this. Having said that, you might be right so say that this does not justify the increase in complexity, especially since I'd consider "Smith, especially 12" not particularly readable style, but anyway.

denismaier:
> It seems to me adding strings is merging`suffix` and `locator`.
> 
> Is this wise?

I'm not sure ;-)

Actually, the goal of my suggestion was just to add a simple way to add strings in front of the locator, or between locators without having to resort to the suffix. Using the suffix means you won't have automatic labels and so. I thought parsing the content form a single suffix (on the input side) could be a workable solution. But perhaps there are other suggestions/ideas?

bwiernik:
I think a locator with a prefix would produce problematic output in a variety of situations (e.g., in styles that don't print the locator label like Vancouver), and they wouldn't be localized anyway. So automatic substitution of terms and localization would be a mess with a prefix. Given those limitations, I don't think it adds much value to offer the prefix-on-locator rather than requiring a user to use a suffix for that level of customization.

bdarcus:
> I think a locator with a prefix would produce problematic output in a variety of situations (e.g., in styles that don't print the locator label, "especially 2" doesn't look right), and they wouldn't be localized anyway. So automatic substitution of terms and localization would be a mess with a prefix. Given those limitations, I don't think it adds much value to offer the prefix-on-locator rather than requiring a user to use a suffix for that level of customization.

What if this referring to? Denis' suggestion to add string?

bdarcus:
Let's defer the string question to another PR. This is already complicated enough, and I'm unsure how to resolve it.

bdarcus:
Another idea:

``` js
  "locator": {"page": [21, 23, [32, 44]]}
```

``` js
  "locator": {"page": [21, {"line": [12, 15}]}
```

Locator is an array of numbers, or objects; range is represented by an inner array.

Isn't this more compact, and expressive?

Still doesn't answer the original question though.

bwiernik:
That dramatically increases the pre-processing burden for no real benefit. CSL processors are already expected to parse flat numeric variables to pluralize labels and set the range delimiters. If there exists a literal tag or similar to prevent normalization, I see no benefit at all in removing the ability to specify flat numeric variables. 

We have tests for how to parse numbers. If there is ambiguity or inconsistencies there, let's fix that, not throw out existing data expectations.

My proposal makes a minimal change that remains largely backwards compatible with existing processors and data generators. We don't need and should not, try to recreate all do the parts of CSL from square one.

PaulStanley:
> Another idea:
> ```
> "locator": {"page": [21, 23, [32, 44]]}
> "locator": {"page": [21, {"line": [12, 15}]}
> ```
>  Locator is an array of numbers, or objects; range is represented by an inner array.

This seems quite elegant to me at first glance.

@bwiernik says
>That dramatically increases the pre-processing burden for no real benefit. CSL processors are already expected to parse flat numeric variables to pluralize labels and set the range delimiters.

I think this overstates it. Taking a "flat" numeric variable and massaging it into this form is the very same task that CSL processors already do. So talk of "dramatically increasing the processing burden" overstates the burden. The advantage is that it provides a nice clear structured data which can avoid ambiguities that exist at present.

> My proposal makes a minimal change that remains largely backwards compatible with existing processors and data generators. 

Backwards compatibility is important. Suppose one said that if locator number part is just a *string* it will be parsed/normalized using the existing rules. If it is an array, it will be dealt with in this way. That gives full backwards compatibility.

bdarcus:
> Suppose one said that if locator number part is just a string it will be parsed/normalized using the existing rules. If it is an array, it will be dealt with in this way. That gives full backwards compatibility.

That's how we're dealing with dates in 1.1; a string is EDTF to be parsed, but can also use the structured representation, which is a representation of (a subset of) the EDTF.

schemas/input/csl-citation.json
```diff
@@ -70,85 +70,114 @@
...
-          "type": "string"
+          "$ref": "#/definitions/locator-variable"
         },
         "table": {
-          "type": "string"
+          "$ref": "#/definitions/locator-variable"
         },
         "timestamp": {
+          "description": "TODO",
           "type": "string"
```
bdarcus, line 171: What should a timestamp format be? Do we care?

If this is specific to a point in multimedia, then it seems we should use a structured format; maybe:

```suggestion
          "description: "An hour, minute, second timestamp for video or audio sources.",
          "type": "array",
          "examples": [
            [1, 12, 32]
          ]
```

bdarcus: On timestamps, based on a quick search, the big English language styles use `hour:minute` timestamps.

Also, notable: timestamps are also required as part of item metadata on some resources, like a last updated on an online article. EDTF supports times, we're covered there on the input end. But that raises the question again about the datatype; on an item it's part of an EDTF date. What about here?

What about the styling end? And do these conventions differ by language?

bdarcus:
Status:

- I still need to test this
- I have a comment about timestamps I could use input on, though part of that is about issues beyond this PR
- I think @PaulStanley's suggestion of allowing string OR array is a good compromise, and should probably be used everywhere a variable can be a number list. 
- Any comments on my string vs number suggestion here? E.g. rather than adding an explicit `literal` property, simply treat strings within the array as a literal?

schemas/input/csl-citation.json
```diff
@@ -70,85 +70,119 @@
...
+          "$ref": "#/definitions/locator-variable"
         },
         "table": {
-          "type": "string"
+          "$ref": "#/definitions/locator-variable"
         },
         "timestamp": {
-          "type": "string"
+          "description": "TODO",
+          "description: "An hour, minute, second timestamp for video or audio sources.",
```
bdarcus, line 171:
```suggestion
          "description": "An hour, minute, second timestamp for video or audio sources.",
```

bwiernik:
Re: arbitrary string locators, pandoc has this syntax to deal with that: https://github.com/citation-style-language/schema/issues/94#issuecomment-651167640

bwiernik:
@bdarcus I suggest changing version back to a string. CSLm for example, includes "mz" in its version numbers. If we adopted discrete extensions, I could similarly see those being string suffixes.

schemas/input/csl-citation.json
bwiernik: I commented in the other issue you opened, but I don't think I've ever seen an editing or retrieval _time_ be cited, only a date. For timestamps in `locator` or `dimensions` (running time), I think the `hh:mm:ss` format is pretty universal).

schemas/input/csl-citation.json
```diff
@@ -70,120 +70,114 @@
...
-            ],
-            [
-              {
-                "page": "34",
-                "line": "12"
+          "description": "When using the string ...",
+          "type": "string"
+        },
+        {
+          "description": "Strings represent a literal number, while numbers (integer or floating point) should otherwise be used. Ranges are represented by a two-item inner array.",
```
bwiernik, line 81: I don't understand what this description is saying.

schemas/input/csl-citation.json
```diff
@@ -70,120 +70,114 @@
   "required": ["schema", "citationID"],
   "additionalProperties": false,
   "definitions": {
-    "locators": {
-      "title": "Locators List",
-      "description": "A list of locators. Typically, there will only be one, but there are cases where there may be more than one.",
-      "examples": [
+    "locator-variable": {
+      "description": "A locator is a list, that can be a string, or a (preferred) array.",
```
bwiernik, line 74: With `locator-variable` being used to define both the whole array and the individual sub-elements, I think the "preferred" language is a little confusing. 

In the example `{ "page": [2, 3, [5, 12]] }`, `"page"` is an array with three elements, each defined to be a `locator-variable`. The elements are, in order, a number, a number, and an array with two child `locator-variable` elements, each of which is a number. Some more clarity is needed here.

georgd:
Brenton M. Wiernik <notifications@github.com> schrieb am Do., 30. Juli
2020, 19:59:

> *@bwiernik* commented on this pull request.
> ------------------------------
>
> In schemas/input/csl-citation.json
> <https://github.com/citation-style-language/schema/pull/326#discussion_r463173971>
> :
>
> >          },
>          "timestamp": {
> +          "description": "TODO",
>            "type": "string"
>
> I commented in the other issue you opened, but I don't think I've ever
> seen an editing or retrieval *time* be cited, only a date.
>
I know at least of one style which requires access date + time for online
resources.

schemas/input/csl-citation.json
```diff
@@ -70,86 +70,114 @@
...
+      "table": {
+        "$ref": "#/definitions/locator-variable"
+      },
+      "timestamp": {
+        "description": "An hour, minute, second timestamp for video or audio sources.",
+        "type": "string",
+        "pattern": "^[0-9]{2}:[0-9]{2}(:[0-9]{2})?$",
+        "examples": ["01:12:32"]
+      },
+      "title": {
```
bdarcus, line 173: Two technical issues ATM:

1. This is causing a conflict, and an error, because JSON schema has a reserved `title` property for annotations. If I move it out of the definitions area, it goes away, but then I don't know how to refer to it so it can be used within the array.

2. I haven't figured out how to remove the `locators` array.

bdarcus:
> CSLm marks its versions using "mz", and I could imagine us using string suffices to indicate extension versions (e.g., multilingual) in the future, so we might want to keep this as a string (but keep the stricter regex pattern).

Done.

schemas/input/csl-citation.json
bwiernik: "title" as in of a law ala "Civil Rights Act of 1964, Title VII". 

We could change it to "title-locator".

bwiernik:
Let’s change it to article-locator too. That will match the term name.

bdarcus:
> Let’s change it to article-locator too. That will match the term name.

Done.

Only thing I need to do to the CI to pass is fix the tests.

Final substantive issue:

So here's what we have now, which validates against this schema.

``` yaml
---
version: '1.1'
schema: https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json
citationID: 1
citationItems:
-
  id: 1
  itemData:
    type: book
    title: test
    issued: 2010-11
  locators:
    - page: [1, 2, {'line': 12}, [4,15]]
```

To me, having that as an array now seems excessive, as is even the `locators` property, and we could just do:

``` yaml
---
version: '1.1'
schema: https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json
citationID: 1
citationItems:
-
  id: 1
  itemData:
    type: book
    title: test
    issued: 2010-11
  page: [1, 2, {'line': 12}, [4,15]]
```

 ... or:

``` yaml
---
version: '1.1'
schema: https://resource.citationstyles.org/schema/latest/input/json/csl-citation.json
citationID: 1
citationItems:
-
  id: 1
  itemData:
    type: book
    title: test
    issued: 2010-11
  locators:
    page: [1, 2, {'line': 12}, [4,15]]
```

bwiernik:
We should definitely keep `locator` or `locators` (I'd lean toward `locator` given that it's the existing name).

I'm actually a little confused, though. Is `'1, 2, line 12, 4-15'` intended to mean "(page 1), (page 2 at line 12), (pages 4 through 15)"? If not, what is the "line" attached/subordinate to?

bwiernik:
Also, I am not sure that it's clear when an array is a list of discrete locators and when it's a range. If I want to cite "page 2 and page 5" and it is parsed into `page: [2, 5]`, that is going to be interpreted as a range "pages 2–5", no?

bwiernik:
To be sure we are covering bases, let's write out how each of these is represented.

- p. 26
- p. 5, 7
- p. 4–10
- p. 32, line 4
- section 2, paragraph 2, line 4
- {volume 2, p. 26}, {volume 3, p. 32}
- {p. 32, line 4}–{p. 34, line 2}
- {section "Principles"}, {section "Extensions"}, {section "Limitations"}

bdarcus:
> I'm actually a little confused, though. Is `'1, 2, line 12, 4-15'` intended to mean "(page 1), (page 2 at line 12), (pages 4 through 15)"? If not, what is the "line" attached/subordinate to?

It's not supposed to mean anything in particular; it's a string. Am just demonstrating that.

> I am not sure that it's clear when an array is a list of discrete locators and when it's a range. If I want to cite "page 2 and page 5" and it is parsed into page: [2, 5], that is going to be interpreted as a range "pages 2–5", no?

It's a range when it's a sub-array.

This is two pages:

``` yaml
page: [1, 5]
```

... while this is a range:

``` yaml
page: [[1, 5]]
```
That's how we're dealing with date ranges in the date object.

The alternative would be an internal range object.

``` yaml
page:
  begin: 1
  end: 5
```

bwiernik:
> It's not supposed to mean anything in particular; it's a string. Am just demonstrating that.

But here: `[1, 2, {'line': 12}, [4,15]]`. What is the line 12 subordinate to? It doesn't make sense just floating there.

bwiernik:
> It's a range when it's a sub-array.

That doesn't seem to how it's defined here. An array is defined to contain elements of class `locators`, which contain locator variables. 

There is nothing to stop:
```
page: [1, [2, [3, [4, 8]]]]
```

bdarcus:
> > It's not supposed to mean anything in particular; it's a string. Am just demonstrating that.
> 
> But here: `[1, 2, {'line': 12}, [4,15]]`. What is the line 12 subordinate to? It doesn't make sense just floating there.

Ok, gotcha.

The idea behind this is that the order within the array would determine that; so in this case, page 2. So a processor would just output it in order, adding the requisite label as needed.

I think the question is how complex we want to get for what might be reasonably argued are corner cases.

By far the most common will be a single page, after all.

bwiernik:
I am thinking that this nested array structure is introducing more complexity and ambiguity than it is solving.

This might be clearer:

1. `locators` is an array.
1. Each element is a `locator` object (containing page, volume, etc. elements, in order, potentially multiple of each ala `p. 2, line 1, p. 4, line 3`; these are all rendered in order by the processor) or an array of length two with `locator` object elements to indicate a range. 
1. Within `page` and other `locator` elements, contents can be a string (parsed for plural and range delimiters) or an array whose elements are (1) numbers, (2) unparsed strings, or an array of length 2 whose elements are numbers or unparsed strings (indicating a range).

bdarcus:
> I am thinking that this nested array structure is introducing more complexity and ambiguity than it is solving.

Maybe, but I _think_ it's solvable.

> * `locators` is an array.
> * Each element is a `locator` object (containing page, volume, etc. elements, in order, potentially multiple of each ala:

That's what the initial PR was ;-)

But that causes other problems, because what happens if you have a range of 20 pages?

And it's still the same basic idea (no true containment; just relying on order), except more verbose. 

And you then have the ambiguity that @PaulStanley initially exposed on the first PR, which is you could end up with different ways to represent the same thing.

``` yaml
- page: '1, 3, 5-10'
```
``` yaml
- page: 1
- page: 3
- page: '5-10'
```
``` yaml
- page: 1
- page: 3
- page: '5-10'
```

Also, keep in mind, all of these cases are not that common (a range of 20 pages, multiple pages + lines, etc.).

bwiernik:
The most important thing is that we don't allow endlessly nesting locator objects. The nested structure of `page: [1, 2, {line: 5}, [4, 5]]` is ambiguous about exactly what should be rendered (e.g., should it be "p. 1, p. 2, line 5, pp. 4–5" or "pp. 1, 2, line 5, pp. 4–5") and can potentially lead to endlessly nesting subarrays or ambiguities about whether an array of length 2 is a range or a list of two pages. 

I really don't think the nested structure is workable. I think we need to say that the final components of a `locator-variable` array need to exclusively be a number, an unparsed string, or an array of two of those, _NOT_ a nested child locator object. That is, your current structure in general, but without the nested locators objects.

> And you then have the ambiguity that @PaulStanley initially exposed on the first PR, which is you could end up with different ways to represent the same thing.

Two points here. First, it's not a problem if multiple data structures yield the same output (many to one), so long as each structure has unambiguous output (not one to many). Second, I think your examples would lead to different output. 

I am suggesting a simple rendering rule: each `locator-variable` element is interpreted as rendered separately with its own label (as described in the style). This is unambiguous.

**1.**
```
- page: '1, 3, 5-10'
```
This is a page string. It gets parsed for range-delimiter and pluralization of the label. One label.
**Output:** "pp. 1, 3, 5–10"

**2.**
```
- page: [1, 3, [5, 10]]
```
This is a page array. Its elements are rendered in sequence, the last as a range. One label.
**Output:** "pp. 1, 3, 5–10"


**3.**
```
- page: 1
- page: 3
- page: '5-10'
```
This is three page arrays. They are each rendered separately, with separate labels. The last is a string. It gets parsed for range-delimiter and pluralization of the label.
**Output:** "p. 1, p. 3, pp. 5–10"

**4.**
```
- page: 1
- page: 3
- page: ['5-10']
```
This is three arrays. They are each rendered separately, with separate labels. The last is an array. The string it contains is not parsed and is rendered literally.
**Output:** "p. 1, p. 3, p. 5-10"

**5.**
```
- range:
  - page: 2
    line: 4
  - page: 3
    line: 1
```
This is a range. The two locator objects are rendered and separated by page-range-delimiter. The elements of each locator object are rendered each with its own label.
**Output:** "p. 2, line 4–p. 3, line 1"

The outer array is useful if we allow separate delimiter attributes on `locators` and `locator` (e.g., to allow `[{volume: 2, page: 13}, {volume: 4, page: 36}] to be rendered as "vol. 2:13; vol. 4:36"`)

bdarcus:
> Maybe, but I _think_ it's solvable.
> 
> The current problem is the definitions for `locators` and `locator-variable` are recursive. Just need to figure out how to fix that.

I think I figured it out. The commit I just pushed removes the recursion, and adds some definition to that inner array.

Not fully tested though.

bdarcus:
@PaulStanley - could you weigh in again please?

First, an update on my technical investigation:

I don't believe it's possible with JSON Schema to restrict depth of recursion. 

Further, it's likely no matter what we do, the schema cannot represent it on its own, without accompanying documentation.

What we're wanting to do is reasonably complex, and I just don't think JSON schema handles complexity like RNG.

So best for us to just focus on what we want in the data, and what features we want to support, and which we don't, and leave ti to developers to implement it according to instructions.

Second, response to these proposals ...

First, thanks for these well-structured examples; they demonstrate the issues well.

> I am suggesting a simple rendering rule: each `locator-variable` element is interpreted as rendered separately with its own label (as described in the style). This is unambiguous.
> 
> **1.**
> 
> ```
> - page: '1, 3, 5-10'
> ```
> 
> This is a page string. It gets parsed for range-delimiter and pluralization of the label. One label.
> **Output:** "pp. 1, 3, 5–10"

I just want to first point out the spec says zero about the input model for this. It's super vague about locators in general.

I suggest we not enshrine the string, but keep it for compatibility, because if we do, we need to detail how to parse (see earlier discussion), and developers must support it.

I'd rather just leave it up to them, than require it. The purpose of the structured representation is so that it is closer to self-documenting.

Either way, though, we should be able to get from one to the other.

> **2.**
> 
> ```
> - page: [1, 3, [5, 10]]
> ```
> 
> This is a page array. Its elements are rendered in sequence, the last as a range. One label.
> **Output:** "pp. 1, 3, 5–10"

So far so good.
 
> **3.**
> 
> ```
> - page: 1
> - page: 3
> - page: '5-10'
> ```
> 
> This is three page arrays. They are each rendered separately, with separate labels. The last is a string. It gets parsed for range-delimiter and pluralization of the label.
> **Output:** "p. 1, p. 3, pp. 5–10"

Why in this example is the range a string?

More fundamentally, why should we support this in the _input model_, or in CSL at all? 

That output could in theory be configured in the style, even if I don't think it makes any sense.

If someone wants that level of control _at the individual citation level_ (as opposed to globally, and so defined in the style), maybe they should just use strings?

``` js
['p. 1', 'p. 3', 'pp. 5–10']
```
OR

``` js
['p. 1', 'p. 3', [5, 10]]
```

I agree, though, this would be justification for the outer array. And I don't really have any big objection to it myself. For me, really depends on the developers.
 
> **4.**
> 
> ```
> - page: 1
> - page: 3
> - page: ['5-10']
> ```
> 
> This is three arrays. They are each rendered separately, with separate labels. The last is an array. The string it contains is not parsed and is rendered literally.
> **Output:** "p. 1, p. 3, p. 5-10"

Same question: I don't see why we should support it, or why the last page item is a string within the array, rather than an array.
 
> **5.**
> 
> ```
> - range:
>   - page: 2
>     line: 4
>   - page: 3
>     line: 1
> ```
> 
> This is a range. The two locator objects are rendered and separated by page-range-delimiter. The elements of each locator object are rendered each with its own label.
> **Output:** "p. 2, line 4–p. 3, line 1"

This one is a more reasonable feature to support, I think.

My druthers are not to support, and this PR doesn't currently, but of course would welcome further feedback.

Again, I think whether we should support it should depend on some balance of developer feedback, and actual real-world use cases.

Below, while I think about it, is brainstorming for string parsing rules:

1. simple locators are delimited by commas (1, 3, 5)
2. simple ranges are delimited by hyphens (23-35)
3. compound locators (p2, line 4) are delimited by semi-colons (p2, line 12; p4, line 2)
4. compound ranges are delimited with a forward-slash (p2, line 12-14/p4, line 2)
5. labels precede the value for compound locators where the sub-locator differs from the locator object (so, for a page locator, '12" means page 12)

bwiernik:
> I just want to first point out the spec says zero about the input model for this. It's super vague about locators in general.

Yes, multiple locators are currently completely off-spec behavior. We have never formally permitted multiple locators before.

bwiernik:
@bdarcus I am not sure that we are interpreting the YAML syntax the same way (I am probably misunderstanding it). Can we switch to just writing in actual JSON?

bwiernik:
Taking a big step back to describe my proposal at a high level, let's define some things:

1. `locators`: an array of multiple `locator` objects.
1. `locator`: an object with two possible types of properties: `locator-variable` and `locator-range`.
1. `locator-variable` can be one of `string` or `array`. This is the definition for `page`, `
  1. If `string`, it is parsed for pluralization of the label and normalization of range-delimiter
  1. If `array`, allowed elements are a (1) number, (2) a string, or (3) an array of length 2 containing numbers or strings.
      - When `locator-variable` is an array, string are not parsed but are rendered literally.
1. `locator-range` is an array of length 2 containing two `locator` objects.
     - We could define another `locator-range-point` object class that is the same as `locator` but only allowing `locator-variable` properties (not the `locator-range` range property). That might not be necessary. As far as I can see, the recursion problem happens when we allow `locator-variable` to have elements other than number, string, or array of those two.

Does this make my examples clearer?

When there are multiple `page` properties, these are each their own `locator-variable` objects, so they each go through the string vs. array parsing choice separately.

bwiernik:
> Why in this example is the range a string?
>
> More fundamentally, why should we support this in the input model, or in CSL at all?

This is just the "locator variables may be either a string or an array" compromise we agreed to.

Each individual `page` property is a separate `locator-variable` that can be a string or an array.

bdarcus:
Sure.

JSON arrays are actually more clear anyway.

bdarcus:
Ok, but the only reason to use a string is if it's not a number or a range.

bwiernik:
So in JSON syntax:

**1.**
```
locators: [{"page": "1, 3, 5-10"}]
```
`locators` is an array, containing one `locator` object. `page` is a `locator-variable` containing a string. Because it's a string, it gets parsed for range-delimiter and pluralization of the label. One label.
**Output:** "pp. 1, 3, 5–10"

**2.**
```
locators: [{"page": [1, 3, [5, 10]]]
```
`locators` is an array, containing one `locator` object. `page` is a `locator-variable` containing an array. Because it's an array, its elements are rendered literally. The last element is an array, rendered as a range. One label.
**Output:** "pp. 1, 3, 5–10"


**3.**
```
locators: [{"page": "1"}, {"page": "3"},  {"page": "5-10"}]
```
`locators` is an array, containing three `locator` objects. In the first `locator` object, `page` is a `locator-variable` containing a string. It's parsed for plural labels and range delimiters (both false here). Same for the second `locator` object. Same for the third `locator` object; here, when the string `locator-variable` is parsed, the label is plural and the range delimiter is normalized. Each `locator-variable` gets its own label because they were specified separately.
**Output:** "p. 1, p. 3, pp. 5–10"

This example might be appear more useful if it looked like this:
**3b.**
```
locators: [{"page": "1", "line":"2"}, {"page": "3", "line":"4"},  {"page": "5-10"}]
```
Again, `locators` is an array, containing three `locator` objects. Within each of the `locator` objects, each property is evaluated in turn. Each `locator-variable` gets its own label, and `locator-variable`s expressed as a string (versus as an array) are parsed. 
**Output:** "p. 1, line 2; p. 3, line 4; pp. 5–10"

(Here I used separate delimiters for `locators` versus `locator`.)

**4.**
```
locators: [{"page": [1]}, {"page": [3]},  {"page": ["5-10"]}]
```
This is similar to **3.** but the elements of the the `page` `locator-variables` are arrays instead of strings. Because they are arrays, they are not parsed and are rendered literally and without pluralization. `5-10` is taken to mean "the single with the hyphenated number "5-10" (e.g., perhaps page 10 of part 5 of a report).

Each `locator-variable` is rendered with its own label.
**Output:** "p. 1, p. 3, p. 5-10"

**5.**
```
locators: [{"range":[{"page": "1", "line":"2"}, {"page": "3", "line":"4"}]}]
```
`locators` is an array, containing one `locator` object. `range` is a `locator-range` containing an array. The two elements of the array are each `locator` objects, each containing two `locator-variable` properties. The two `locator` objects are rendered and separated by page-range-delimiter. The elements of each `locator` object are rendered each with its own label.
**Output:** "p. 2, line 4–p. 3, line 1"


The point of this proposal is that it has a single simple set of rendering rules defined for `locator-variable` that are applied consistently no matter what the locator variable is. The processor doesn't have to be concerned about any hierarchical structure because the rendering choices always happen at the lowest level on `locator-variable`.

bwiernik:
> In my proposal, processors won't parse strings.

I don't think we are understanding things the same at all. Please look at my latest description.

bdarcus:
> 4\. **compound ranges** are, as with EDTF, delimited with a forward-slash (p2, line 12-14/p4, line 2)

A modification of current PR to support this feature could just be this (now in JSON):

``` js
"page": [
  [2, { "line": [12, 14]}],
  [4, { "line": [2]}]
]
```

... though I wonder if the nested arrays get brittle; maybe be better as a range object?

``` js
{
	"page": {
		"begin": [2, {
			"line": [{
				"begin": 12,
				"end": 14
			}]
		}],
		"end": [4, {
			"line": [2]
		}]
	}
```

bwiernik:
I _really_ _really_ don't think that we should allow "line" to be nested under "page". _That_ is the major thing I'm trying to communicate is causing ambiguities and problems. Each "page", "line", "chapter", etc. property needs to be handled as a discrete processing step.

bwiernik:
Each "page", "line" "chapter" is a `locator-variable` object that is rendered separately from any other `locator-variable` objects. They can contain only strings or arrays of strings. That is what I'm trying to communicate. We shouldn't allow a deeper nesting than that.

bwiernik:
@bdarcus Perhaps it would be easier if I opened a new PR with what I'm trying to communicate?

bdarcus:
Sure.

>

bdarcus:
> I don't think we are understanding things the same at all. Please look at my latest description.

There are two discrete issues, just so you can keep them clear in the separate PR.

1. is parsing required by a CSL 1.1 processor. In my proposal, no; that's the whole point! Where there are strings, they are formatted as is (notwithstanding the compatibility string representation).
2. how the thing is modeled (objects vs arrays, in particular)

bdarcus:
Is something like [this](https://gist.github.com/bdarcus/c5a31b2e3ef40ca4b16dd27f06467510) what you mean @bwiernik?

The advantage of that is the validation is simpler, and there's no recursion problem.

The change is just to go back to the initial PR, but to change the "string" value to an array, and allow individual items to be only number or string.

bdarcus:
And, of course, there's always @cormacrelf's [solution](https://github.com/citation-style-language/schema/pull/325#issuecomment-662353431) (see also this [comment](https://github.com/citation-style-language/schema/pull/325#issuecomment-661957498)):

> { "locators": [[["part", "2"], ["division", "7, 9"]], ["schedule", "4"]] }

So the value is just a straight up two-level array.

Not sure, however, how he was planning to handle ranges in that, or the significance of the order within the sub-array (does label/type always come first, for example?).

bdarcus:
I believe the initial post now reflects the full range of requirements,  a couple of alternatives, including doing nothing, and explanation of the PR.

It's pretty similar to what @bwiernik was proposing most recently, with some minor differences.

I'm not sure the schema perfectly reflects the text yet, and as I suggested earlier, it's probably not possible anyway (It will be looser than we'd want). But I think as long as the other components (schema annotations, etc.) are clear, we'll be OK.

bwiernik:
I think I've worked out what the style syntax needs to be to handle multiple locators. Will open a PR tomorrow.

bwiernik:
> I believe the initial post now reflects the full range of requirements, a couple of alternatives, including doing nothing, and explanation of the PR.

There are three things this doesn't handle. 
1. Two-level arrays/a higher-level `locators` array containing subordinate `locator` arrays. That is needed if there are different delimiters between vs with a `locator`. For example, Chicago style calls for citing volume-page combinations like this:
> Complete Tales of Henry James, 5:34.

But otherwise delimit multiple locators with commas:
> Complete Poems of Michelangelo, p. 89, lines 135-36

I think this is a minor enough thing that we could ignore it.

2. Multi-locator ranges (e.g., "p. 2, line 4–p. 3, line 10"). The range endpoints need to be a an array.

3. A way to distinguish a parsed versus unparsed string. A user should be able to enter `page: 1, 2` and get "pp. 1, 2" or enter `page: 5-6` and get "pp. 5–6". As you've said may times, this is the most common case--it needs to be doable. This structure is way too common, both in `locator` and more importantly in other fields, especially `page`, to not do in the processor. This is the point I've been returning to several times. 

Moreover, there needs to be a way to get one label if there are multiple of the same locator:
"pp. 5, 6"

But a new label for each locator if they are different:
"p. 6, line 10"

If each locator variable is restricted to be one number or one unparsed string, then there is no way to do that without either adding more style parameters or requiring processors to incorporate some sort of label-collapsing logic.

bdarcus:
> Two-level arrays/a higher-level locators array containing subordinate locator arrays.... I think this is a minor enough thing that we could ignore it.

I agree.

Could be handled with a string.

> * Multi-locator ranges (e.g., "p. 2, line 4–p. 3, line 10"). The range endpoints need to be a an array.

Let's use consistent language for the features/requirements. 

This is what I called "Compound Ranges" in the requirements list. 

If people find your language more clear, we can revise it to "Multi-Locator Ranges." 

That aside, definitely this is the intention, and remaining decision actually.

I think my preference would be to have the values always arrays, but my thinking so far has been that it could be a single number/string, or an array.

This is really where we need input from developers.

Aside: I've asked @andras-simonyi to take a look at this and weigh in, and he said he would. He's done a lot of work with locators, so let's see what he says.

> * A way to distinguish a parsed versus unparsed string. 

I know you keep saying this, but I don't understand why. 

We had a long discussion, with input from multiple developers, on structured data vs strings, and we agreed we'd focus on the former for the input model.

Why is this different?

> A user should be able to enter ...

While I will come back to this at the end, we're not designing user facing apps. We're describing an input model for developers.

> `page: 1, 2` and get "pp. 1, 2" or enter `page: 5-6` and get "pp. 5–6". 

I don't see what the specific problem here is. The PR addresses this case without issue.

``` js
"locators": [ {"page": [1 , 2} ]
```
In the range-array option (based on your suggestion, which as I say above, I prefer):

``` js
"locators": [ {"page": { "range": [5 , 6] } ]
```

And in the range-object option:

``` js
"locators": [ {"page": {"begin": 5 ,"end": 6} ]
```

To go to my point about parsing, and the section at the end of the main post, that's easily generated from string data.

So while it's not our job to tell app developer how to do it, I could imagine a user entering this in some page field, and generating that.

```
1, 2, 5-6
```

... and if they needed the example @PaulStanley mentioned, they'd do:

```
1, 2, "A-10"
```

Right? What am I missing?

denismaier:
>     1. Two-level arrays/a higher-level `locators` array containing subordinate `locator` arrays. That is needed if there are different delimiters between vs with a `locator`. For example, Chicago style calls for citing volume-page combinations like this:
> 
> 
> > Complete Tales of Henry James, 5:34.
> 
> But otherwise delimit multiple locators with commas:
> 
> > Complete Poems of Michelangelo, p. 89, lines 135-36

I'm not sure this is a question for the input model.
Couldn't that be handled on a style basis, similar to the existing editortranslator collapsing? Like with a style level attribute `collapse-volume-page-in-locator`.

E.g. if we have this:

```
"locators": [{"volume": 2}, {"page": 12}]
```

`collapse-volume-page-in-locator="true"` => "2:12"
`collapse-volume-page-in-locator="false"` => "vol. 2, p. 12"

bwiernik:
> To go to my point about parsing, and the section at the end of the main post, that's easily generated from string data.

I agree--but @PaulStanley 's objection was that there should be an option to specify a string that should not be parsed. That was the origin of the "array or string" proposal. 

As a simpler alternative, we could adopt some BibTeX-style syntax to indicate "do not parse this string"--that would be useful for many places to control number parsing.

bdarcus:
> I agree--but @PaulStanley 's objection was that there should be an option to specify a string that should not be parsed. That was the origin of the "array or string" proposal.

Now would be a good time for him to weigh in. 

But my proposal on this piece, which none of the developers have yet commented probably because it's a subtle thing, is the string itself is just what I understood him to be asking for. The processor would just render that as is.

bwiernik:
For the volume-page issue, we could just expect the user to enter "2:124" as a `page`. 

I suggest we just skip this point for now. If we decide to revisit in a future revision, there are ways we could build on this framework to handle that.

denismaier:
> I suggest we just skip this point for now. If we decide to revisit in a future revision, there are ways we could build on this framework to handle that.

Fair enough. As I said above, I don't think this has much to do with the input data model, We can revisit later. (I'll open a separate issue.)

schemas/input/csl-citation.json
```diff
@@ -35,122 +35,262 @@
...
+          "type": ["number", "string"]
         },
-        "volume": {
-          "type": "string"
+        {
+          "title": "Locator Range",
+          "type": "object",
+          "properties": {
+            "begin": {
+              "type": ["number", "string"]
```
bwiernik, line 284:
```suggestion
              "type": "#/definitions/locator-variable"
```

This needs to be a `locator-variable` object. Right now, a range object doesn't specify the type of locator it is. Changing the definition also supports compound ranges.

schemas/input/csl-citation.json
```diff
@@ -35,122 +35,262 @@
...
+            },
+            "end": {
+              "type": ["number", "string"]
```
bwiernik, line 287:
```suggestion
              "type": "#/definitions/locator-variable"
```

This needs to be a `locator-variable` object. Right now, a range object doesn't specify the type of locator it is. Changing the definition also supports compound ranges.

bwiernik:
Added some suggestions to address the range issue (which affects both both simple ranges and compound ranges).

andras-simonyi:
Sorry for reacting that late (again).  Predictably, I'm absolutely in favour of the proposal to bring as much structure into the representation as possible, even at the price of added complexity, so I'm strongly against the "Do Nothing Alternative". 

As for the structured representation, this is probably a totally stupid question, but would it be too radical to say that 

(i) the basic building blocks are 'atomic locators'  represented as

    {"page": 101}

    {"chapter": 2, "page": 122,"line": 14} 

etc., in general, as objects in which keys are locator terms and values can only be numbers and strings. Literal strings could also be added to this category. 

(ii) More complex 'locators' are built from simple(r) ones  by forming lists (represented as arrays) or ranges (represented by {"begin": <begin>,  "end": <end>} objects)? I'm not sure how far recursion should be taken here, lists can obviously have ranges as elements, but ranges can probably have only "atomic locators" as their "begin" or "end".  Do we want to support hierarchical lists like  "pp. 1, 4, 6; 45, 47, 50-53"?

Would something like this make sense at all? I see that in many cases it'd be more verbose than the present proposal, e.g.  instead of `{page: [1, 3]}` for "pp. 1, 3" we would have `[{page: 1}, {page: 3}]` etc.

bdarcus:
> As for the structured representation, this is probably a totally stupid question, but would it be too radical to say that
> 
> (i) the basic building blocks are 'atomic locators' represented as

So in other words, don't allow value array for the locator objects?

What would be the advantage from your standpoint?

> (ii) More complex 'locators' are built from simple(r) ones by forming lists (represented as arrays) or ranges (represented by {"begin": , "end": } objects)? I'm not sure how far recursion should be taken here, lists can obviously have ranges as elements, but ranges can probably have only "atomic locators" as their "begin" or "end". Do we want to support hierarchical lists like "pp. 1, 4, 6; 45, 47, 50-53"?

Your example would be represented by this in this proposal currently:

``` js
[
  "page": [1, 4, 6, 45, 47, {"begin": 50, "end": 53}]
]
```

... or, in the range-array alternative:

``` js
[
  "page": [1, 4, 6, 45, 47, {"range": [50, 53] } ]
]
```

Are you suggesting instead this?

``` js
[
  { "page": "1"}, { "page": "4"}, { "page": "6"}, { "page": "45"}, { "page": "47"}, 
  { "page": {"begin": 50, "end": 53}}
]
```

denismaier:
> In certain respects it seems simpler and more uniform to me than the current proposal, but I see that in many cases it'd be more verbose

As we are talking about an input format for processors, I don't think verbosity is a strong argument against whichever approach.

andras-simonyi:
> So in other words, don't allow value arrays for the locator objects?

Yes, but at the same time move all "atomic information" about location (chapter, line etc.) into these objects.

> What would be the advantage from your standpoint?

I guess one advantage would be a uniformity between locators containing "simple" and "compound" elements in the sense that structurally parallel locators that differ only by adding a detail like "pp. 23, 25"  vs "p. 23, p. 25 line 6" would have similarly parallel representations. 

    [{"page": 23}, {"page": 25}] vs [{"page": 23}, {"page": 25, "line": 5}]
 
Depending on how this is handled  on the style side this might  make processing easier. OTOH this example also shows that the "surface forms" can be rather different, so this is far from being a strong argument.  BTW, I'm not totally sure what would be the representation of these lists containing "compounds" according to the current proposal, would they be unambiguous?

> > Do we want to support hierarchical lists like "pp. 1, 4, 6; 45, 47, 50-53"?
> 
> Your example would be represented by this in this proposal currently:
> 
> ```js
> [
>   "page": [1, 4, 6, 45, 47, {"begin": 50, "end": 53}]
> ]
> ```

This  disregards the semicolon in the example which was intended to indicate a grouping/hierarchy, but it's totally legitimate to say that this kind of grouping isn't  supported. (I vaguely remember seeing it in the wild but am unsure how frequent it is, might be the effect of too much Lisp programming :-) ).

> Are you suggesting instead this?
> 
> ```js
> [
>   { "page": "1"}, { "page": "4"}, { "page": "6"}, { "page": "45"}, { "page": "47"}, 
>   { "page": {"begin": 50, "end": 53}}
> ]
> ```
Without the grouping it would be almost this but even more verbose:
```js
 [
   { "page": "1"}, { "page": "4"}, { "page": "6"}, { "page": "45"}, { "page": "47"}, 
   {"begin": {"page": 50}, "end": {"page": 53}}
 ]
 ```
With grouping (if such a thing is supported), on the other hand, it'd be
```js
 [
   [{ "page": "1"}, { "page": "4"}, { "page": "6"}], 
   [{ "page": "45"}, { "page": "47"},  {"begin": {"page": 50}, "end": {"page": 53}}]
 ]
 ```

**Comment by bdarcus** [comment]
> Yes, but at the same time move all "atomic information" about location (chapter, line etc.) into these objects.

OK, I think I understand now.

Let's take my compound locator example from the parsing section, which I think may clarify:

> p2, line 12; p4, line 2

My proposal would result in this (giving the full locator property representation):

``` js
"locators": [
  {"page": 2, "line": 12},
  {"page": 4, "line": 2}
]
```

... or, if we required an array for the values (let's assume this below):

``` js
"locators": [
  {"page": [2], "line": [12]},
  {"page": [4], "line": [2]}
]
```
So to your point about atomicity, I think we have that.

.... but I list of pages would be:

``` js
"locators": [
  {"page": [21, 23]},
]
```
.. or:

``` js
"locators": [
  {"page": [21]},
  {"page": [23]},
]
```

So you're pointing out we have two different ways to represent a list of locators, and that could lead to confusion, particularly when we through in compound locators?

Your proposal would only have one way to represent the list of locators, at the expense of verbosity, which as @denismaier said is not an issue in a machine-only input format, but could be in one also designed for humans.

Aside: I had hoped we could do both, but if we have to choose, we have to err on the machine side for these purposes I think.

Also worth keeping in mind by far the most common locator will be a single page.

> This disregards the semicolon in the example which was intended to indicate a grouping/hierarchy,

So you mean like a primary locator, and secondary one that is often, but maybe not always, a "see also" locator?

We maybe should add that to the requirements list somewhere.

bdarcus:
I just pushed a commit that (I think) implements @andras-simonyi's suggestion.

Here's the embedded examples:

``` js
            "examples": [
              { "locators": [{ "page": 23 }] },
              {
                "locators": [{ "begin": { "page": 23 }, "end": { "page": 25 } }]
              },
              { "locators": [{ "page": 23 }, { "line": 2 }] },
              {
                "locators": [
                  { "page": 23 },
                  { "begin": { "page": 23 }, "end": { "page": 25 } }
                ]
              }
            ]
```

Note: this does not include the "hierarchical" feature currently, mostly because I don't yet understand it.

And a reminder: JSON schema [kind of sucks](https://www.tbray.org/ongoing/When/201x/2016/04/30/JSON-Schema-funnies). The schema is intended to describe the model, but we shouldn't have any expectation any tool will tightly validate it.

andras-simonyi:
@bdarcus:
> .... but a list of pages would be:
> 
> ```js
> "locators": [
>   {"page": [21, 23]},
> ]
> ```
> 
> .. or:
> 
> ```js
> "locators": [
>   {"page": [21]},
>   {"page": [23]},
> ]

> So you're pointing out we have two different ways to represent a list of locators, and that could lead to confusion, particularly when we throw in compound locators?
> 

Yes, this was one of the points, and it is indeed a question how should a human-digestible YAML representation look like. What I'm unsure about is that there are obviously a lot of "collapsing" rules at play for typical styles (e.g, "p. 23, p. 27" => "pp. 23, 27", "p. 23 line 17, p. 23 line 28" => "p. 23 lines 17, 27" etc.) and these are either universal, in which case your original proposal which in a sense represents these collapses directly might have the advantage, or not, but then there should be means of expressing the differences on the style side. Unfortunately I'm not well versed enough in the various formatting styles to decide which is the case. 

> So you mean like a primary locator, and secondary one that is often, but maybe not always, a "see also" locator?

Yes, actually I didn't think about "see also" but this is a very good example, although leading to the question of how should an introductory string like "see also" be handled here.

> We maybe should add that to the requirements list somewhere.
> And if we support it, you're suggesting a sub-array for `locators` to denote the hierarchy?

Yes, I think this would be the proposal.

bdarcus:
> What I'm unsure about is that there are obviously a lot of "collapsing" rules at play for typical styles (e.g, "p. 23, p. 27" => "pp. 23, 27", "p. 23 line 17, p. 23 line 28" => "p. 23 lines 17, 27" etc.) and these are either universal, in which case your original proposal which in a sense represents these collapses directly might have the advantage, or not, but then there should be means of expressing the differences on the style side. Unfortunately I'm not well versed enough in the various formatting styles to decide which is the case.

My own impression was indeed that the collapsing is the norm, but that in any case, if we wanted to express the difference it would happen on the styling end, or not at all; that by default we'd only do collapsing.

Part of why I came up with that approach.

But it seems if we did support the distinction, it should be easy to do; perhaps with an attribute like `@collapse-locators` with a default value of `true`?
 
> > So you mean like a primary locator, and secondary one that is often, but maybe not always, a "see also" locator?
> 
> Yes, actually I didn't think about "see also" but this is a very good example, although leading to the question of how should an introductory string like "see also" be handled here.

Indeed; this is where this all gets even more complicated.

"see also", now that I think of it, is a prefix for a group of cites within a larger citation; example: (Doe, 2018; see also Smith, 2019).

If we're talking locators, it might be something like this: (Doe, 2018, p45; see also p54).

So cites can be grouped (and often are in certain fields), and locators within cites can also be grouped (probably less commonly).

It seems _if we wanted to express locator groups like this_, perhaps the group itself would need to allow affixes?

``` js
"locators": [
  [
     {"text": "see also"},
     {"page": 54}
  ]
]
```
... which maybe bring us back to @denismaier's suggestion on allowing text, which we said we'd defer; but at least, I think, this clarifies how it could work.

Does that make sense?

> > We maybe should add that to the requirements list somewhere.
> > And if we support it, you're suggesting a sub-array for `locators` to denote the hierarchy?
> 
> Yes, I think this would be the proposal.

So the question, then: do we want to support sub-locator groups? @bwiernik @denismaier?

bdarcus:
> Also, if we do include the sub-array, how is a non-grouped (or single group, really) locators represented?

What if we just used the array for additional groups, beyond the main/default, @andras-simonyi?

denismaier:
> "see also", now that I think of it, is a prefix for a group of cites within a larger citation

That remembers me a bit of our earlier discussion on the citation syntax on the orgmode list where global affixes have been an issue. Adding citation groups would give us about this:

```yaml
cite:
  prefix: text
  suffix: text
  citatation-group:
    prefix:
    suffix: 
    citations:
      - prefix: text
         suffix: text
         locators: 
         citation-key: 
      - prefix: text
         suffix: text
         locators: 
         citation-key: 
  citatation-group:
    prefix:
    suffix: 
    citations:
      - prefix: text
         suffix: text
         locators: 
         citation-key: 
      - prefix: text
         suffix: text
         locators: 
         citation-key: 
```

bdarcus:
> That remembers me a bit of our earlier discussion on the citation syntax on the orgmode list where global affixes have been an issue.

Yes, I was thinking this work could inform the org citation work.

bdarcus:
> What if we just used the array for additional groups, beyond the main/default, @andras-simonyi?

To cover requirement now number 5 (grouping), I have just pushed a new "sub-group" definition. This also , I think, addresses @denismaier's earlier request.

``` js
"locators": [
  { "page": 259, "line": 12 },
  ["see also", { "page": 273, "line": 3 }]
]
```

Last call for comments; I plan to merge this tomorrow (Friday, Aug 6).

PR #329, opened 2020-07-28
Add repeatable attribute to rendering elements

### LINKED ISSUE #287: Extend cs:substitute to work with other/all variables?
Currently, `cs:substitute` can only be used as a child element of `cs:names`. Would it make sense to extend this to other/ types of variables? Think of:

```xml
<text variable="title>
  <substitute>
    <text variable="genre"/>
  </substitute>
</text>

### PR DESCRIPTION
**Pull request by denismaier**
## Description

In this [discussion](https://discourse.citationstyles.org/t/dropping-cs-substitute/697/52) on discourse, we somehow agreed on introducing general one time rendering, and a forced re-rendering with an attribute `repeatable=true`.

If we indeed introduce this, this will need an addition to the specs and a new attribute for rendering elements.

This PR already adds the required attribute for re-rendering otherwise suppressed variables.

Fixes #287 

## Type of change

## Checklist

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -1012,6 +1012,13 @@ div {
 }
 # ==============================================================================
 
+## Allow repeated rendering of a rendering-element
+div {
+  ## By default, variables are rendered only once.
+  ## The "repeatable" attribute allows re-rendering of otherwise suppressed variables.
+  repeatable = attribute repeatable { xsd:boolean }?
+}
+
 ## Formatting attributes.
 div {
   affixes =
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -1019,7 +1019,8 @@ div {
 div {
   ## By default, variables are rendered only once.
   ## The "repeatable" attribute allows re-rendering of otherwise suppressed variables.
-  repeatable = attribute repeatable { xsd:boolean }?
+  ## The "always-render" attribute allows re-rendering of otherwise suppressed variables.
+  always-render = attribute always-render { xsd:boolean }?
 }
 
 ## Formatting attributes.
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -397,7 +397,7 @@ div {
     delimiter,
     display,
     font-formatting,
-    repeatable
+    always-render
   names.name =
     element cs:name {
       name.attributes,
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -396,7 +396,8 @@ div {
     ## the same cs:names element.
     delimiter,
     display,
-    font-formatting
+    font-formatting,
+    repeatable
   names.name =
     element cs:name {
       name.attributes,
@@ -578,6 +579,7 @@ div {
       affixes,
       display,
       font-formatting,
+      repeatable,
       text-case
     }
   rendering-element.date.date-part.localized =
@@ -642,6 +644,7 @@ div {
       display,
       font-formatting,
       quotes,
+      repeatable,
       strip-periods,
       text-case
     }
@@ -673,7 +676,7 @@ div {
     
     ## Use to render a number variable.
     element cs:number {
-      number.attributes, affixes, display, font-formatting, text-case
+      number.attributes, affixes, display, font-formatting, repeatable, text-case
     }
   number.attributes =
     attribute variable { variables.numbers },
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by bdarcus** [comment]
I think we've already settled we should include this, but just to confirm, copying @fbennett @PaulStanley @cormacrelf

RTM?

**Comment by fbennett** [comment]
Wouldn't it be better for clarity to set this on ``cs:style`` with the repeatable variables as argument? If it's set on renderable elements, there will be edge cases where the effect is not immediately obvious.

**Comment by bdarcus** [comment]
> Wouldn't it be better for clarity to set this on `cs:style` with the repeatable variables as argument? If it's set on renderable elements, there will be edge cases where the effect is not immediately obvious.

I guess the only reason not to do that is if in some cases, on the same style, a variable should be repeatable, and in other cases not.

If we don't need to support that, then it seems we shouldn't, because it's by definition more complicated.

Maybe we should settle this question first before settling how to implement it: is this setting global for a style, or not?

What say you @denismaier @bwiernik?

**Comment by denismaier** [comment]
I think the idea of this was to allow a specific variable to appear for a second time **in a specific place**, not to make a variable to be repeatable as such, like `genre` might appear in 4 different locations in a style, depending on whatever conditions. We will always want in the the last location, but 2 and 3 should be suppressed if the variable already appears in location 1. If it appears in location 3, suppress the variable in location 3. But always print it at location 4.
(I know the example is totally artificial and made up. I really can't think of many examples where you want a variable to appear twice.)

**Comment by denismaier** [comment]
Other question: What should we do with date variables? Exempt those from the one-time rendering rule, as suggest by @bwiernik [here](https://discourse.citationstyles.org/t/dropping-cs-substitute/697/23?u=denis_maier)?

**Comment by bdarcus** [comment]
Is the date issue maybe really a "form" issue?

>

**Comment by fbennett** [comment]
> I think the idea of this was to allow a specific variable to appear for a second time **in a specific place**, not to make a variable to be repeatable as such...

Okay, that's clear. So the meaning is something like "force-render."

**Comment by denismaier** [comment]
Exactly, `force="true"` was actually one of the suggested names for the attribute.

**Comment by bdarcus** [comment]
> Exactly, `force="true"` was actually one of the suggested names for the attribute.

I think `force-render` is more clear.

In that case, maybe change it?

**Comment by bwiernik** [comment]
I like force-render or always-render as attribute names. 

For dates, I think we can think of that as a "form" situation like titles and names. For dates, either a different form or different name-parts would not be suppressed.

File: schemas/styles/csl.rnc [hunk 1]
```diff
@@ -578,6 +579,7 @@ div {
       affixes,
       display,
       font-formatting,
+      repeatable,
```
bdarcus, line 582:
```suggestion
      always-render,
```

schemas/styles/csl.rnc
```diff
@@ -642,6 +644,7 @@ div {
       display,
       font-formatting,
       quotes,
+      repeatable,
```
bdarcus, line 647:
```suggestion
      always-render,
```

schemas/styles/csl.rnc
```diff
@@ -673,7 +676,7 @@ div {
     
     ## Use to render a number variable.
     element cs:number {
-      number.attributes, affixes, display, font-formatting, text-case
+      number.attributes, affixes, display, font-formatting, repeatable, text-case
```
bdarcus, line 679:
```suggestion
      number.attributes, affixes, display, font-formatting, always-render, text-case
```

schemas/styles/csl.rnc
```diff
@@ -1012,6 +1015,13 @@ div {
 }
 # ==============================================================================
 
+## Allow repeated rendering of a rendering-element
+div {
+  ## By default, variables are rendered only once.
+  ## The "repeatable" attribute allows re-rendering of otherwise suppressed variables.
+  repeatable = attribute repeatable { xsd:boolean }?
```
bdarcus, line 1022:
```suggestion
  always-render = attribute always-render { xsd:boolean }?
```

bdarcus: Actually, this docstring should be changed accordingly too.

schemas/styles/csl.rnc
```diff
@@ -396,7 +396,8 @@ div {
     ## the same cs:names element.
     delimiter,
     display,
-    font-formatting
+    font-formatting,
+    repeatable
```
bdarcus, line 400:
```suggestion
    always-render
```

Issue #293, opened 2020-06-30
add new creator types


bwiernik:
Replaces and updates https://github.com/citation-style-language/csl-evolution/issues/33

Some additional creator types we might consider adding. 

This one should be added for book items and the like:
  - `narrator`
    - as for an audiobook
    - similar level of contribution as `illustrator`

These should be added for media items. They finish out the media roles that currently exist in `director`, `producer`, `executive-producer`, and `performer`:
  - `writer`
    - as in a screenwriter for media items
    - distinguished from `author` in that it has parity with `director` and `producer`, whereas `author` would be more analogous to a `creator` with responsibility for the whole work
  - `creator`
    - often cited for media items
    - conceptually really similar to `author` or `container-author` I'd argue, but with a different label
    - I thought about whether we could just use `author` and set the localized labels to be "creator" and "created by", but that would cause a lot of problems
  - `host` 
    - for a regular broadcast
    - more appropriate than `director`; perhaps is analogous to `author`, but `author` might better be conceived
  - `guest`
    - for a broadcast

Basically, what I am realizing is that media creators are a rather different beast from other types of citations. We might create a new variable class `variables.media-creator` with processing rules similar to `editortranslator`. The same could be done with `author` and `illustrator`. That would address most, if not all, of the issues addressed by https://github.com/citation-style-language/csl-evolution/issues/7

There are a few other name labels that have been requested. I'm less sure on these:
  - `principal-investigator`
    - for a grant
    - these are cited very often in grant applications, so some way to formally include them would be good
    - a grant would be cited as a `report`
    - might be best covered by adding a `label` element to names data
  - `co-investigator`
    - same as `principal-investigator` above
  - `cartographer`
    - unclear how necessary this is; geography and related fields don't use this label (they just use `author`)
    - we've only seen it called for in the Vancouver style manual, but outside of public health, who in biomedical research is going to cite a map?

We might address these by adding a `label` element to names data (see ) and allowing for localized terms to appear in input data (see https://github.com/citation-style-language/schema/issues/240)

bwiernik:
@adam3smith What are your thoughts on `principal-investigator`? It is something I run into all the time for grant applications. Should we add it as a variable or look for some other solution?

adam3smith:
Do you have an example for a citation with PI? I've not seen this

bwiernik:
So, for example, from the APA manual:
[...]

PR #330, opened 2020-07-29
Add type and creator variables for grants/projects

bwiernik:
Closes https://github.com/citation-style-language/schema/issues/293

## Description

Add a `project` type for grants and similar. For example, such items are often cited on CVs and grant applications.

@bdarcus You said in the linked thread "supporting grants would require more than just a type and creator variable". What else were you thinking? I think the other existing variables (publisher, genre, number, status) covers things.

## Type of change

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -289,6 +289,7 @@
             "personal_communication",
             "post",
             "post-weblog",
+            "project",
             "regulation",
             "report",
             "review",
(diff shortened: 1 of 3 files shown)


bdarcus:
> You said in the linked thread "supporting grants would require more than just a type and creator variable". What else were you thinking? I think the other existing variables (publisher, genre, number, status) covers things.

For CVs, dollar amount. Not sure if, or how best, to support that initially.

bwiernik:
> dollar value 

Good point. There could be a `budget` field. `dimensions` is a reasonable conceptual fit (the value of a grant is similar to the running time of a film or the size of an artwork).

denismaier:
Looks good to me.
Don't have a strong opinion on `budget` vs. `dimensions`.

bdarcus:
> There could be a `budget` field. `dimensions` is a reasonable conceptual fit (the value of a grant is similar to the running time of a film or the size of an artwork).

Maybe let's just use `dimensions` for now. People can always request something more specific if needed.

Does that mean we can just merge?

PR #344, opened 2020-08-06
input: Add new citation command property

### LINKED ISSUE #214: update csl-citation.json to include intext command
As followup to the new intext element coming in v1.1 (see #193), which adds explicit support for author-in-text citations, it seems we should update this schema along with that, and the spec.

@dstillman

bdarcus:
Removes the two boolean citation command flags, and replaces with a 
new "command" property, with possible values of "suppress-author" and 
"author-only".

Also adds a new "narrative" option to that list, to support textual or 
author-in-text citations like "Doe (2019)" that are now supported with 
the new intext style element.

Closes #214

### DIFF
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -43,7 +43,7 @@
           "command": {
             "type": "string",
             "title": "Special citation command",
-            "descriptions": "Calls for a non-default citation format",
+            "descriptions": "Specifies a non-default citation format; respectively 'Doe (2019)', '(2019)' and 'Doe'",
             "enum": ["narrative", "suppress-author", "author-only"]
           },
           "uris": {
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -43,7 +43,6 @@
           "command": {
             "type": "string"
             "enum": ["author-in-text", "suppress-author", "author-only"]
-            ]
           },
           "uris": {
             "type": "array",
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -42,7 +42,9 @@
           },
           "command": {
             "type": "string",
-            "enum": ["author-in-text", "suppress-author", "author-only"]
+            "title": "Special citation command",
+            "descriptions": "Calls for a non-default citation format",
+            "enum": ["narrative", "suppress-author", "author-only"]
           },
           "uris": {
             "type": "array",
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -41,10 +41,10 @@
             }
           },
           "suppress-author": {
-            "type": ["string", "number", "boolean"]
+            "type": "boolean"
           },
           "author-only": {
-            "type": ["string", "number", "boolean"]
+            "type": "boolean"
           },
           "uris": {
             "type": "array",
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -40,11 +40,10 @@
               "$ref": "#/definitions/locators"
             }
           },
-          "suppress-author": {
-            "type": "boolean"
-          },
-          "author-only": {
-            "type": "boolean"
+          "command": {
+            "type": "string"
+            "enum": ["author-in-text", "suppress-author", "author-only"]
+            ]
           },
           "uris": {
             "type": "array",
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -41,7 +41,7 @@
             }
           },
           "command": {
-            "type": "string"
+            "type": "string",
             "enum": ["author-in-text", "suppress-author", "author-only"]
           },
           "uris": {


denismaier:
May there combinations? Like author-in-text and author-only?

bdarcus:
> May there be combinations? Like author-in-text and author-only? That would speak for booleans, or for a proliferation of commands. Or for a command property and booleans.

That's the right question, but I _think_ the answer is no.

A combination of author-in-text and author-only would simply be author-only; wouldn't it?

The suppress author and such stuff was largely a workaround to address the lack of intext.

denismaier:
> A combination of author-in-text and author-only would simply be author-only; wouldn't it?

I don't know.

Currently, for an author-date style we have:

parenthetical/default: "(Doe 2020)"
suppress author: "(2020)"

What is author-only? I'd guess "(Doe)"? Or is it "Doe"? 
(Zotero currently doesn't even offer a checkbox for "author-only" so I can't check what it does.)

denismaier:
> We could always have it an array of commands, if we think it feasible there might be need for one-or-more commands?

Sure, or a command with modifiers.

bdarcus:
Ah, maybe I'm wrong; author-only would be "(Doe)" and with author-in-text would be "Doe"?

In any case, this is why I asked the question.

Bibtex/biblatex just use a single command, but they have more of them.

bwiernik:
I'm not sure what you mean by `author-in-text`. Do you mean the form "Jones and Wilson (2008)" (what citeproc-js currently calls "composite": https://citeproc-js.readthedocs.io/en/latest/running.html#special-citation-forms)? 

There are four citation modes:
1. regular
1. suppress-author
1. author-only
1. composite

Currently, citeproc-js defaults to `regular` unless one of three boolean flags is raised to indicate the other modes (I don't know what would happen if multiple flags were set to true). I don't see a compelling reason to have discrete booleans; these should be mutually exclusive.

schemas/input/csl-citation.json
```diff
@@ -40,11 +40,9 @@
...
             }
           },
-          "suppress-author": {
-            "type": ["string", "number", "boolean"]
-          },
-          "author-only": {
-            "type": ["string", "number", "boolean"]
+          "command": {
+            "type": "string",
+            "enum": ["author-in-text", "suppress-author", "author-only"]
```
bwiernik, line 45:
```suggestion
            "enum": ["regular", "author-in-text", "suppress-author", "author-only"]
```

bwiernik (dismissed):

bdarcus:
Yes, that's what I mean; author in the text, rather than in the citation.

AKA narrative citation.

Ok, thanks for the info.

bwiernik:
Should we just call it the "narrative" or "in-text" command?

denismaier:
Narrative seems a good choice.

bwiernik:
Edited my suggestion

schemas/input/csl-citation.json
bdarcus: Why would we include "regular," aka "default," aka what the process should do in the absence of some other command?

If we did that, wouldn't we be requiring the command, else the process would fail with an error?

bwiernik, line 45:
```suggestion
            "type": "string",
            "enum": ["author-in-text", "suppress-author", "author-only"]
```
```suggestion
            "type": "string",
            "title": "Special citation command",
            "descriptions": "Calls for a non-default citation format."
            "enum": ["narrative", "suppress-author", "author-only"]
```

bdarcus:
I added some of the processor developers to the cc list. I'm hoping they can weigh in on the PR, and the questions.

Unlike some issues, this one should be pretty simple to resolve and merge.

cc @fbennett @jgm @andras-simonyi @cormacrelf

schemas/input/csl-citation.json
bdarcus: I'll wait on feedback from some of the others on this; it's one of the questions I added to the main post.

Thanks.

bdarcus:
> Should we just call it the "narrative" or "in-text" command?

Some documentation from existing solutions.

1. natbib has `\citet`, where "t" means "textual"
2. biblatex has `\textcite`
3. CMS 15.25 has no specific term, but says "Where the author’s name appears in the text, it need not be repeated in the parenthetical citation."
4. Not sure what APA has as I don't have access to it ATM, but OWL summarizes like so: "When using APA format, follow the author-date method of in-text citation. This means that the author's last name and the year of publication for the source should appear in the text, like, for example, (Jones, 1998)."
5. (much later) I learned of the distinction between "integral" and "non-integral" citations; this is the former

I prefer "narrative" to "in-text" because the latter is needllessly vague from my perspective, particularly when you match it against the CSL class attribute value of "in-text," which means it's not a note or number citation. E.g. CSL in-text does not mean the same thing as this.

But I don't understand the objection to `author-in-text`; it's literally the purpose of this.

For others, please weigh in. Whatever is easiest for you to understand, and explain to your users (well, ideally, you don't need to explain it; it should be obvious).

bwiernik:
The APA manual refers to in-text citations as "narrative citations" vs the regular "parenthetical citations".


> But I don't understand the objection to author-in-text; it's literally the purpose of this.

I don't have a big problem with it, just find it a little ambiguous in combination with "author-only".

bdarcus:
I guess that raises the question Denis raised: what is author-only?

Sounds like you are assuming (perhaps sensibly) that it is also textual?

If yes, I agree that's an issue.

>

bwiernik:
See https://citeproc-js.readthedocs.io/en/latest/running.html#special-citation-forms

bdarcus:
We could just follow `citeproc-js`, but I wonder then if we would want to rename `cs:intext` to `cs:author-only`, so the symmetry was clear?

denismaier:
Chicago calls it narrative. I just can't check where that is, but I'm quite sure...

bdarcus:
> Edited my suggestion

Make sure, @bwiernik, when you request a change that you remove the flag when either a) the request becomes outdated (I think the case here), or b) the PR author addresses the request.

Anyway, I've changed it to narrative.

I plan to merge this tomorrow (Friday), so will leave this here if any remaining feedback from the devs.

bwiernik:
> Make sure, @bwiernik, when you request a change that you remove the flag when either a) the request becomes outdated (I think the case here), or b) the PR author addresses the request.

Frustrating that GitHub doesn't do that automatically.

andras-simonyi:
FWIW, I like "narrative" for the author in text command, but wouldn't have a problem with "in-text" or "textual" either. I had a look now into what citeproc-el supports currently and, in addition to suppressing authors,  there is an option to suppress affixes, i.e., to produce bare, non-parenthetical citations, and other citeprocs have similar functionality IIRC. Should this option be supported here somehow?

bwiernik:
@andras-simonyi What exactly do you mean there? Can you give an example?

andras-simonyi:
> @andras-simonyi What exactly do you mean there? Can you give an example?

Sure, in  a Chicago-like author-date it would mean producing  `Smith 2020, 13` instead of the regular `(Smith 2020, 13)`. I guess the main use is in contexts which are already 'parenthetical' in some sense, e.g, in parenthetical remarks or footnotes. I think Chicago recommends switching the citation delimiters to square brackets in the parenthetical case: `[Smith 2020, 13]` -- could this be also handled by a command/option?

bdarcus:
Doesn't pandoc have that as well; that I think they call "bare"?

bwiernik:
Sure, I think that would be a good to add. There have been sporadic requests for it by Zotero over the years. Would the behavior be to just remove any punctuation affixes (or all affixes regardless)?

denismaier:
I also think we should add something like this.

One question: couldn't we do it the other way round? Define the bare output without affixes, and then add a wrapper that adds affixes. I mean instead of suppressing affixes.

denismaier:
Regarding the use case brought up by @andras-simonyi, haven't we found another solution for this recently? Some parentheses to brackets flip flop behaviour, similar to quotes to single quotes....

bwiernik:
The normal case for citations is with the affixes, and those vary by style.

Issue #347, opened 2020-08-07
input: move noteIndex property


bdarcus:
I'm confused by [this](https://github.com/citation-style-language/schema/blob/master/schemas/input/csl-citation.json#L98), and think it's a mistake.

A fragment of what I think we want:

``` js 
"citationID": 123,
"citationItems": [],
"noteIndex": 1
```

Correct, @fbennett @bwiernik @cormacrelf?

Because I think what the schema now says is this, which if nothing else is confusing just from a maintenance perspective (given the json schema keyword of "properties").

``` js 
"citationID": 123,
"citationItems": [],
"properties": {
   "noteIndex": 1
 }
```

If OK, then, I'd like to fix this now.

bwiernik:
One benefit of the current structure (aside from maintaining existing structures if there isn't a compelling reason to change them) is that if we wanted to support the compound citations used in Chemistry, an object might provide a good place for the necessary values to be stored.

E.g.,
1. a) First citation; b) section citation; c) third citation
1. a) First citation; b) section citation; c) third citation

bdarcus:
>  ... **an object** might provide a good place for ...

Perhaps, but surely not _this_ object?

I don't really understand that use case, but it seems from your example there's just a prefix on each cite within a single citation, so that prefix number or letter seems better placed within the citationItems data, parallel to the `id`?.

bwiernik:
The structure of citations in chemistry is that multiple cites in a single citation are individually enumerated by letters within a single numeric citation. This is used to refer to a whole batch of thematically-similar items as a set. If they are referenced subsequently, they are generally referred to again all at once (citing just the group number), but they might also be referenced individually, citing both the group number and item letter. 

That's currently difficult to grok against the noteIndex counter structure. 


What exactly is the problem the current structure is causing?

fbennett:
I also don't see why the current structure needs to be shifted. It's a fundamental value for processing that all implementations will be referencing at the current location. Implementations wanting to process both 1.0.1 and 1.1 styles would be faced with either applying conditionals across the code base wherever the ``id`` value is referenced, or subverting the schema change by transforming the input object back to the 1.0.1 form on receipt, to avoid the need for further changes internally. If it were me, I would choose the latter.

bdarcus:
> If they are referenced subsequently, they are generally referred to again all at once (citing just the group number), but they might also be referenced individually, citing both the group number and item letter.

Wow; what a headache.

>What exactly is the problem the current structure is causing?

[...]

PR #348, opened 2020-08-07
input: Move noteIndex to main object properties

bdarcus:
See #347. This implements the simple change.

Also addresses #241 by adding an annotation for `id`.

### DIFF
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -21,33 +21,72 @@
     },
     "noteIndex": {
       "title": "Note Index",
-      "description": "The foot/endnote number or symbol in which the citation is located within the document. Citations within the main text of the document have a noteIndex of zero.",
+      "description": "The end/footnote number (or symbol) in which the citation is located within the document. Citations within the main text of the document have a noteIndex of zero.",
       "type": ["number", "string"]
     },
+    "locator": {
+      "type": "string",
+      "description": "[Deprecated - use the locators object array instead.]"
+    },
+    "label": {
+      "type": "string",
+      "description": "[Deprecated - use the locators object array instead.]",
+      "enum": [
+        "appendix",
+        "article",
+        "book",
+        "canon",
+        "chapter",
+        "column",
+        "elocation",
+        "equation",
+        "figure",
+        "folio",
+        "issue",
+        "line",
+        "note",
+        "opus",
+        "page",
+        "paragraph",
+        "part",
+        "rule",
+        "section",
+        "sub verbo",
+        "supplement",
+        "table",
+        "timestamp",
+        "title",
+        "verse",
+        "volume"
+      ]
+    },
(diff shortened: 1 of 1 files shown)


schemas/input/csl-citation.json
```diff
@@ -15,8 +15,13 @@
       "pattern": "1.1"
     },
     "citationID": {
+      "title": "Citation ID",
+      "description": "A stable and unique ID (for example, a UUID) assigned to the citation, for internal use by the processor.",
       "type": ["string", "number"]
     },
+    "noteIndex": {
+      "type": "number"
```
bdarcus, line 23: Shouldn't this be an integer @fbennett?

fbennett: In theory, at least, the ``noteIndex`` isn't used to determine sequence, it's just the marker associated with the note in which the reference is placed. Again in theory, the marker might be ``*`` (for the author bio in the first note), or numbers might be reset at the start of a new chapter, and ``noteIndex`` should carry those literal values for back-referencing purposes. Sequence (for the purpose of evaluating ``ibid` etc.) can be determined from the order in which citations are submitted to the processor.

Unfortunately (from memory--this might not be totally accurate), this doesn't work out so nicely in practice. IIRC, Zotero's integration plugins supply the *sequence* number of the note in ``noteIndex``, since for technical reasons extracting the note marker from some word processors is difficult. With an initial author bio in a ``*`` note, the value of ``first-reference-note-number`` (derived from ``noteIndex`` is thrown off; and therefore the recommended practice (worked out by @adam3smith) is to set the initial author bio as a footer that spoofs the appearance of a footnote.

For a consolidated doc that resets note numbering at the start of each chapter, the problems are more severe. There is no indication in the citation input of the chapter in which the note occurs, so the behavior of ``first-reference-note-number`` can best be described as "undefined" (I'm not actually sure what happens in such a document, but it wouldn't be pretty).

If the status quo is acceptable, though, constraining the value of ``noteIndex`` to positive integer values would be fine.

bdarcus: Not sure whether we should the change, but we should at least add an annotation. 

This is copy-and-pasted from your docs:

> Indicates the footnote number in which the citation is located within the document.

This is why I was assuming a positive integer, but I see that's actually not the case. 

So maybe we revise it to this, and allow a string too?

> Indicates the footnote number, character, or symbol in which the citation is located within the document.

bdarcus: Also, are you OK with me removing `properties` and putting `noteIndex` alongside `citationID` instead, per #347?

bdarcus:
OK, I've added back the 1.0 properties to better ensure compatibility, but marked them deprecated. We can decide separately when and how to manage that.

I have also added annotations where I thought needed them,, though would like to see add one for "uris" if someone can provide one.

PR #351, opened 2020-08-09
input: Change date range from array to object

bdarcus:
This aligns range representation for dates with ranges in locators (see #326).

``` js
"issued": {
  "begin": { "year": 2000 },
  "end": { "year": 2001 }
}
```

Note: assuming no objections , I'll probably want to fix some of the formatting so it's relatively compact and consistent, along with the test code for dates so CI passes.

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -82,7 +82,7 @@
       "title": "EDTF date string",
       "description": "CSL input supports EDTF, levels 0 and 1.",
       "type": "string",
-      "pattern": "^[0-9-%~X?.\/]{4,}$"
+      "pattern": "^[0-9-%~X?./]{4,}$"
     },
     "date-object": {
       "title": "Date Object",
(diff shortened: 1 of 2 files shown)


schemas/input/csl-data.json
```diff
@@ -82,7 +82,7 @@
       "title": "EDTF date string",
       "description": "CSL input supports EDTF, levels 0 and 1.",
       "type": "string",
-      "pattern": "^[0-9-%~X?.\/]{4,}$"
+      "pattern": "^[0-9-%~X?./]{4,}$"
```
bdarcus, line 85: Prettier did this.

bwiernik: Ah, I see. It's "which flavor of regex" are you using thing. In JavaScript, "/" doesn't _need_ to be escaped, but in PHP it does.

bwiernik:
This looks fine to me.

PR #355, opened 2020-08-12
rename page-range-delimiter and add date-range-delimiter

### LINKED ISSUE #122: Allow range formatting for all numbers (esp. dates)
The `page-range-format` attribute only applies to page numbers (as one would expect), but style guides typically extend these rules to all numbers. This is most notably an issue with date ranges; for example, the following citation (using [Chicago author-date](http://www.chicagomanualofstyle.org/16/ch15/ch15_sec039.html)) is currently impossible to achieve with CSL:

> Karenberg, Axel. 2012–13. ‘The World of Gods and the Body of Man: Mythological Origins of Modern Anatomical Terms’. _Anatomy_ 6–7: 7–22. doi:10.2399/ana.11.142.

I originally proposed on [xbiblio-devel](http://xbiblio-devel.2463403.n2.nabble.com/Expanding-Page-Range-Formats-td7579124.html) to simply extend the `page-range-format` attribute to all numbers, but @rmzelle noted the following:

> My gut feeling is that it's better to have separate definitions of range formatting for dates and all other numbers, so that they can be configured separately. Volume, issue, and series number ranges don't come up very often (in practice and in citation manuals), so we haven't really thought about whether the page-range-format should affect them.

This is an issue, for example, in the [OUP House Style](http://global.oup.com/uk/academic/authors/AuthorGuidelinesMain/HouseStyle/#lev21), where dates use a different formatting scheme from other numbers.

denismaier:
## Description

This PR consists of two changes:

1. rename `page-name-delimiter` to `number-range-delimiter`, which will affect all number variables.
2. add `date-range-delimiter`

partly addresses #122 (issue)

We will need another attribute `collapse-year-range` or `year-range-format` to specifiy the formatting of year ranges.

## Type of change

- [X] Term addition (adds a new term)

## Checklist


bwiernik:
@bdarcus Ready to merge?

Issue #136, opened 2017-02-09
Make year-suffix localizable


rmzelle:
Year suffixes for disambiguation (e.g. "2000a, 2000b") are currently hard-coded in the specification as Latin characters ("a" through "z", "aa" through "zz", etc.). I'm not aware of any other common schemes in English, but we got a report that in Farsi, the Farsi alphabet should be used (see the discussion at https://github.com/citation-style-language/locales/pull/157#issuecomment-277882223). We should check what happens in other languages with non-Latin alphabets, but it might be enough to introduce year suffix terms. We would need at least one term for each alphabet letter, though. Or we could introduce a preference to just select an alphabet.

bwiernik:
I think a preference just for the alphabet would be much better.

hamedheydari:
@rmzelle What's up??!

rmzelle:
@hamedheydari, don't expect this to be implemented soon. We probably first want to check how year suffixes work in some other non-Latin languages like Russian and Chinese.

bwiernik:
Chinese journals usually use latin letters like English.

rmzelle:
@hamedheydari, I have two questions for you. (I uploaded your screenshots to GitHub so they won't get lost in time)

![image](https://user-images.githubusercontent.com/77951/33054590-0e61ab1a-ce72-11e7-8b38-d3ae52f73684.png)

In https://github.com/citation-style-language/locales/pull/157#issuecomment-278203008, you write "And 26 letters (English) or 28 (Abjad) or 32 (Persian) would be more than enough.". But in the example above you restart counting after 26 letters ("الفالف" instead of the English "aa"). Just to make sure, you would normally want to finish out the entire Abjad or Persian alphabet before restarting counting, right?

![image](https://user-images.githubusercontent.com/77951/33054615-39489d5c-ce72-11e7-8d84-916fb968b00a.png)

In https://github.com/citation-style-language/locales/pull/157#issuecomment-277942142, you asked whether it's possible to put an en-dash before the year-suffix. How typical is it to use one of those (or a hyphen or something similar) in e.g. Persian? Always? Sometimes?

hamedheydari:
@rmzelle
1. Yes. But I deleted 2 letters from Abjad from end. Because I was thinking that will be introduced as year-suffix terms, and will be limited to 26 letters.
(In addition, Abjad is preferable to Persian.)
2. Sometimes. By the way this was removed In new style for many unviersities of Iran.

[...]

PR #358, opened 2020-08-12
Make year suffix localizable

bwiernik:
Allows locales to specify the alphabet set to use for year-suffixes.

Fixes https://github.com/citation-style-language/schema/issues/136

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -253,7 +253,7 @@ div {
       ## Alphabet to use for generating year suffixes.
       [ a:defaultValue = "latin" ]
       attribute year-suffix-alphabet {
-        "latin" | "abjad" | "cyrillic" | "greek" | "hebrew" | "persian"
+        "latin" | "arabic" | "cyrillic" | "greek" | "hebrew" | "persian"
       }?
     }
   locale-file.locale =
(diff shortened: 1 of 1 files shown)


denismaier:
I think specifying the alphabet should work.  
Just: Hebrew is also an abjad.

schemas/styles/csl.rnc
```diff
@@ -248,7 +248,13 @@ div {
       ## Specify whether punctuation (a period or comma) is placed within
       ## or outside (default) the closing quotation mark.
       [ a:defaultValue = "false" ]
-      attribute punctuation-in-quote { xsd:boolean }?
+      attribute punctuation-in-quote { xsd:boolean }?,
+      
+      ## Alphabet to use for generating year suffixes.
+      [ a:defaultValue = "latin" ]
+      attribute year-suffix-alphabet {
+        "latin" | "abjad" | "cyrillic" | "greek" | "hebrew" | "persian"
```
denismaier, line 256: Is the order intentional? Perhaps that should be alphabetical?

denismaier, line 256: Why "abjad" and not "arabic"? Just asking because "hebrew" is also an abdjad.

bwiernik: I was just using the term the requester used in the linked issue.

bwiernik: Alphabetical, but with the default alphabet first.

denismaier: Ok. LGTM

bwiernik:
@bdarcus Good to merge?

PR #364, opened 2020-08-12
Add string locator type

bwiernik:
For custom special locator types (e.g., "Psalms 4:3–6"). No label is applied to string. After realizing that pandoc has a way to distinguish these from suffixes (https://github.com/citation-style-language/schema/issues/94#issuecomment-651167640), I don't see a problem with this existing.

Closes https://github.com/citation-style-language/schema/issues/94

### DIFF
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -248,6 +248,11 @@
           }
         },
         {
+          "string": {
+            "$ref": "#/definitions/locator-variable"
+          }
+        },
+        {
           "sub-verbo": {
             "$ref": "#/definitions/locator-variable"
           }
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -145,6 +145,9 @@
     "locator": {
       "anyOf": [
         {
+          "type": "string"
+        },
+        {
           "$ref": "#/definitions/locator-range"
         },
         {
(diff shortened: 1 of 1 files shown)


bwiernik:
@bdarcus Good to merge?

bdarcus:
Actually, lemme tag a few now, to see if we can get feedback here. Do we need a dedicated "string" property in the new locators array, or can we just use ... strings as a fallback?

@cormacrelf @andras-simonyi @paulstanley

bwiernik:
> "locators": [ "Psalm 1" ]

That's not allowed by the current schema. The only allowable items in a `locators` array are `"$ref": "#/definitions/locator"` (or `"$ref": "#/definitions/locators-subgroup"`, which is just a collection of `locator`). 

The only allowable items in a `locator` is `locator-range` or one of the enumerated locator types: https://github.com/citation-style-language/schema/blob/0167c484d86e11ade66be588b1311f3bc85e32f9/schemas/input/csl-citation.json#L145

There is no provision for just entering a string. We could instead add "string" to the "anyOf" for `locator`. That makes more sense to me than having "string".

bwiernik:
Okay, that change is made. This now allows:
```
"locators": [ "Psalm 1" ]
```

Previously that wasn't allowed.

bwiernik:
@bdarcus RTM?

Issue #13, opened 2011-03-19
Move publisher to cs:names


bdarcus:
As has been noted before, the publisher variable can be used with both cs:names and cs:text:

http://sourceforge.net/mailarchive/message.php?msg_id=20080622122017.GB27924%40laptop.nowhere.net

If the processor accepts plain strings on cs:names (treating them as "literal" rather than personal names), this could be fixed in the migration to 1.0.  Zotero and other applications could then shift to treating publisher as a names variable, which would allow more graceful handling of works published by a subunit of a larger institution.

---
- Bitbucket: https://bitbucket.org/bdarcus/csl-schema/issue/15
- Originally Reported By: Frank Bennett
- Originally Created At: 2009-12-29 02:36:48

bdarcus:
Should we add this to 1.1 @bwiernik and @denismaier?

denismaier:
Generally yes.
  
  
   But it think we should also discuss whether we can extend the data model here making `publisher-place` part of the `publisher` variable. That's needed if you have multiple publishers and you want the name if a publisher and the corresponding place appear in the correct order.
  
  
   
    bdarcus <notifications@github.com> hat am 21.06.2020 02:48 geschrieben:
   
   Should we add this to 1.1 @bwiernik and @denismaier?

bdarcus:
Yes, that would be the only reason to do it.

denismaier:
So then, I'd say we should do this.

bwiernik:
What’s the style syntax for rendering publisher and publisher-place then?

bdarcus:
IDK; suggestions?

If we did this, we're talking about rendering of an organization or
enterprise, with a (full) name and a place (location).

Would need to be able to specify that so will format correctly if
multiple publishers (one of the use cases that generated this proposal
to begin with, IIRC).

denismaier:
```
<names variable="publisher" delimiter=": "/>

bwiernik:
Styles vary in whether they require publisher-place or not, so this would require specific new publisher syntax in names to control. I’m really not sure that making publishers/places an array is all that necessary. Do citation styles actually vary in how multiple publishers are listed such that more than a flat field is necessary?

denismaier:
Chicago wants only the first publisher and includes the place. 
SBL wants each publisher together with the place:
X Press: X place; Y Press: Y place

[...]

PR #368, opened 2020-08-18
Make publisher a name variable.

bwiernik:
## Description

Make `publisher` a name variable and add `place` attribute to the name variable schema. This will permit correct rendering of multiple publisher/publisher places.

Existing styles can be programmatically updated to achieve the current output behavior by replacing:
```
<text variable="publisher"/>
```
with
```
<names variable="publisher">
  <name form="long" delimiter="; "/>
</names>
```

and
```
<text variable="publisher-place"/>
```
with
```
<names variable="publisher-place">
  <name form="none"/>
  <place delimiter="; "/>
</names>
```

Fixes https://github.com/citation-style-language/schema/issues/13

## Type of change


schemas/input/csl-data.json
```diff
@@ -243,7 +243,17 @@
...
-              "type": "string"
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/name-variable"
+              }
+            },
+            "place": {
+              "title": "Place associated with the name",
+              "type": ["string", "array"],
+              "items": {
```
bdarcus, line 254: I'm not certain, but I don't think this is right. You may need to use `oneOf` instead.

schemas/input/csl-data.json
```diff
@@ -472,6 +482,13 @@
             "$ref": "#/definitions/name-variable"
           }
         },
+        "publisher": {
+          "description": "Publisher of the item",
+          "type": "array",
```
bdarcus, line 487: Can you include annotations here, including two examples; one with a single publisher, and the other with multiple?

schemas/input/csl-data.json
```diff
@@ -634,7 +651,7 @@
           "type": "string"
         },
         "original-publisher-place": {
-          "description": "[Deprecated - Use `original` related `publisher-place` property instead]",
+          "description": "[Deprecated - Use `original` related `publisher` property `place` attribute instead]",
```
bdarcus, line 654:
```suggestion
          "description": "[Deprecated - Use `original` related `publisher` property `place` property instead]",
```

schemas/styles/csl-names.rnc
```diff
@@ -0,0 +1,352 @@
...
+          ## Use the "and" term "short" form. Not common in English, but is seen in other languages.
+          ## In German, for example: "Drehbuchautor u. Regisseur, where "u." is short for "und".
+          "short"
+      }?
+    }
+  
+  ## Render the "place" associated with a name and specify formatting;
+  ## used to render a publisher's place.
+  names.place =
+    element cs:place {
```
bdarcus, line 276: I don't see why this should be a dedicated element, when everywhere (or almost everywhere?) else in the schema variables are accessed with attributes. 

Can you clarify?

bdarcus (dismissed):
Left some feedback, but was hard to tell the diff since you moved names to a separate file.

Any chance you have a diff you can post of just the changes there?

schemas/input/csl-data.json
bdarcus: Also, annotations?

bwiernik:
> Any chance you have a diff you can post of just the changes there?

Perhaps look at the individual commits? I initially moved the names to a new file without any changes.

bdarcus:
That works; I missed that.

bdarcus:
Actually, no it doesn't. The first commit is that; there's no commit with a diff of the main schema file.

Or if it's straightforward, I suppose you could just tell us what changed on that schema.

bwiernik:
On the main schema file, I just removed the cs:names content and the names-related style options.

schemas/input/csl-data.json
bwiernik: I think you can have an array of allowed types. Restrictions are only applied to the types they are valid for (here, "items" is applied to "array" but not to "string" type values). See https://cswr.github.io/JsonSchema/spec/multiple_types/

We have `["string", "number"]` several places in the schema already.

schemas/styles/csl-names.rnc
bwiernik: It's like `label`, `abbreviation`, or `alternate`--this element controls the position, formatting, and affixes on the place relative to the `name`. This is consistent with how `names` formatting has always worked.

denismaier:
> Make `publisher` a name variable and add `place` attribute to the name variable schema. This will permit correct rendering of multiple publisher/publisher places.

In addition, this PR adds a new option for `@form` ("none"), right? Anything else? Maybe you could add that to the initial comment.

bwiernik:
> > Make `publisher` a name variable and add `place` attribute to the name variable schema. This will permit correct rendering of multiple publisher/publisher places.
> 
> In addition, this PR adds a new option for `@form` ("none"), right? Anything else? Maybe you could add that to the initial comment.

Update.

> Also:
> 1. adds new cs:name @form "none" to omit name and et-al elements from a names citation (e.g., to permit only the publisher-place to be rendered)
> 1. adds @form attribute to cs:et-al. form="none" suppresses the et-al term. The options here are consistent with the same attribute used in cs:citation-label-format

denismaier:
Maybe I've missed it, but I've only seen delimiters between names and between places. But how do you actually specify the colon between the place and the publisher? Simply an affix on the name element?

schemas/styles/csl-names.rnc
denismaier: How the affixes? I don't see them.

denismaier: By the way, I also don't see them on `cs:label`. What am I missing?

denismaier (changes requested):

bwiernik:
@denismaier Yes, like with a label.

denismaier:
> @denismaier Yes, like with a label.

I've tested with the 1.1 branch:

```xml
<names variable="author">
  <label suffix=". "/>
</names>
```

Jing complains about the suffix on `label`. "Attribute 'suffix' not allowed here."

denismaier:
If I remember correctly, 1.0.1 had affixes on `cs:label`. Why have we removed them?

bwiernik:
That's a bug. `label.attributes-shared` should be part of the names.label list.

bwiernik:
@denismaier That's fixed now.

denismaier:
But it is still only on label, not on place, right? Is that on purpose?

bwiernik:
Nope! Fixed.

bwiernik:
I added all of the formatting attributes. Any that place shouldn't have? Only one I'm not really sure about is strip-periods?

```
      affixes,
      font-formatting,
      strip-periods,
      text-case,
```

denismaier:
Not sure. 
On the other hand, do we need an attribute to specify state formatting rules. "Cambridge/MA" vs "Cambridge (MA)"

bwiernik:
No, we've never had something like that before where we edit the contents of a string. I suggest we not do that now.

bwiernik:
@bdarcus RTM?

schemas/styles/csl-names.rnc
bdarcus: > It's like `label`, `abbreviation`, or `alternate`--this element controls the position, formatting, and affixes on the place relative to the `name`. This is consistent with how `names` formatting has always worked.

I haven't been following the name stuff closely, but I'm not sure I buy this ATM.

Are we saying whenever we add a new variable to names we need a new CSL element, whereas everywhere else in the schema we'll use an attribute value?

bwiernik: You're just not appreciating how the `names` element works (and has always worked). `names` defines a layout for name variables. The order, formatting, and spacing of its components have always been controlled by sub-elements of `names`. This is great because there is huge diversity in the placement, spacing, and formatting of name parts, labels, et-al, etc. I assume this is why you designed it this way in the first place. 

Placement of various pieces within `names` has never been done using attributes.

Previously, the sub-components of names were label, name, and et-al. We've added abbreviation and alternate. This is adding place. place is just like label in that styles vary in whether it is included, its position, its spacing and punctuation, its text formatting, etc. Adding place as an element within is entirely consistent with the existing design of `names`. 

To make place now be an attribute would be to make it. **unlike** every other part of name formatting. It would require a slew of new attributes: place-position, place-name-delimiter, place-text-case, place-font-style, etc. This just isn't how we control formatting of components within `names`.

bdarcus: I think I need to clarify that I meant by "attributes".

I do NOT mean on the names element; I mean on some sub-element, such that it can be used to hold other variables in the future; like (don't take this as a specific recommendation; it's just an illustration):

``` xml 
<name-part variable="place"/>
```

denismaier: Just that `name-part` currently doesn't render variables. It only controls the formatting of a name part. Rendering itself is controlled via `@form` on `cs:name`. 

But I get your point, `place` is not really the same thing as a label since it's not a static string or so, but provided in the data.

What we could do:
- control the rendering and placement of institutional name parts with a new attribute on `cs:name`; `@institutional-form` or so. (Or, we could just allow a list of values as options for `@form`; with some values applying to personal names, others to publishers, e.g. `name form="long institutional-without-place"`.)
- specifiy the delimiter via `@institutional-name-part-delimiter`
- use `name-part name="place"` to control the styling.

Or:
Completely revise the current design to a model where `name-part` actually does control whether a `name-part` is rendered,-i.e. get rid of form here.

bwiernik: I really don't think we should deviate from the existing general structure and behavior of `names`. 

The operating behavior of `names` is that it renders the name variable, with the name parts rendered controlled by `@form` and the usage of et-al controlled by `@et-al-min`. The `name-parts` and `et-al` elements only control formatting, not which elements are rendered. 

In contrast, `label` only appears if it is explicitly called. The relative position of `label` and `name` controls the layout of the two pars. That is the key difference between name-parts/et-al and label—the element controls its appearance and its layout. In my view, `label` comes just as much from the data as name-parts do—it's supplied by the variable type (editor, performer, etc.). This is even more clear if we allow user-entered labels to override the defaults (https://github.com/citation-style-language/schema/pull/337). 

So, we have two types "name parts"--those who always appear (the `name-parts` that are children of `name`: family, given, dropping-particle, non-dropping-particle, suffix, institution; `et-al`), and those whose appearance is controlled by the presence of an element (`label`, `abbreviation`, `alternative`, and proposed `place`). 

Following the precedent of `label` getting its own element (and the apparently non-controversial previous additions of `abbreviation` and `alternative`), I suggest we stay consistent and use an element for `place`. I don't really see the advantage of having `<supplemental-name-component="place"/>` or `<supplemental-name-part="abbreviation"/>` versus
`<place/>` or `<abbreviation/>`. That makes styles less readable and is less consistent with `<label/>` being its own element already (which I don't think we should change). 

Additionally, looking at it more closely again, `<alternative/>` does need to be an element because it can take all of the children of `name`. So, if we went with the `<supplemental-name-component="place"/>` structure, then we would have two name components with their own elements (label and alternative) and two name components that are called as attributes on an element (abbreviation and place). I would much prefer to be consistent and use the same structure we have used for `label` consistently.

If we did take the common element approach, we would need a new element other than `name-part`, as that is already used for the specific task of controlling formatting (but not appearance) of name-parts (family, given, suffix, particles, institution). I suggested `supplemental-name-component`. That's obviously very verbose. We could do something like `supplemental-part`. But again, this seems like unnecessary departure from the existing design of `names`. 







 There is already precent for such an element in `names` formatting--

bwiernik:
@bdarcus @denismaier I reworked the names structure to use a single `cs:names` element for all of place, abbreviation, and alternate. Can you look over the csl-names.rnc schema again?

denismaier:
@bwiernik Can you please give a brief summary of your changes? And perhaps a couple of examples.

bwiernik:
The current state: If a style wants to show an institution abbreviation, an alternative name (e.g., a pseudonym or screenname), or a place, they would add a `cs:component` element as a child to `cs:names`.

This is a change from the previous version, where each of these three had separate child elements on `cs:names`. 

So, for example, to render publisher and publisher place:
```
<names variable="publisher">
  <component component="place" suffix=": "/>
  <name form="long"/>
</names>
```

So, to render an institutional name in long form with the abbreviation the first time, then just the abbreviation subsequently:
```
<names variable="author">
  <name>
    <name-type name="personal" form="short"/>
    <name-type name="institutional" form="long" form-subsequent="short"/>
  <name/>
   <component component="abbreviation" prefix=" [" suffix="]"/>
</names>
```

To render a name with its alternative form:
```
<names variable="author">
  <name form="long" name-as-sort-order="all"/>
  <component component="alternative" prefix=" [" suffix="]"/>
</names>
```

A single `cs:names` element might have multiple of these, such as both abbreviation and alternative.

bdarcus:
In general and at first glance, looks good ... except for that
component/@component duplication.

Is that correct, or a typo?

bwiernik:
> In general and at first glance, looks good ... except for that
component/@component duplication.
> 
> Is that correct, or a typo?

Not following. The name of the element is `component`, and it has a `component` attribute indicating which component to render.

bdarcus:
A component of a component, same node name for element and attribute, makes
no sense.

Maybe we need a different name for that attribute?

I'll take a closer look later to see if I have a concrete suggestion.

bwiernik:
I really don't see the problem. The element is "Add another component". The attribute is "which component should we add?". It's not ambiguous. We could rename the attribute `which` or `which-component`, but why make it more ambiguous or verbose?

denismaier:
Ok, this change does not affect the general behaviour here, or does it? It's only that there's now one element instead of multiple, correct?

Anyway, it's hard for me to get a grasp of this using the web interface. Is it possible to checkout the PR or a certain commit? I could do some tests then in with Atom or Oxygen. @bdarcus ?

bwiernik:
Exactly. 

See here for how to check out a PR http://stackoverflow.com/questions/27567846/ddg#30584951

bdarcus:
Where do we stand with this one?

I will plan to follow @denismaier's thought to actually test it soon.

bwiernik:
@bdarcus Is this good to merge now?

bdarcus:
> @bdarcus Is this good to merge now?

Can you just respond to my last note, then I can test it tomorrow.

IIRC, the only issues was the name/name node name issue.

bwiernik:
I resolved the component element/attribute issue yes.

bdarcus (changes requested):
I added a few minor requests. In the meantime, I'll spend some time testing the rnc schema changes.

EDIT: I just tried to load the csl.rnc file in nxml-mode, and got the following error:

``` console
nxml-display-file-parse-error: Reference to undefined pattern names-inheritable-options
```

Is it because you don't include the `csl-names.rnc` file?

When you resolve, can please indicate that you've done so by clicking whatever github button. I forget how to actually do so, but I think it's straightforward.

schemas/styles/csl-names.rnc
```diff
@@ -0,0 +1,388 @@
...
+      ## default to the values specified on the sibling cs:name
+      name.forms,
+
+      ## If name="alternate", specify formatting attributes for names;
+      ## default to the values specified on the sibling cs:name
+      name.attributes,
+
+      ## If name="alternate", use to apply separate formatting to individual name parts or to personal vs. institutional names 
+      name.name-part,
+      name.name-type
```
bdarcus, line 307:
```suggestion
      name.name-type }
```
I think this is the change you need for the one error, but then I get others.

Unless I'm doing something wrong with git, maybe double check the syntax problems and let me know when I can try again?

bdarcus:
Status of this PR?

PR #375, opened 2020-09-05
Style syntax for related variables

### LINKED ISSUE #357: Style syntax for @related variables
We need to write the style syntax for calling related items.

denismaier:
## Description

This adds syntax for calling related variables.
Overall, I think this should be a simple change. The only shortcoming I see is that it won't be possible to call name variables on different levels from a single `cs:names` element. (Don't know if that's a real problem.)

Closes #357 

## Type of change

## Checklist

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -777,10 +777,12 @@ div {
       ## Specify verbatim text.
       attribute value { text }
     | ((attribute variable { variables.numbers | variables.strings },
-        [ a:defaultValue = "long" ] attribute form { "short" | "long" })
+        [ a:defaultValue = "long" ] attribute form { "short" | "long" },
+        related?)
        | (attribute variable { variables.titles },
           [ a:defaultValue = "long" ]
-          attribute form { "short" | "long" | "sub" | "main" })?)
+          attribute form { "short" | "long" | "sub" | "main" },
+          related?)?)
 }
 # ==============================================================================
 
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -414,6 +414,7 @@ div {
     attribute variable {
       list { variables.names+ }
     },
+    related?,
     affixes,
     
     ## Specify the delimiter for name lists of name variables rendered by
@@ -690,6 +691,7 @@ div {
         date.form,
         rendering-element.date.date-part.localized*)
        | (rendering-element.date.date-part.non-localized+, delimiter)),
+      related?,
       affixes,
       display,
       font-formatting,
@@ -805,7 +807,8 @@ div {
     ## Number forms: "numeric" ("4"), "ordinal" ("4th"), "long-ordinal"
     ## ("fourth"), "roman" ("iv").
     [ a:defaultValue = "numeric" ]
-    attribute form { "numeric" | "ordinal" | "long-ordinal" | "roman" }?
+    attribute form { "numeric" | "ordinal" | "long-ordinal" | "roman" }?,
+    related?
 }
 # ==============================================================================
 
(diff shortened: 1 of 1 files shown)


bwiernik:
I don't think that's a real problem. Rendering both standard and related names just requires a group in that case, and allowing otherwise would be messy with the multiple="combine" attribute.

bwiernik (approved):

denismaier:
Another problem: if we keep 
`original-title`, what happens if you call `text variable="original-title" related="original"`.
Probably related to my question on testing syntax.

bwiernik:
I think that would just be empty---I don't think anyone would write that and processors don't need to design any special case around that.

denismaier:
We could disallow it on styles. But that would be complicated. But maybe it's n9t necessary.
What would you think about another syntax: `text original-variable="title"'` instead of `@related`? (That wouldn't work on `cs:group` though...)

bwiernik:
@related is much more extensible for the future.

denismaier:
So, you say better leave it as is?

bwiernik:
Yes.

denismaier:
I've now also added test syntax.

Also, I've changed the wording from `@related` to `@relation`. Ok with that?

schemas/styles/csl.rnc
```diff
@@ -1136,6 +1144,12 @@ div {
   always-render = attribute always-render { xsd:boolean }?
 }
 
+## Select a related-variable
+div {
+  relation = attribute relation { relations }?
```
bwiernik, line 1149:
```suggestion
  related = attribute relation { relations }?
```

bwiernik (changes requested):

schemas/styles/csl.rnc
denismaier: I guess you want me to change the attribute name as well...

bwiernik: No, everywhere else in CSL.rnc you refer to "related" but never define that variable. I don't care much what variable name is chosen. The attribute name is fine.

denismaier: Hmmm. I've thought I had fixed that...  maybe I'll need to update my branch

denismaier: Oh no. I've only changed it in csl-choose.rnc

schemas/styles/csl-choose.rnc
```diff
@@ -11,8 +11,8 @@ div {
   
   ## When specifying conditions, we allow either the attribute syntax or the element-based one.
   choose.condition =
-    (condition.atts, relation?, match?)
-    | element cs:conditions { match?, relation?, condition.elem+ }
+    (condition.atts, relation, match?)
+    | element cs:conditions { match?, relation, condition.elem+ }
```
bwiernik, line 15: Wait, why is this non-necessary? If I want to test conditions for the base item, I would omit `@relation`.

bwiernik (changes requested):

schemas/styles/csl-choose.rnc
denismaier: The pattern for relation is this:
```
relation = attribute relation { relations }?
```
So, it's always optional.

bwiernik: Okay

bwiernik (approved):

schemas/styles/csl-choose.rnc
denismaier: Here's the definition for `@form` (in `names.form`)

```
 attribute form { "long" | "short" | "count" }?
```

On `cs:names` it appears only as `names.form`. No need to make it optional twice.

bdarcus:
Sorry I've been MIA. At some point (hopefully soon) I'll get back to these.

I note this has conflicts that need resolving, though.

bdarcus:
Just bumping this. Where does it stand?

I am assuming, @denismaier, the conflicts will need to be resolved by rebasing this branch on v1.1. See what happens if you do this from the branch dir:

```console
git rebase v1.1
```

PR #384, opened 2020-11-02
Allow cs:updated to be date or dateTime

bwiernik:
## Description

Suggested by @cormacrelf
> It's a small pain point but the format allowed in cs:updated is an xsd:dateTime when it should probably have been an xsd:date. One shouldn't have to look up xsd:dateTime and type in a bunch of zeroes and colons in the right order to make CSL validate. For a field that really, really does not need timestamp precision.
>
>Simple solution: allow either. Then people can update with 2020-05-19, much more human friendly.

I don't see much issue with allowing just a date. Would that interfere with anything @dstillman?

Re: https://github.com/citation-style-language/documentation/issues/129

## Type of change

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -197,9 +197,8 @@ div {
     element cs:title-short { info-text }
   info.updated =
     
-    ## Specify when the style was last updated (e.g.,
-    ## "2007-10-26T21:32:52+02:00")
-    element cs:updated { xsd:dateTime }
+    ## Specify when the style was last updated (e.g., "2007-10-26" or "2007-10-26T21:32:52+02:00")
+    element cs:updated { xsd:dateTime | xsd:date }
   info-text =
     attribute xml:lang { xsd:language }?,
     text


dstillman:
> New feature (non-breaking change which adds functionality)

First, this would be a breaking change. It can't really be assumed that clients don't rely on this to be in the previously specified format.

But I don't think we should change this.

It's not accurate to say that this field "really, really does not need timestamp precision". The timestamp serves primarily as a mechanism to check for style updates. While we don't require CSL authors to update it when they make changes, it's updated automatically to the git commit timestamp in both the CSL styles-distribution repo and the Zotero styles repo/page. When Zotero checks for updated styles, it sends the current `updated` value for each installed style, and the repo returns styles with more recent timestamps. If this only used date precision and there were two updates to a style in the same day, either the latter one would be missed (`> lastUpdated`) or the style would have to be redownloaded for every check that day (`>= lastUpdated`).

Now, since the request here is to make the time portion optional, the official repos could continue doing the above, and unofficial styles would likely be updated using standard HTTP mechanisms where the field would be irrelevant. But I just don't think it's an issue that's worth adding variability to the field. People shouldn't be starting CSL styles from scratch at this point — they should copy and paste from an existing style or use an authoring tool that generates the metadata fields — and once a timestamp exists, it's trivial to change it if you really need to (which, again, you don't for official styles).

bwiernik:
I was thinking similarly.

PR #392, opened 2021-06-16
bug fix for 1.0/1.0.2: make @form optional again

denismaier:
With this change `@form` is again optional. With that change Chicago-author-date.csl validates successfully again. (I don't know if there are other leftovers from 1.1 in the 1.0 branch.)


bwiernik:
Can you do a comparison against the v1.0.1 tag and be sure that the only differences are types/variables/terms as expected? https://github.com/citation-style-language/schema/releases/tag/v1.0.1

denismaier:
Argh, I've opened some changes from master... I'll close and open a new PR

PR #396, opened 2021-09-01
Restore titles allowing simple string values

cormacrelf:
Targeting v1.1.

## Description

Fixes title variables not allowing plain strings, but being documented being string or title-object. There was a `title-string` definition but it was unused, so title-variable is now an anyOf union.

See https://github.com/citation-style-language/schema/pull/393#event-4911844030 for general discussion of deprecating instead of outright removing.

## Type of change

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,19 +1,24 @@
 ---
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.5.0
+    rev: v4.0.1
     hooks:
       - id: check-merge-conflict
-  - repo: https://github.com/prettier/prettier
-    rev: 2.0.5
+      - id: check-json
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.3.2
     hooks:
       - id: prettier
-        types: [json,yaml]
+        name: Prettier JSON
+        types: [json]
+      - id: prettier
+        name: Prettier YAML
+        types: [yaml]
   - repo: local
     hooks:
       - id: rnc-format-validate
         name: Check rnc schema files
         entry: tools/rnc-validate-format.py
-        args: ['--commit']
+        args: ["--commit"]
         language: python
         files: ^schemas/styles/csl*.rnc$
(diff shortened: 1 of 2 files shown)


bdarcus:
Per my comment here, I think we need to settle a broader strategy before deciding particular changes.

https://github.com/citation-style-language/schema/pull/397#issuecomment-910093763

bdarcus:
While I'm hear, thought I'd circle back on this.

Where do we stand on this?

bwiernik:
My original position was that I wanted to permit string values and have parsing main vs sub be a part of the processor steps in that case. That got pushback. I would like to have that option--a preferred structured approach and a string-parsing alternative

bdarcus:
> My original position was that I wanted to permit string values and have parsing main vs sub be a part of the processor steps in that case. That got pushback. I would like to have that option--a preferred structured approach and a string-parsing alternative.

I'm still strongly opposed to adding an entire title parsing section to the CSL spec, and new elements and/or attributes to the style schema, in order to support independent formatting of titles and subtitles. 

Whether there's a way to do it without that I'm not sure (though don't think there is). Or maybe @cormacrelf has a practical way to do this that isn't as painful as I foresee.

But I think this needs input from @cormacrelf and the zotero team (cc @dstillman) with regard to their plans vis-a-vis CSL going forward. And of course, if there are other developers/implementers reading this, the more feedback the better.

I know @cormacrelf has done work which contributes towards supporting what we've been calling v1.1, and this (along with the ETDF date library) is actually part of it, but it's just hard to have these discussions in the abstract, without an understanding of concrete roadmaps and such.

For brief context for those new to this, the big changes in 1.1 are in title and dates.

- **dates**: promotes ISO 8601-2 (EDTF) as the primary input date format; a string, but an international standard that is an extension to the 8601 date-time standard; also happens to be the default format in biblatex
- **tittles**: changes from string to object, so that the object matches the new CSL styling support for titles and subtitles.

The questions here are of course about WTD with the latter.

bdarcus:
I've just tagged this waiting-on-reply. I'm not merging this until I get one from @cormacrelf.

PR #397, opened 2021-09-01
Restore title-short variables for 1.0 compat

cormacrelf:
## Description

Like #396. Again see commentary in https://github.com/citation-style-language/schema/pull/393 where these were added back to the RNC. Adding back to the JSON schema is a slightly different proposition, but if anything I think it's a stronger one. This addresses the durability of systems that produce CSL-JSON for consumption by a processor. I estimate these are significantly less nimble than style authoring, particularly because these are systems that are essentially never written by hand, whose output is never seen by human eyes, and whose errors require much deeper debugging. Even where they are (CSL test suite), that's a lot of work to update. CSL processors, *including newly written ones*, are going to have to support CSL-JSON 1.0 for a long while yet, so I think this should be reflected in deprecation rather than removal.

## Type of change

## Checklist

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -573,8 +573,16 @@
         "collection-title": {
           "$ref": "#/definitions/title-variable"
         },
+        "collection-title-short": {
+          "description": "[Deprecated - use `collection-title` with a `short` key",
+          "type": "string"
+        },
         "container-title": {
           "$ref": "#/definitions/title-variable"
+        },
+        "container-title-short": {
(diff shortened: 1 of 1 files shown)


bdarcus:
Thanks.

 It seems you never did reply to this comment, though, and subsequent discussion:

https://github.com/citation-style-language/schema/pull/393#issuecomment-863220880

I'm not sure where and how best to pick up that thread, but this is a pretty important question of strategy.

Maybe best to reply there, and point people here?

bdarcus:
And just double-checking this: OK to merge, given other discussions?

cc @bwiernik @denismaier

PR #403, opened 2021-12-06
Add "none" option for 'and' attribute

### LINKED ISSUE #402: Author lists in bibliography without "and" or "?"
I was working on a CSL style for a journal that requires to have the bibliography in the following format:

Baker, A.J., Brown, P,T, Bullock, J.T, 2020, Title of the manuscript, ...

I could not find any solution that would allow me to achieve this format. As I understand it, this is because in the CSL schema `and` can only hold the values `text` and `symbol`.

Would it be possible to add a third option like `and=none`?

**Describe alternatives you've considered**
I can set `and=""`, which works out for me locally, using Zotero and Libreoffice, however it doesn't comply with the CSL schema.

**Additional context**
See the author instructions here: http://boletinsgm.igeolcu.unam.mx/bsgm/index.php/instrucciones-instructions#new

Best wishes,
Mario

bdarcus:
Tasks:

1. [X] moved two separate duplicate "and" patterns into a separate one
2. [X] add "none" as option
3. [X] run formatting
4. [ ] test

Close #402.

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -411,6 +411,24 @@ div {
        & names.alternate?),
       names.substitute?
     }
+  
+  ## Specifies what to add between the second-to-last and last item in a name or label list.
+  and.attribute =
+    attribute and {
+      
+      ## Use the "and" term "long" form (e.g., "Doe, Johnson and Smith").
+      "long"
+      | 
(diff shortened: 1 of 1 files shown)


bdarcus:
OK, change made, but still needs testing and review.

Issue #395, opened 2021-08-19
add ability to remove protocol from URL 


eroux:
**Is your feature request related to a problem? Please describe.**

MLA Handbook 8th edition says in 2.5.2 URLs and DOIs:

> When giving a URL, copy it in full from your Web browser, but omit http:// or https://.

but `modern-language-association.csl` can only copy the URL that is given as argument. 

see
- https://github.com/citation-style-language/styles/issues/5545
- https://discourse.citationstyles.org/t/new-way-to-specify-unprefixed-urls-in-styles/1731

**Describe the solution you'd like**

Add an `npurl` to the list of link renderings:

https://github.com/citation-style-language/documentation/blob/master/specification.rst#appendix-vi-links

that would remove the protocol from a url

*For changes to the CSL RNC schemas, this should include links to citation style guides.* It's better to include links to more than one style guide, and no more than five, for us to assess how widely needed this feature might be.

MLA Handbook 8th edition (published book, not openly available online)

bwiernik:
This is a quirky rule that is unique to MLA as far as I am aware. 

I personally don't really understand why MLA bothers with it and my standard preference is that more clarity rather than less is preferable. For example, many publishers are named "something.com" and including the protocol more clearly distinguishes what is an actual URL where the item is located. 

Given the rarity of this rule, I'm not sure it's worth the additional complexity to ask citation processors to include an option to strip the protocol from a web link. 

What do you think @denismaier @bdarcus @adam3smith ?

bdarcus:
@eroux 

> Add an npurl to the list of link renderings:

So you are not proposing schema changes, but instead for this to be handled in the data? 

If yes, that seems less-than-ideal, given that it's style dependent?

> What do you think @denismaier @bdarcus @adam3smith?

IDK. As you are aware, I'm not fond of many citation rules.

adam3smith:
I'm also skeptical: I have searched around a bit and haven't found a single example of this in a citation style other than MLA

eroux:
well, don't count on me to defend the idea of removing the protocol, I find it quite silly to be honest... but for better or worse, MLA is quite popular in the humanities so that would still be helpful to a number of users

bdarcus:
My worry is where it stops?

Stripping the protocol seems straightforward enough, but are people then
going to ask for other iterations of this?

bdarcus:
BTW, they promote this on twitter; a recent tweet pointing to this:

https://style.mla.org/citing-twitter/

[...]

PR #405, opened 2021-12-30
Strip protocol

bwiernik:
## Description

Adds a style attribute specifying that `https://` and `http://` should be stripped from rendered URLs. 

Fixes [# (issue)](https://github.com/citation-style-language/schema/issues/395)

## Type of change

## Checklist

### DIFF
--- a/schemas/styles/csl.rnc
+++ b/schemas/styles/csl.rnc
@@ -853,6 +853,7 @@ div {
     style.demote-non-dropping-particle,
     style.initialize-with-hyphen,
     style.page-range-format,
+    style.strip-url-protocol,
     title-inheritable-options,
     names-inheritable-options,
     name-inheritable-options,


cormacrelf:
Implementation note: Technically you can come across FTP links, however rare these days (and browsers have stopped supporting them for whatever reason). It's possible we'll have academia using IPFS or something soon enough too. The appropriate way to deal with this is to have a whitelist that it applies to (not configurable per style), not to do it blindly. FTP and IPFS links should not have the protocol stripped because they're not implied by overwhelming common usage like http/https are.

Http/https should be the only ones. I'd also strip `tel:` and `mailto:` regardless of the flag, but not very important.

bwiernik:
Yep, this would only apply to http/https

denismaier:
That's not ready for review, right?

bwiernik:
It's ready for review. It's just adding the one style attribute we discussed in the linked issue

denismaier:
But this commit only adds this to the list of possible options on `cs:style`. However, I think it does not define possible values for that option. Cf., e.g., this:

```
style.initialize-with-hyphen =
    
    ## Specify whether compound given names (e.g., "Jean-Luc") are
    ## initialized with ("J-L") or without a hyphen ("JL").
    [ a:defaultValue = "true" ]
attribute initialize-with-hyphen { xsd:boolean }?
```

So, I'd think something similar would be needed:

```
style.strip-url-protocol =
    
    ## Specify whether http/https should be stripped from urls
    [ a:defaultValue = "true" ]
attribute strip-url-protocol { xsd:boolean }?
```

Should the default value be true or false?

bdarcus:
FWIW, relax ng doesn't truly support default values.

We include them for documentation purposes, where I think it does help add to the clarity.

bwiernik:
@denismaier ah, right. Can you add that? Default to false

denismaier:
I don't know how to commit directly to your PR, but I'll add this as a suggestion.

PR #415, opened 2022-04-20
Add validation for et-al-min and et-al-use-first

georgd:
## Description

See discussion in https://github.com/citation-style-language/documentation/pull/141

## Type of change

- [ ] Variable addition (adds a new variable or type string)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [ ] I have installed the repo [pre-commit hook](https://github.com/citation-style-language/schema/blob/master/CONTRIBUTING.md#pre-commit-hooks); if not, and I have modified any of the schema files, I have run trang and/or prettier on the files, per [CONTRIBUTING](https://github.com/citation-style-language/schema/blob/master/CONTRIBUTING.md)

### DIFF
--- a/schemas/styles/csl.sch
+++ b/schemas/styles/csl.sch
@@ -11,5 +11,8 @@
     <sch:rule context="/cs:style/cs:macro">
       <sch:assert test="count(/cs:style/cs:macro/@name[. = current()/@name]) = 1">This macro does not have a unique name.</sch:assert>
     </sch:rule>
+    <sch:rule context="*[@et-al-min]">
+      <sch:assert test="number(current()/@et-al-min) gt number(current()/@et-al-use-first)">et-al-min must be greater than et-al-use-first.</sch:assert>
+    </sch:rule>
   </sch:pattern>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks!

@denismaier @bdarcus How should we handle bug fixes like this? As x.x.x.1 point releases?

denismaier:
Good question, I've been thinking about this as well recently. (I've been pondering whether we could be more lenient when adding new terms. As this doesn't seem to affect processors they could probably added to the existing schema as well.)

Anyway, one option I was thinking about was to use a timestamp for these kind of additions, similar to emacs package versioning on melpa: e.g. https://melpa.org/#/citar

Like, if we added this today we'd still be on CSL 1.0.2, but we'd add a timestamp "20220421" or so to indicate these additions.

But, I can't really estimate how that might affect the larger ecosystem.

bdarcus:
Seems this got stalled. Where do we stand?

PR #416, opened 2022-05-01
add descriptions from spec

FelixBenning:
## Description

This adds some of the descriptions from https://docs.citationstyles.org/en/stable/specification.html#appendix-iv-variables to the JSON Schema

It took me some time to find the spec and it would be nice to have tooltips in vscode by simply using the `$schema` key in the json.

## Type of change

## Checklist

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -464,7 +464,12 @@
       "status": {
         "title": "Publication status of the item",
         "type": "string",
-        "examples": ["forthcoming", "in press", "advance online publication", "retracted"]
+        "examples": [
+          "forthcoming",
+          "in press",
+          "advance online publication",
+          "retracted"
+        ]
       },
       "supplement": {
         "type": ["string", "number"]
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -60,6 +60,7 @@
         "type": ["string", "number"]
       },
       "citation-key": {
+        "description": "Identifier of the item in the input data file (analogous to BibTeX entrykey);\nUse this variable to facilitate conversion between word-processor and plain-text writing systems;\nFor an identifer intended as formatted output label for a citation (e.g. “Ferr78”), use citation-label instead",
         "type": "string"
       },
       "categories": {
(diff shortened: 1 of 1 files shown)


schemas/input/csl-data.json
```diff
@@ -234,21 +236,28 @@
...
+        "title": "Date the item has been accessed",
         "$ref": "#/definitions/date-variable"
       },
       "available-date": {
+        "title": "Date the item was initially available",
+        "description": "e.g. the online publication date of a journal article before its formal publication date; the date a treaty was made available for signing",
         "$ref": "#/definitions/date-variable"
       },
       "event-date": {
+        "title": "Date the event related to an item took place",
```
bdarcus, line 248: Thanks for doing this; we intended to, but just hadn't figured the best approach.

The only obvious question I have is why some properties have "descriptions", some "titles" and others both?

FelixBenning: Both are used for tooltips. If I remember correctly (haven't worked with json schemas for some time) vscode displays the title above the description on hover if both are there. And one or the other are displayed if only one of them are there. So in some sense they are equivalent. So I tried to use "title" for short "titley" descriptions like DOI, "Digital Object Identifier". And use "description" for longer descriptions. But there is no system behind it just rule of thumb. Basically took 10-20 min to throw as many descriptions into the file as I could find in that time. It won't be complete either. But I did not want to spend too much time before getting a reaction about it.

FelixBenning: Later on I started to split the description into a more punchy title and the remaining description. But not from the start so that is why that only happens sometimes

FelixBenning: The order in the documentation is not the same as in the schema, so it was sometimes hard to match the stuff. Ideally you would generate one from the other. Probably the docs from the json schema, since it is easier to read json programmatically

bdarcus: Here's from the json schema docs:

> The title and description keywords must be strings. A “title” will preferably be short, whereas a “description” will provide a more lengthy explanation about the purpose of the data described by the schema.

On this:

> Ideally you would generate one from the other. Probably the docs from the json schema, since it is easier to read json programmatically.

Yes, this is what I was alluding to above on the "best approach."

But it's even more complicated, given the overlap with the csl style schema variables.

bdarcus: As I've been using VSCode completion as part of #420 ...

> If I remember correctly (haven't worked with json schemas for some time) vscode displays the title above the description on hover if both are there. And one or the other are displayed if only one of them are there.

Not from what I can see. It seems if a description is present, it uses that, even if there's a title.

But I can certainly imagine both being useful; for example, a title could be used to extract human-readable documentation. Or some editors might use both?

PR #420, opened 2022-06-02
input: Convert root into object, require version

### LINKED ISSUE #319: Convert input schema to an object with metadata, add id property
We should convert the input schema from an array to an object, with top-level metadata properties:

- version (required)
- title
- description

bdarcus:
Convert the root of schema into an object with metadata, and a required
`version` property.
    
Include optional `title` and `description` properties.

While I'm at it, also removing the edtf regex on a separate commit.
    
Close #319

-------

## Examples

Aside from this being a better, forward-looking, design, it's also more consistent with the pandoc YAML model.

The below shows this change, but in larger context of the other changes suggested in the `v1.1` branch; notably with dates and titles.

An example, validated against the schema in YAML, with EDTF dates (though the regex here doesn't cover the date-time example ATM, so the validator throws an error on that property; really need a proper EDTF regex, which is not my skillset, or maybe just drop the constraint for now; see #421).

```yaml
---
version: 1.1
description: An example CSL YAML 1.1 file
references:
  - type: book
    id: test1
    issued: '2020'
    title: 
      main: Title
      sub:
        - Subtitle
    author:
      - given: Jane
        family: Doe
  - type: article-journal
    id: test2
    issued: 2020-07/2020-08
    title: Simple title
    author:
      - given: John
        family: Smith
  - type: post
    id: test3
    title: “Most kids are fine!”, people say
    issued: 2022-06-02T01:57:29Z
    URL: https://twitter.com/zeynep/status/1532356192567189506
```

... and converted to JSON:

```json
{
  "version": 1.1,
  "description": "An example CSL YAML 1.1 file",
  "references": [
    {
      "type": "book",
      "id": "test1",
      "issued": "2020",
      "title": {
        "main": "Title",
        "sub": [
          "Subtitle"
        ]
      },
      "author": [
        {
          "given": "Jane",
          "family": "Doe"
        }
      ]
    },
    {
      "type": "article-journal",
      "id": "test2",
      "issued": "2020-07/2020-08",
      "title": "Simple title",
      "author": [
        {
          "given": "John",
          "family": "Smith"
        }
      ]
    },
    {
      "type": "post",
      "id": "test3",
      "title": "“Most kids are fine!”, people say",
      "issued": "2022-06-02T01:57:29.000Z",
      "URL": "https://twitter.com/zeynep/status/1532356192567189506"
    }
  ]
}
```

... with schema-backed completion in VSCode:

Screenshot from 2022-06-01 21-09-33

Seems this ecosystem has evolved in the past two years. Once we settle on a final version, we really need to get this schema published on https://www.schemastore.org/json/, which makes editor support all the easier.

cc @jgm @bwiernik @denismaier @cormacrelf @retorquere

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -99,8 +99,7 @@
     "edtf-string": {
       "title": "EDTF date string",
       "description": "CSL input supports EDTF, levels 0 and 1.",
-      "type": "string",
-      "pattern": "^[0-9-%~X?./]{4,}$"
+      "type":"string"
     },
     "date-object-v2": {
       "title": "Date Object (v2)",
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -2,18 +2,38 @@
   "title": "CSL JSON/YAML",
   "description": "JSON schema for CSL input data",
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "https://resource.citationstyles.org/schema/latest/input/json/csl-data.json",
-  "type": "array",
-  "items": {
-    "$ref": "#/definitions/refitem"
+  "$id": "https://resource.citationstyles.org/schema/v1.1/input/json/csl-data.json",
+  "type": "object",
+  "required": ["references", "version"],
+  "properties": {
+    "version": {
+      "type": "number",
+      "enum": [1.1]
+    },
+    "title": {
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    },
+    "references": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/refitem"
+      }
+    }
   },
   "definitions": {
     "title-variable": {
       "title": "Title variable",
       "description": "Titles are represented as a string or as an object.",
       "anyOf": [
-        { "$ref": "#/definitions/rich-text-string" },
-        { "$ref": "#/definitions/title-object" }
+        {
+          "$ref": "#/definitions/rich-text-string"
+        },
+        {
+          "$ref": "#/definitions/title-object"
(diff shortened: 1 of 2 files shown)


retorquere:
Regexes cannot realistically test for valid edtf dates across the range of possible dates/date ranges/date combos expressable in edtf, so it'd either be too lenient or too restrictive.

bdarcus:
@retorquere - do you think I should just remove the regex "pattern" here entirely? 

```json
    "edtf-string": {
      "title": "EDTF date string",
      "description": "CSL input supports EDTF, levels 0 and 1.",
      "type": "string",
      "pattern": "^[0-9-%~X?./]{4,}$"
    }
```

Or do you have a suggestion for a more lenient regex, that would still be useful?

retorquere:
I've been trying to find the place where I got it without success but I think I read at some point that EDTF dates are not regular, so regexes are problematic to parse/validate them. I know just enough about EDTF to leave validation to @inukshuk.

I for sure could not write a regex that would sensibly leniently tell valid from invalid dates. Personally, I'd add edtf validation as custom format attributes, and provide a reference implementation using ajv + EDTF.js. If I'm reading https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.2.3 right, processors that don't know about these custom properties would validate presence but not format of the property.

bdarcus:
> the regex above would reject date sequences and significant-digit markers (`S`), letter-prefixed calendar year (`Y`), and dates with time.

The last is important for us. Are the others?

retorquere:
I don't know, honestly. Does the processor do anything with date sequences? Letter-prefixed seems valid, even if I've not seen it in the wild yet.

PR #423, opened 2022-06-12
input: Add 'format' property for edtf dates

### LINKED ISSUE #421: Update EDTF regex to be more lenient, or use custom format instead
While working on #420, I realize the EDTF regex is too strict. In that case, I noticed it doesn't allow date-times, but I'm sure there are others, and my regex skills are too limited to feel comfortable tackling it on that PR.

See also https://github.com/citation-style-language/documentation/issues/145

-----

See https://github.com/citation-style-language/schema/pull/420#issuecomment-1145082063

> Regexes cannot realistically test for valid edtf dates across the range of possible dates/date ranges/date combos expressable in edtf, so it'd either be too lenient or too restrictive.

So really the question is what, more lenient, regex will work here?

bdarcus:
This allow use of an edtf.js-based validator on the dates.

https://github.com/retorquere/json-schema-edtf

The json schema spec says tools should ignore the property
if they don't know what to do with its value.

Close #421

### DIFF
--- a/schemas/input/csl-data.json
+++ b/schemas/input/csl-data.json
@@ -99,7 +99,8 @@
     "edtf-string": {
       "title": "EDTF date string",
       "description": "CSL input supports EDTF, levels 0 and 1.",
-      "type":"string"
+      "type":"string",
+      "format": "edtf/level-1+season-intervals"
     },
     "date-object-v2": {
       "title": "Date Object (v2)",


bdarcus:
So is this good @retorquere, and maybe I just add a link in the documentation https://github.com/retorquere/json-schema-edtf?

One substantive question: adding season intervals is important for us. Does it mean, however, that those aren't valid edtf dates,  and that other edt validators will flag them?

retorquere:
I welcome seeing the documentation improved - I didn't know what else to put there.

It seems likely to me that other edtf processors would report season intervals as invalid dates, but given that we need these season intervals, the CSL dates would be "EDTF plus season intervals", not "EDTF", and anyone that would choose to treat dates as strictly EDTF in something claiming to be CSL would be out of line with the CSL spec. A clear case of "don't do that" I think.

bdarcus:
> I welcome seeing the documentation improved - I didn't know what else to put there.

To be clear, I was meaning here.

PR #428, opened 2023-01-25
use absolute URI when referencing CSL schema

### LINKED ISSUE #427: use canonical $id when referencing CSL data schema
In the JSON Schema for Cite Items, the `itemData` property points to the CSL Data schema, here:

https://github.com/citation-style-language/schema/blob/5b8bbc824e026959417757d4ce4012a26b10e637/schemas/input/csl-citation.json#L30-L32

However, the `$id` for the CSL Data schema is `https://resource.citationstyles.org/schema/v1.0/input/json/csl-data.json`, not just `csl-data.json`.

Because of this, validation with most validation tools will always fail for the Cite Items schema by default, since these tools won't know where to look for `csl-data.json`, because the [retrieval URI](https://json-schema.org/understanding-json-schema/structuring.html#retrieval-uri) can vary.

For example, I'm using [ajv](https://ajv.js.org/) to validate some data that includes CSL objects, but validation always fails when it reaches `itemData`. The workaround is to either a) save the CSL schema locally and then manually change the `itemData/$ref` property, which isn't an ideal development workflow, and will break when other people try to use my library; or b) use `csl-data.json` as the schema key when adding the CSL Data schema to ajv, which is hacky and forces me to choose a key that isn't consistent with the rest of my schema keys.

Can the `$ref` for the `itemData` property be updated to use the absolute URI of the CSL Data's `$id` property?

dwhieb:
## Description

The JSON Schema for a Cite Item references the CSL Data schema (with `$ref`) using a relative URI, `csl-data.json`. However, the `$id` for the CSL Data schema is an absolute URI, which means that attempts to validate data against the Cite Item schema will typically fail without a workaround.

This PR changes the reference to the CSL Data schema to use its absolute URI instead of a relative one.

Fixes #427

## Type of change

Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/schemas/input/csl-citation.json
+++ b/schemas/input/csl-citation.json
@@ -28,7 +28,7 @@
                         ]
                     },
                     "itemData": {
-                        "$ref": "csl-data.json#/items" 
+                        "$ref": "https://resource.citationstyles.org/schema/v1.0/input/json/csl-data.json#/items" 
                     },
                     "prefix": {
                         "type": "string" 


bdarcus:
Thank you!

